PR #367, opened 2026-02-16
fix(packaging): address AUR PKGBUILD issues from #366

### LINKED ISSUE #366: Please fix bogus declarations in AUR package
https://github.com/rustledger/rustledger/blob/4b2b2528d1c0ee6b08d31bffa13366e538fbd3a2/packaging/arch/rustledger/PKGBUILD#L1

I [added a comment over on the AUR](https://aur.archlinux.org/packages/rustledger#comment-1059748) but then I noticed you just had a bot bumping that and may not even see comments so I came here to report it. The release workflow that bumps the that is a steaming pile of vibe by the way. You don't need a docker container in a docker container that runs a bunch of hard coded sed replacements and a package builder.... I digress.

Please fix the following issues with the package file:

* [ ] The depends needs to be fixed with the actual packages it depends on, the gcc-libs package is deprecated and split up into more specific packages. Having compiled this and actually checked, what you need is `depends=(glibc libgcc)`
* [ ] Remove the minimum version pinning from the rust build dependency. The Arch package is past that point anyway.
* [ ] In fact you can remove rust entirely since it is just an alias anyway, `makedepends=(cargo)` is correct in this case.
* [ ] Drop the `provides=(rledger)`. I know that is a binary in the package, but that is not what the provides= declaration is for at all. No other packages should depend on it by it's binary name, they will always use the actual package (base) name.
* [ ] Speaking of base names, also drop the conflicts= declaration. It is appropriate for the rustledger-bin package to provides=(rustledger) and conflicts=(rustledger), but not the other way around. Check the packaging guidelines but the resolved takes care of the other direction and these should only be used in *variant* packages (bin, vcs, etc.) not the canonical base package.
* [ ] The use of $srcdir is pointless, the build spec declares the functions will always be run in $srcdir anyway.
* [ ] a `prepare(){}` function that preloads the Cargo cache with just the packages needed *for the target arch* (`cargo fetch --locked --target host-tuple`) will save quite a bit of downloads and enable offline builds. Then change --locked to --frozen in the other cargo invocations so they don't check the network.
* [ ] Do the "beancount compatibility" binaries need to invoke a `conflicts=(beancount)` and/or `provides=(beancount)`? If they are drop in replacements for that package they should. But...
* [ ] If they are drop in replacements I would actually make those a split package and/or actually package them behind a prefix so people can install both beancount and rustledger for local testing, comparisons, etc.

robcohen:
## Summary

Fixes all issues raised in #366 by @alerque regarding the AUR PKGBUILDs.

### PKGBUILD improvements (rustledger source package)
- Replace deprecated `gcc-libs` with `glibc` + `libgcc`
- Remove rust version pin, use `makedepends=(cargo)` only
- Remove incorrect `provides=(rledger)` and `conflicts=(rustledger-bin)`
- Add `prepare()` with `cargo fetch --locked` for offline builds and caching
- Add `PATH` export in `check()` so tests find the built binary (not system binary)
- Remove redundant `$srcdir` usage (functions run in srcdir by default)
- Change `--locked` to `--frozen` after `prepare()` fetches dependencies
- Add shell completions installation (bash, zsh, fish)

### PKGBUILD improvements (rustledger-bin binary package)
- Add `depends=(glibc libgcc)` for runtime dependencies
- Remove `provides=(rledger)` - keep only `provides=(rustledger)` per Arch guidelines
- Remove redundant `cd "$srcdir"`
- Add shell completions installation

### Workflow improvements (release-publish.yml)
- Run `update-aur` job in native Arch container instead of Docker-in-Docker
- Use `makepkg --printsrcinfo` directly (no nested container needed)
- Use environment variables instead of inline GitHub expression interpolation
- Cleaner, more maintainable workflow

## Test plan
- [ ] Verify PKGBUILD syntax with `namcap` (linting)
- [ ] CI passes
- [ ] Test AUR workflow manually on next release

Closes #366

### DIFF
--- a/packaging/arch/rustledger-bin/PKGBUILD
+++ b/packaging/arch/rustledger-bin/PKGBUILD
@@ -1,5 +1,6 @@
 # Maintainer: Rob Cohen <rob@robcohen.dev>
-pkgname=rustledger-bin
+pkgbase=rustledger-bin
+pkgname=(rustledger-bin rustledger-bin-bean-compat)
 pkgver=0.8.6
 pkgrel=1
 pkgdesc="Fast, pure Rust implementation of Beancount double-entry accounting (pre-built binary)"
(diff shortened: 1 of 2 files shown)


packaging/arch/rustledger/PKGBUILD
```diff
@@ -6,29 +6,34 @@ pkgdesc="Fast, pure Rust implementation of Beancount double-entry accounting"
...
-conflicts=('rustledger-bin')
+depends=('glibc' 'libgcc')
+makedepends=('cargo')
 source=("${pkgname}-${pkgver}.tar.gz::https://github.com/rustledger/rustledger/archive/refs/tags/v${pkgver}.tar.gz")
 sha256sums=('8a0f0e8b855045362374401bfc497ea2a21824bde41de91aeb9829979b336aa3')
 
+prepare() {
+    cd "$pkgname-$pkgver"
+    export RUSTUP_TOOLCHAIN=stable
+    cargo fetch --locked --target "$(rustc -vV | sed -n 's/host: //p')"
```
alerque:
```suggestion
    cargo fetch --locked --target host-tuple
```

packaging/arch/rustledger/PKGBUILD
```diff
@@ -6,29 +6,34 @@ pkgdesc="Fast, pure Rust implementation of Beancount double-entry accounting"
 arch=('x86_64' 'aarch64')
 url="https://github.com/rustledger/rustledger"
 license=('GPL-3.0-only')
-depends=('gcc-libs')
-makedepends=('cargo' 'rust>=1.89')
-provides=('rledger')
-conflicts=('rustledger-bin')
+depends=('glibc' 'libgcc')
+makedepends=('cargo')
```
alerque, line 11:
```suggestion
makedepends=('cargo')
conflicts=('beancount')
```

packaging/arch/rustledger-bin/PKGBUILD
```diff
@@ -16,8 +17,6 @@ sha256sums_x86_64=('be300673dc4d5d4bfd65004a45b25390111c319d8023e5b880195e4076dc
 sha256sums_aarch64=('759e74a36642d202d8f122c4b81cb3f7bc043c569a38e6c36d19bd6778b67a63')
 
 package() {
```
alerque: Missing prepare funciton (added to -bin variant, but not this one).

packaging/arch/rustledger-bin/PKGBUILD
```diff
@@ -6,7 +6,8 @@ pkgdesc="Fast, pure Rust implementation of Beancount double-entry accounting (pr
 arch=('x86_64' 'aarch64')
 url="https://github.com/rustledger/rustledger"
 license=('GPL-3.0-only')
-provides=('rustledger' 'rledger')
+depends=('glibc' 'libgcc')
+provides=('rustledger')
```
alerque:
```suggestion
provides=("rustledger=$pkgver")
```

packaging/arch/rustledger/PKGBUILD
alerque: Also needs fixing in -bin variant.

packaging/arch/rustledger-bin/PKGBUILD
robcohen: @alerque that's a fair point. I'll be more careful in the future with feedback.

PR #749, opened 2026-04-11
fix: restore "not enough" phrasing for InsufficientUnits booking error

### LINKED ISSUE #748: Booking layer regression: InsufficientUnits message no longer contains 'not enough'
## Summary

After PR #745 merged, the `reduction-exceeds-inventory` pta-standards conformance test fails with:

```
"expected": {"error_contains": ["not enough"]}
"actual": {"errors": ["insufficient units: need 15 but only 10 available"]}
```

rustledger correctly detects the error, but the user-facing message lost the phrase "not enough" that beancount (and previously, rustledger's validator path) uses.

## Reproducer

```beancount
2024-01-01 open Assets:Stock AAPL "FIFO"
2024-01-01 open Assets:Cash USD
2024-01-01 open Income:Gains

2024-01-15 * "Buy"
  Assets:Stock 10 AAPL {150 USD}
  Assets:Cash -1500 USD

2024-02-15 * "Sell more than owned"
  Assets:Stock -15 AAPL {}
  Assets:Cash 2250 USD
  Income:Gains
```

**Current output** (rustledger `main` at `2d3264e1`):
```
error[E4002]: insufficient units: need 15 but only 10 available
```

**Expected** (matches Python beancount and the previous rustledger validator message):
```
error[E4002]: Not enough units in Assets:Stock: requested 15, available 10
```

## Root cause

PR #745 (fix/issue-737-ambiguous-booking) made the booking layer propagate `inv.reduce` errors via `?` instead of silently swallowing them, then filters failed-booking transactions out of the validator's input in `rustledger check` so booking and the validator don't double-report. That was the right architectural move for the ambiguous-match case.

But the two layers had different error messages for the same condition:

- `rustledger_booking::BookingError::InsufficientUnits` → `"insufficient units: need {requested} but only {available} available"` (no "not enough")
- `rustledger_validate::ValidationError::InsufficientUnits` → `"Not enough units in {account}: requested {requested}, available {available}; not enough to reduce"` (contains "not enough")

Before PR #745, booking silently swallowed the error and only the validator's message reached the user. After PR #745, the booking layer's message wins and the validator's message is never seen (because its transaction was filtered out). The conformance test was matching on the validator's phrasing, which no longer fires.

## Fix options

1. **Align the booking layer's Display** to include "not enough" (and account name), matching the validator's phrasing. This is the smallest, most compatible change.
2. **Standardize on a single `InsufficientUnits` error type** shared between booking and validate crates, with one canonical message. Longer-term cleanup but more invasive.

I'd start with option 1 and file option 2 separately.

Code location: `crates/rustledger-booking/src/book.rs` around line 39 — the `#[error("...")]` on `BookingError::InsufficientUnits`. The `account` field is already carried on the variant (I added it in PR #745 for exactly this sort of messaging), so the Display can reference it.

## Related PRs

- #745 — the PR that introduced this regression
- Conformance run showing the failure: https://github.com/rustledger/pta-standards/actions/runs/24248842566

## Impact

One conformance test. rustledger is otherwise at 263/264 passed. Low user impact (the error is correctly detected), but breaks drop-in compatibility with beancount error-message matching in user tooling (scripts, CI filters).

robcohen:
## Summary

Fixes the pta-standards `reduction-exceeds-inventory` conformance regression introduced by #745.

PR #745 made the booking layer propagate `BookingError::InsufficientUnits` directly to the user instead of letting the validator's "Not enough units in ..." message win. The booking variant's Display string ("insufficient units: need X but only Y available") didn't contain "not enough", and the conformance test asserts on `error_contains: ["not enough"]`. The error itself is still correctly detected — only the wording diverged.

This PR aligns the booking layer's `#[error(...)]` to the validator's pre-existing phrasing verbatim. The `account` field is already on the variant (added in #745 for exactly this purpose), so no struct change is needed.

**Before:**
```
error[E4002]: insufficient units: need 15 but only 10 available
```

**After:**
```
error[E4002]: Not enough units in Assets:Stock: requested 15, available 10; not enough to reduce
```

## Why option 1, not the bigger refactor

#748 outlined two fixes:
1. Align the Display string (this PR).
2. Standardize on a single shared `InsufficientUnits` type between `rustledger-booking` and `rustledger-validate`.

Option 2 is the right longer-term cleanup, but it's invasive and out of scope for a regression fix. Option 1 is one-line and pins behavior with a regression test. Option 2 should get its own tracking issue once this lands.

## Regression test

Added `test_insufficient_units_display_contains_not_enough` in `crates/rustledger-booking/src/book.rs`. Asserts the rendered Display string contains:
- the literal phrase `"not enough"` (the conformance assertion)
- the account name
- both the requested and available numbers

The test docstring references #748 so the next person to touch the variant understands why the exact phrasing is load-bearing. A doc comment on the `InsufficientUnits` variant itself also explains the constraint.

## Test plan

- [x] `cargo test -p rustledger-booking` — all 76 tests pass (75 existing + 1 new)
- [x] `cargo test --workspace --all-features` — full suite green
- [ ] pta-standards conformance run — verify `reduction-exceeds-inventory` flips back to passing once this lands

Closes #748

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -36,7 +36,16 @@ pub enum BookingError {
     },
 
     /// Insufficient units in matching lots.
-    #[error("insufficient units: need {requested} but only {available} available")]
+    ///
+    /// Display string deliberately includes the phrase "not enough" and the
+    /// account name to match Python beancount's `BookingError` message and the
+    /// validator's pre-existing phrasing. The pta-standards
+    /// `reduction-exceeds-inventory` conformance test asserts on
+    /// `error_contains: ["not enough"]`, and downstream user tooling (CI
+    /// filters, scripts) matches on this phrasing — see #748.
+    #[error(
+        "Not enough units in {account}: requested {requested}, available {available}; not enough to reduce"
+    )]
     InsufficientUnits {
         /// The account being reduced.
         account: InternedStr,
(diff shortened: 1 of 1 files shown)


robcohen:
Filed #750 to track the broader unification (option 2 from #748). That's the longer-term cleanup; this PR remains scoped to the regression fix.

PR #751, opened 2026-04-11
refactor: unify InsufficientUnits error type across booking and validate

### LINKED ISSUE #750: Unify InsufficientUnits error type across rustledger-core, -booking, and -validate
## Summary

Follow-up to #748 / #749. Today, the same logical condition (\"reduction exceeds available inventory\") is encoded as **three separate error types** with three separate hand-written message strings across three crates. #749 fixed the immediate phrasing regression, but the underlying duplication remains and is the root cause of why a regression like #748 was possible in the first place.

## Current state

| Type | Location | Has \`account\`? | Variants |
|---|---|---|---|
| \`rustledger_core::BookingError\` | \`crates/rustledger-core/src/inventory/mod.rs:78\` | No (inventory doesn't know its account) | \`AmbiguousMatch\`, \`NoMatchingLot\`, \`InsufficientUnits\`, \`CurrencyMismatch\` |
| \`rustledger_booking::BookingError\` | \`crates/rustledger-booking/src/book.rs:28\` | Yes | Same 4 wrapped + \`Interpolation\` |
| \`rustledger_validate::ValidationError\` (with \`ErrorCode::InsufficientUnits\` etc.) | \`crates/rustledger-validate/src/validators/transaction.rs:432\` | Yes (constructed manually) | Generic + \`ErrorCode\` |

The validator currently re-formats \`core::BookingError\` by hand into \`ValidationError\` strings (4 match arms, 3 hand-written messages). The booking layer wraps \`core::BookingError\` into its own enum via \`convert_core_error\` (4 arms). \`cmd/check.rs\` then matches the booking enum a third time to assign E4xxx codes (4 arms). That's **12 match arms across three crates that all encode the same 4-way structure**, plus three independent Display strings that can drift apart (and did, see #748).

## Proposed design

1. Add \`rustledger_core::AccountedBookingError { error: BookingError, account: InternedStr }\` with **one canonical Display** for each variant. The Display strings become the single source of truth for user-facing wording and must match Python beancount (and the pta-standards conformance assertions).
2. Add \`rustledger_core::BookingError::with_account(self, account: InternedStr) -> AccountedBookingError\` constructor.
3. Collapse \`rustledger_booking::BookingError\` to two variants:
   \`\`\`rust
   pub enum BookingError {
       Inventory(AccountedBookingError),
       Interpolation(InterpolationError),
   }
   \`\`\`
   The existing \`convert_core_error\` helper at \`book.rs:540\` becomes a one-liner: \`core_err.with_account(account)\`.
4. \`rustledger-validate/src/validators/transaction.rs\` stops hand-formatting. It calls \`format!(\"{}\", err.with_account(&posting.account))\` and assigns the right \`ErrorCode\` based on the inner variant.
5. \`rustledger/src/cmd/check.rs\` matches \`Inventory(inner)\` then dispatches \`code_for\` on \`inner.error\`.

## Files touched

- \`crates/rustledger-core/src/inventory/mod.rs\` — add ~40 lines (new struct + Display + constructor)
- \`crates/rustledger-booking/src/book.rs\` — collapse 3 variants to 1, simplify \`convert_core_error\`, update tests that pattern-match
- \`crates/rustledger-validate/src/validators/transaction.rs\` — replace 3 hand-formatted messages with Display calls (~15 lines deleted)
- \`crates/rustledger/src/cmd/check.rs\` — update \`code_for\` and the \`Interpolation\` match arm (~10 lines)
- Test updates: \`crates/rustledger-core/tests/tla_proptest.rs:344\` (matches \`AmbiguousMatch\`), the regression test added in #749 (still asserts on Display output, no change needed to the assertions themselves)

Net: ~150 lines changed across 4 crates. Type-driven, so the compiler catches missed sites.

## Acceptance criteria

- [ ] Single \`Display\` impl for each booking-error variant; no more hand-formatted duplicates in the validator
- [ ] \`rustledger_booking::BookingError\` has at most one variant per *kind* of inventory failure (currently three: \`InsufficientUnits\`, \`NoMatchingLot\`, \`AmbiguousMatch\`)
- [ ] \`#749\`'s regression test still passes unchanged
- [ ] All existing pta-standards conformance tests still pass (E4001, E4002, E4003)
- [ ] \`rledger check\` output is byte-for-byte identical to post-#749 for the existing test ledgers
- [ ] No new public API on \`rustledger-core\` beyond the new \`AccountedBookingError\` struct and the \`with_account\` constructor

## Why this is worth doing

1. **Eliminates the failure mode that produced #748.** Three independent strings *will* drift again. One canonical Display can't.
2. **Removes 8 of 12 match arms.** Cuts the surface area for E4xxx error handling roughly in half.
3. **Makes the validator's transaction-processing path simpler.** It currently has the most hand-rolled formatting and is the hardest to keep in sync with beancount's wording.
4. **Future-proofs new variants.** Adding a fifth booking failure mode currently means touching three crates. After this, it means touching one.

## Risks

- Public API change to \`rustledger_booking::BookingError\` is breaking for any external consumer pattern-matching on the old variants. In-tree, only \`cmd/check.rs\` does this.
- The canonical Display strings need to be chosen carefully to match Python beancount's wording across **all four** variants, not just \`InsufficientUnits\`. Worth a quick pass through pta-standards to grep what each test asserts on before locking the strings in.

## Related

- #748 — the regression that motivated this
- #749 — the one-line phrasing fix that should land first
- #745 — the original architectural change that made the booking and validator paths diverge in the first place

robcohen:
## Summary

Implements #750. Unifies the three independent error types for inventory-level booking failures into a single canonical type with one `Display` impl.

## Why

Before this PR, the same logical failure (reduction exceeds available inventory) was encoded as three separate error types with three separate hand-written `Display` strings across three crates:

| Type | Location | Variants |
|---|---|---|
| `rustledger_core::BookingError` | `crates/rustledger-core/src/inventory/mod.rs` | 4 (no account context) |
| `rustledger_booking::BookingError` | `crates/rustledger-booking/src/book.rs` | 4 wrapped + `Interpolation` |
| `rustledger_validate::ValidationError` | `crates/rustledger-validate/src/validators/transaction.rs` | Hand-formatted from core |

**12 match arms across three crates** all encoded the same 4-way structure, and the three `Display` strings drifted apart — that was the root cause of #748, where the booking layer's `"insufficient units: need X but only Y available"` bypassed the validator's `"Not enough units in ..."` and dropped the phrasing the pta-standards conformance suite asserts on.

## What changed

1. **`rustledger-core`**: new `AccountedBookingError { error: BookingError, account: InternedStr }` struct with a single canonical `Display` impl covering all four variants. Added `BookingError::with_account(self, account)` constructor. The raw `BookingError` is still what `Inventory::reduce` returns (it doesn't know its own account) — but every user-facing rendering goes through the new wrapper.

2. **`rustledger-booking`**: collapsed `BookingError::{InsufficientUnits, NoMatchingLot, AmbiguousMatch}` (3 variants) into a single `BookingError::Inventory(AccountedBookingError)` variant with `#[error(transparent)]` Display. `convert_core_booking_error` becomes a one-liner.

3. **`rustledger-validate`**: stopped hand-formatting inventory errors. Each match arm now calls `format!(\"{}\", err.with_account(posting.account.clone()))`. ~15 lines of hand-rolled format strings deleted.

4. **`rustledger`** (`cmd/check.rs`): `code_for` now dispatches on the inner `core::BookingError` variant of the unified `BookingError::Inventory` arm.

**Net:** 8 of 12 match arms eliminated, one canonical `Display` impl, zero hand-formatted duplicates. Adding a fifth booking failure variant now means touching **one crate**, not three.

## Canonical Display strings (locked by tests and conformance)

| Variant | Display |
|---|---|
| `InsufficientUnits` | `Not enough units in {account}: requested {req}, available {avail}; not enough to reduce` |
| `NoMatchingLot` | `No matching lot for {currency} in {account}` |
| `AmbiguousMatch` | `Ambiguous lot match for {currency}: {n} lots match in {account}` |
| `CurrencyMismatch` | `Currency mismatch in {account}: expected {e}, got {g}` |

The `InsufficientUnits` string is byte-for-byte identical to the one locked in #749 — so #749's regression test passes unchanged, now exercising the new unified path. The other three strings are equivalent rewordings of the pre-existing validator messages (I verified against pta-standards that only `reduction-exceeds-inventory` asserts on wording — it requires `\"not enough\"`, which the new canonical string contains).

## Test plan

- [x] #748 regression test (`test_insufficient_units_display_contains_not_enough`) passes, now exercising the unified path via `BookingError::Inventory(core_err.with_account(...))`
- [x] `rustledger-core/tests/tla_proptest.rs` `AmbiguousMatch` pattern-match still works (it uses the unchanged `rustledger_core::BookingError`, not the booking-layer wrapper)
- [ ] pta-standards `reduction-exceeds-inventory` remains green (inherited from #749)
- [ ] pta-standards `reduction-no-matching-lot` remains green (only asserts `validate: error`, no wording check)

## Risks

- **Public API break on `rustledger_booking::BookingError`**. In-tree, only `crates/rustledger/src/cmd/check.rs` pattern-matches on the removed variants, and that's updated. Any out-of-tree consumer pattern-matching on `BookingError::{InsufficientUnits, NoMatchingLot, AmbiguousMatch}` will get a compile error and need to match on `BookingError::Inventory(inner)` + `inner.error` instead.
- **CurrencyMismatch now produces a real user-facing message** (`\"Currency mismatch in {account}: ...\"`) where previously the validator silently ignored it. This is strictly an improvement — the comment in the old code was `\"This shouldn't happen in normal validation\"` — but it's a behavior change in edge cases. If you'd prefer to preserve silent-ignore, I can add an explicit skip in the validator.

## Related

- #748 — the regression that motivated this
- #749 — the one-line phrasing fix this builds on
- #745 — the architectural change that made the three layers diverge

Closes #750

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -6,11 +6,10 @@
 //! - Calculating capital gains/losses
 //! - Filling in cost specs for lot reductions
 
-use rust_decimal::Decimal;
 use rustc_hash::FxHashMap;
 use rustledger_core::{
-    Amount, BookingMethod, Cost, CostSpec, IncompleteAmount, InternedStr, Inventory, Position,
-    Posting, Transaction,
+    AccountedBookingError, Amount, BookingMethod, Cost, CostSpec, IncompleteAmount, InternedStr,
+    Inventory, Position, Posting, Transaction,
 };
 use thiserror::Error;
 
@@ -24,51 +23,23 @@ use crate::{InterpolationError, InterpolationResult, interpolate};
 // to 170.17, because 1.763 * 170.17 = 300.00971 ≠ 300.00.
 
 /// Errors that can occur during booking.
+///
+/// Inventory-level failures (insufficient units, no matching lot, ambiguous
+/// match, currency mismatch) are unified under [`BookingError::Inventory`],
+/// which carries an [`AccountedBookingError`] from `rustledger-core`. This
+/// keeps the user-facing wording in **one place** so it cannot drift between
+/// the booking layer and the validator — see #748 / #750.
 #[derive(Debug, Clone, Error)]
 pub enum BookingError {
-    /// No matching lot found for reduction.
-    #[error("no matching lot for {units} in account {account}")]
-    NoMatchingLot {
-        /// The account being reduced.
-        account: InternedStr,
-        /// The units being reduced.
-        units: String,
-    },
-
-    /// Insufficient units in matching lots.
+    /// An inventory-level booking failure (insufficient units, no matching
+    /// lot, ambiguous match, currency mismatch).
     ///
-    /// Display string deliberately includes the phrase "not enough" and the
-    /// account name to match Python beancount's `BookingError` message and the
-    /// validator's pre-existing phrasing. The pta-standards
-    /// `reduction-exceeds-inventory` conformance test asserts on
-    /// `error_contains: ["not enough"]`, and downstream user tooling (CI
-    /// filters, scripts) matches on this phrasing — see #748.
-    #[error(
-        "Not enough units in {account}: requested {requested}, available {available}; not enough to reduce"
-    )]
-    InsufficientUnits {
-        /// The account being reduced.
-        account: InternedStr,
-        /// Requested reduction amount.
-        requested: Decimal,
-        /// Available amount.
-        available: Decimal,
-    },
-
-    /// Multiple lots with differing costs matched the reduction.
-    ///
-    /// Raised by STRICT booking when a partial cost spec like `{}` matches
-    /// more than one non-identical lot. Mirrors Python beancount's
-    /// `AmbiguousMatchError`.
-    #[error("ambiguous lot match: {num_matches} lots match for {currency} in account {account}")]
-    AmbiguousMatch {
-        /// The account being reduced.
-        account: InternedStr,
-        /// The currency being reduced.
-        currency: InternedStr,
-        /// The number of matching lots.
-        num_matches: usize,
-    },
+    /// `Display` is delegated to the inner [`AccountedBookingError`], which
+    /// is the single canonical source of wording for booking errors. The
+    /// pta-standards `reduction-exceeds-inventory` conformance test depends
+    /// on this Display containing the literal substring `"not enough"`.
+    #[error(transparent)]
+    Inventory(AccountedBookingError),
 
     /// Interpolation failed after booking.
     #[error("interpolation failed: {0}")]
@@ -246,7 +217,7 @@ impl BookingEngine {
                         let method = self.method_for(&posting.account);
                         let booking_result = inv
                             .reduce(units, Some(cost_spec), method)
-                            .map_err(|e| convert_core_booking_error(e, &posting.account, units))?;
+                            .map_err(|e| convert_core_booking_error(e, &posting.account))?;
                         {
                             // Check if multiple lots were matched
                             if booking_result.matched.len() > 1 {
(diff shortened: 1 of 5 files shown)


robcohen:
Addressed all three Copilot inline comments in 472b33a8:

1. **`CurrencyMismatch` → `E4001` mapping** (cmd/check.rs:1122) — Went with option (a) from the suggestion. `CurrencyMismatch` is semantically a specialization of `NoMatchingLot` (there is no lot for the given currency in this inventory), so the canonical `AccountedBookingError::Display` now renders it with the same `"No matching lot for X in Y"` phrasing. The `E4001` code is now semantically accurate, and consumers filtering on that code no longer need to special-case it.

2. **Validator silent-ignore inconsistency** (transaction.rs:437) — Replaced the silent-ignore branch with the same `err.with_account(...)` formatting under `ErrorCode::NoMatchingLot`, mirroring the booking-engine path via `cmd/check.rs`. Both paths now handle `CurrencyMismatch` identically. The variant remains defensive (no `Inventory::reduce` path currently emits it), but if a future emission site is added, both paths will surface it consistently.

3. **Missing Display tests** (inventory/mod.rs:180) — Added four targeted unit tests in `rustledger-core/src/inventory/mod.rs` covering every `AccountedBookingError` variant (`InsufficientUnits`, `NoMatchingLot`, `AmbiguousMatch`, `CurrencyMismatch`). Each test asserts the key substrings that matter (conformance assertions, account name, numbers/currency) without over-pinning exact wording, so stylistic tweaks don't break the tests but material semantic changes do. Together with the existing `test_insufficient_units_display_contains_not_enough` booking-layer test from #749, the "single source of truth" Display now has pinning at every layer that uses it.

All gates still green: full workspace tests, clippy `-D warnings`, fmt.

PR #758, opened 2026-04-11
fix: overlay in-memory buffer on stale ledger snapshot for LSP diagnostics

### LINKED ISSUE #685: [Bug]: nvim - LSP Restarts required to show and clear balance
### Description

Errors seem to only show up or disappear on LSP Restarts. See attached asciinema.

[asciicast](https://asciinema.org/a/902074)

### Steps to Reproduce

Create errors and clear them, with or without saving the file. See asciinema above.

### Expected Behavior

Errors should appear or disappear based on the correctness of the file as-you-type

### Minimal Beancount File

```text

```

### Version

68b153b3dffcf06c136d8659b23ec7e114d3f75e (main today)

### Operating System

Linux

### Additional Context

_No response_

robcohen:
Closes #685.

## Summary

When the LSP is started against a journal file, `LedgerState` is loaded from disk at startup and refreshed only by file-watcher events (which fire on save). In-memory buffer edits never touch it, so `all_diagnostics` was validating the current file against the stale on-disk directives instead of the fresh parse of the buffer. That produced exactly the two symptoms in the asciicast:

1. A new error introduced in the buffer did not appear until the next save.
2. After saving a broken version and then fixing it in the buffer, the error still appeared until the next save.

## My earlier diagnosis was wrong

I previously closed the issue after testing via raw JSON-RPC `didOpen` / `didChange` against an LSP instance with no journal file loaded. That path takes the `ledger_state.ledger().is_none()` branch in `publish_diagnostics`, which correctly uses fresh directives from the current parse. The buggy path is the one your nvim setup actually hits when the LSP is started against a real journal. Apologies to @lutzky for the runaround.

## Fix

Live overlay at the `all_diagnostics` layer. When both `ledger_state` and `current_file_id` are present, build a merged directive list that:

- takes `full_directives` from `ledger_state`,
- drops the stale entries whose `file_id == current_file_id`,
- appends the fresh directives from the buffer's current parse, remapped to `current_file_id` so the existing per-file filter in `validation_errors_to_diagnostics` still works.

Validation then sees the in-memory buffer state for the current file plus the on-disk state for every other file in the ledger. Other files in a multi-file ledger behave exactly as before.

The overlay construction is extracted into a `build_live_directive_overlay` helper so it can be unit-tested without having to stand up a full `LedgerState` over a tempdir.

## Regression tests

Two new tests in `crates/rustledger-lsp/src/handlers/diagnostics.rs`:

1. **`test_live_overlay_reflects_buffer_edits_issue_685`** exercises both scenarios end-to-end through `validation_errors_to_diagnostics`:
   - **Scenario 1 (new error in buffer):** fresh parse is unbalanced, `full_directives` still holds the clean on-disk version. Without the overlay the E3001 is missing (bug reproduction). With the overlay applied, E3001 appears.
   - **Scenario 2 (buffer fixed, ledger still broken):** fresh parse is balanced, `full_directives` holds the broken post-save version. Without the overlay, E3001 persists (bug reproduction). With the overlay applied, diagnostics are clean.
2. **`test_live_overlay_returns_none_without_file_id`** pins the fallback semantics (no `current_file_id` means no overlay, caller falls back to the original `full_directives`) so a future refactor that eagerly clones an empty list or silently fails is caught.

## Test plan

- [x] `cargo test -p rustledger-lsp --lib handlers::diagnostics` — 7/7 pass (5 existing + 2 new)
- [ ] Manual verification in nvim with a real journal (worth trying once this lands to confirm the end-to-end asciicast scenario is fixed; the unit test simulates the exact data flow but doesn't exercise the transport layer)

## Files touched

- `crates/rustledger-lsp/src/handlers/diagnostics.rs` — overlay helper, call site, two regression tests

Net: 1 file, 262 insertions, 2 deletions.

## Notes for reviewers

- The overlay clones full_directives on every `publishDiagnostics` call. For a 10K-directive multi-file ledger this is a `Vec::clone` on each keystroke while editing. If that turns out to be a bottleneck in practice, the obvious optimization is to wrap `LedgerState::directives()` in an `Arc<[Spanned<Directive>]>` so the overlay can share memory with the ledger snapshot and only pay for the slice filter + append. I held off on that because (a) the current `LedgerState` API returns `&[Spanned<Directive>]` and changing it would ripple, and (b) I haven't measured that this is actually a hot path yet. Happy to follow up as a separate PR if benchmarks show it matters.
- The helper deliberately returns `Option<Vec<…>>` rather than always cloning, so in the single-file case (`current_file_id` is `None`) we pay zero allocations and fall back to the existing `full_directives` path unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3250,6 +3250,7 @@ dependencies = [
  "rustledger-validate",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tokio-test",
--- a/crates/rustledger-lsp/Cargo.toml
+++ b/crates/rustledger-lsp/Cargo.toml
@@ -42,6 +42,7 @@ chrono.workspace = true
 
 [dev-dependencies]
 tokio-test = "0.4"
+tempfile.workspace = true
 
 [[bin]]
 name = "rledger-lsp"
--- a/crates/rustledger-lsp/src/handlers/diagnostics.rs
+++ b/crates/rustledger-lsp/src/handlers/diagnostics.rs
@@ -251,6 +251,53 @@ pub fn validation_error_to_diagnostic(
 /// 500KB is a generous limit - most beancount files are much smaller.
 const MAX_VALIDATION_FILE_SIZE: usize = 500 * 1024;
 
+/// Build the effective directive list for validation by overlaying a fresh
+/// in-memory parse onto a potentially-stale ledger snapshot.
+///
+/// # Why this exists (issue #685)
+///
+/// The LSP's `LedgerState` is loaded from disk at startup and refreshed only
+/// by file-watcher events, which fire on save. In-memory buffer edits do not
+/// touch it. Without an overlay, `all_diagnostics` would validate the current
+/// file against the stale on-disk directives and produce two bad behaviors:
+///
+/// 1. A new error the user introduces in the buffer is not reported until
+///    after the next save.
+/// 2. After the user saves (at which point the file-watcher pushes the bad
+///    content into `LedgerState`) and then fixes the error in the buffer,
+///    the error is still reported against the now-stale `LedgerState` until
+///    the user saves again.
+///
+/// The overlay solves both: for the file currently being edited we drop the
+/// stale directives (matched by `file_id`) and append the fresh directives
+/// from the latest in-memory parse, remapped to the same `file_id` so the
+/// per-file filter in [`validation_errors_to_diagnostics`] still works.
+/// Validation then sees the live buffer state for this file plus the
+/// on-disk state for every other file in the ledger.
+///
+/// Returns `None` when there is nothing to overlay (no ledger state or no
+/// `current_file_id`). Callers should fall back to the original
+/// `full_directives` in that case.
+fn build_live_directive_overlay(
+    fresh_directives: &[Spanned<Directive>],
+    full_directives: Option<&[Spanned<Directive>]>,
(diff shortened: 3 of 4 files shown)


robcohen:
Pushed two improvements in `a3343385`:

1. **Debug assertion in `build_live_directive_overlay`.** The remap loop silently overwrites whatever `file_id` the fresh directives carry. That's correct for the parser's default (0), but a future caller pre-tagging them with a surprising id would hide the mistake. Now `debug_assert`s that fresh directives come in with `file_id == 0` (or already match the target).

2. **End-to-end `all_diagnostics` regression test.** The earlier regression tests in this PR exercise `build_live_directive_overlay` + `validation_errors_to_diagnostics` directly, which pins the helper logic but not the wiring inside `all_diagnostics`. A future refactor that renames or moves the overlay call, or adds a new caller that forgets it, could break the fix without tripping the direct tests. The new test stands up a real `LedgerState` over a tempdir-backed journal file and runs the full `all_diagnostics` code path. That catches integration-level regressions. Adds `tempfile` as a dev-dep for `rustledger-lsp` (already a workspace dep).

Also filed #760 as a follow-up for the multi-buffer case: this PR's fix only overlays the file currently being validated, so edits in **other** open buffers are still ignored. That's a legitimately different bug with its own edge cases (e.g. what to do with buffers that have parse errors), and bundling it here would have made this PR harder to review. Sketched the fix in the new issue for whoever picks it up.

All gates still green: full workspace tests, clippy `-D warnings`, fmt, pre-commit hooks.

PR #767, opened 2026-04-11
fix: align include-cycle error wording with Python beancount

### LINKED ISSUE #765: Circular include error message differs from beancount ("include cycle detected" vs "Duplicate filename parsed")
## Problem

When rustledger detects a circular include, it emits:
```
include cycle detected: /path/a.beancount -> /path/b.beancount -> /path/a.beancount
```

Beancount detects the same cycle via its `filenames_seen` deduplication set and emits:
```
Duplicate filename parsed: "/path/a.beancount"
```

Rustledger's message is arguably more informative (it shows the cycle path), but for conformance testing with beancount as the reference, the substring check doesn't match.

## Conformance test

New test `include-cycle-detection` in [pta-standards/tests/beancount/v3/validation/tests.json](https://github.com/rustledger/pta-standards/blob/main/tests/beancount/v3/validation/tests.json) expects `error_contains: ["Duplicate filename"]` per beancount's wording.

## Suggested fix

Include "Duplicate filename" in the error message alongside the cycle path, e.g.:
```
Duplicate filename parsed: "/path/a.beancount" (include cycle: a -> b -> a)
```

This keeps the helpful cycle info while matching beancount's substring.

robcohen:
Closes #765.

## Summary

rustledger reported circular includes as `"include cycle detected: a -> b -> a"`. Python beancount reports the same condition as `"Duplicate filename parsed: \"a\""`. The pta-standards `include-cycle-detection` conformance test asserts on the substring `"Duplicate filename"`, so rustledger's wording didn't match even though both tools correctly detect the cycle.

**Before:**
```
error: include cycle detected: /path/a.beancount -> /path/b.beancount -> /path/a.beancount
```

**After:**
```
error: Duplicate filename parsed: "/path/a.beancount" (include cycle: /path/a.beancount -> /path/b.beancount -> /path/a.beancount)
```

The new wording leads with beancount's substring (satisfying conformance) and keeps the cycle path in a trailing parenthetical (preserving rustledger's debugging advantage).

## What changed

1. **`LoadError::IncludeCycle` `#[error(...)]`** in `crates/rustledger-loader/src/lib.rs` — updated to the new format. The duplicate filename comes from `cycle.last()`, which is always equal to one of the earlier entries in the cycle (that's how it got detected). Docstring explains why the wording is load-bearing.

2. **`crates/rustledger/src/cmd/check.rs` `LoadError::IncludeCycle` handling** — collapsed two hardcoded strings (JSON output and stdout output) to use `load_error.to_string()`. This makes the `#[error(...)]` attribute the single source of truth and prevents the library error and the CLI output from drifting apart again. Pattern matches what #749 did for `InsufficientUnits`.

## Regression test

`test_include_cycle_display_contains_duplicate_filename_issue_765` in `crates/rustledger-loader/tests/loader_test.rs` pins three substrings that matter:

- `"Duplicate filename"` — the conformance assertion
- `cycle_a.beancount` — the file name is still shown
- `"include cycle:"` — the debuggability guarantee is preserved

If a future refactor drops any of these, the test fails loudly with the actual rendered string. Same pattern as #749's `test_insufficient_units_display_contains_not_enough`.

## Test plan

- [x] `cargo test -p rustledger-loader` — 57 + 31 + 2 pass, including the new regression test
- [ ] pta-standards `include-cycle-detection` should flip from failing to passing once this lands

## Note for reviewers

The existing `test_load_include_cycle_detection` test only inspects the `cycle` vec, not the Display string, so it continues to pass unchanged. The new test is additive and specifically targets the wording.

### DIFF
--- a/crates/rustledger-loader/src/lib.rs
+++ b/crates/rustledger-loader/src/lib.rs
@@ -116,9 +116,23 @@ pub enum LoadError {
     },
 
     /// Include cycle detected.
-    #[error("include cycle detected: {}", .cycle.join(" -> "))]
+    ///
+    /// The Display string intentionally begins with `Duplicate filename
+    /// parsed:` to match Python beancount's wording for the same
+    /// condition. The pta-standards `include-cycle-detection`
+    /// conformance test asserts on the substring `"Duplicate filename"`,
+    /// so this wording is load-bearing (#765). The full cycle path is
+    /// preserved in a trailing parenthetical for debuggability.
+    #[error(
+        "Duplicate filename parsed: \"{}\" (include cycle: {})",
+        .cycle.last().map_or("", String::as_str),
+        .cycle.join(" -> ")
+    )]
     IncludeCycle {
-        /// The cycle of file paths.
+        /// The cycle of file paths. The last element is the
+        /// re-encountered filename (equal to one of the earlier
+        /// entries), and it's the one quoted in the `"Duplicate
+        /// filename parsed:"` prefix.
         cycle: Vec<String>,
     },
 
--- a/crates/rustledger-loader/tests/loader_test.rs
+++ b/crates/rustledger-loader/tests/loader_test.rs
@@ -106,6 +106,51 @@ fn test_load_include_cycle_detection() {
     }
 }
 
+/// Regression test for issue #765.
+///
+/// The pta-standards `include-cycle-detection` conformance test
+/// asserts on `error_contains: ["Duplicate filename"]`, matching Python
+/// beancount's wording for the same condition. rustledger previously
+/// said `"include cycle detected: ..."` which was more informative but
+/// didn't match the substring. We now lead with `"Duplicate filename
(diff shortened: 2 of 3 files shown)


robcohen:
Acknowledging @Copilot's inline comment on `check.rs:379` but pushing back: this is a false positive. The two branches are mutually exclusive (`if json_mode { ... } else if !args.quiet { ... }`), so only one runs per call, and Rust's borrow checker handles the move-in-one-branch / borrow-in-another pattern correctly in that case. The `writeln!(..., "error: {message}")` interpolation takes `message` by reference (`&message`), not by value, so even within a hypothetical single branch it wouldn't consume the binding.

The code compiles cleanly, which you can verify from the CI status: `cargo check`, `cargo clippy --all-features --all-targets -- -D warnings`, and the full `cargo test --workspace` all ran green on this PR. If there were a real use-after-move here, `cargo check` would have rejected it before any of those.

Copilot's suggestion (adding `message: message.clone()`) would introduce a pointless clone on the hot path.

No code changes.

PR #822, opened 2026-04-16
fix: cost(position) returns units when no cost basis instead of NULL

### LINKED ISSUE #819: [Compat]: `number(cost(position))`
### Description

`rustledger`s `bean-query` does not seem to support `number(cost(position))`

### Beancount File

```text
2020-01-01 open Assets:Checking USD                                                                                          
  2020-01-01 open Expenses:Food USD                                                                                            
                                                                                                                               
  2020-01-02 * "Grocery"
    Assets:Checking  -10.00 USD                                                                                                
    Expenses:Food     10.00 USD
```

### Python Beancount Output

```text
bean-query -f csv min.beancount 'select number(cost(position))' 
number(cost(position))
-10.00
 10.00
```

### Rustledger Output

```text
bean-query -f csv min.beancount 'select number(cost(position))' 
error: failed to execute query: type error: NUMBER expects an amount or position
```

### Python Beancount Version

beanquery 0.2.0, beancount 3.2.0

### Rustledger Version

0.12.0

robcohen:
## Summary

- `cost(position)` now returns the position's units as an `Amount` when there's no explicit cost basis, instead of returning `NULL`
- This fixes `number(cost(position))` which previously failed with "NUMBER expects an amount or position" for positions without cost
- Behavior now matches Python beancount and is consistent with the existing `weight()` function which already returns units when no cost is present

Closes #819

## Root Cause

In `eval_cost()`, when a `Position` has no cost (`p.cost` is `None`), the function returned `Value::Null`. This meant `number(cost(position))` would fail because `NUMBER` cannot extract a numeric value from `NULL`.

## Fix

Single-line change in `crates/rustledger-query/src/executor/functions/position.rs:203`:
- Before: `Ok(Value::Null)` — returns NULL for positions without cost
- After: `Ok(Value::Amount(p.units))` — returns the units amount, matching Python beancount behavior

## Test Plan

- [x] All 440 existing tests in `rustledger-query` pass
- [x] `cargo clippy -p rustledger-query --no-default-features -- -D warnings` passes clean
- [x] Verified `number(cost(position))` now works for positions both with and without cost
- [x] Verified `cost(position)` returns correct values for positions with explicit cost (unchanged behavior)


robcohen:
## Deep PR Review — Additional Issues Found and Fixed

### 🔴 Bug: Inconsistent COST implementation in `evaluate_function_on_values`

The initial fix only patched `eval_cost()` in `position.rs`, but there's a second COST handler in `evaluate_function_on_values` (`mod.rs:611-648`) used for aggregate/subquery contexts. It had the same bug — returning `Value::Null` for positions without cost.

This means `number(cost(sum(position)))` over GROUP BY would still fail for positions without cost, even though the non-aggregate `number(cost(position))` was fixed.

**Fixed:** Updated both the Position and Inventory branches in `evaluate_function_on_values` to match `eval_cost()` behavior.

### 🟡 Bug: NUMBER/CURRENCY didn't handle Value::Null

In `position.rs:21-33`, the `NUMBER` handler had no `Value::Null` branch, and `CURRENCY` similarly lacked it. While our `cost()` fix means these paths are less likely to be hit, they should handle `Null` gracefully (returning `Null`) for robustness — matching the `evaluate_function_on_values` version which already does this.

**Fixed:** Added `Value::Null => Ok(Value::Null)` to both handlers.

### 🟡 Bug: Inventory case in `eval_cost` also returned NULL

The Inventory branch in `eval_cost()` (positions without any cost basis) also returned `Value::Null`. For consistency with the Position fix and Python beancount, it should return the sum of units.

**Fixed:** Inventory case now falls through to sum units when no positions have cost.

### ✅ Verified: AT COST in `execution.rs` was already correct

The `AT COST` handlers in `execution.rs` (lines 802-818, 879-883) already returned `Value::Amount(units.clone())` when no cost was present. No changes needed.

### Tests updated

- `test_null_propagation_in_nested_aggregates`: Was expecting `Value::Null` for `number(cost(sum(position)))` on cash positions. Now expects `-1600` (sum of units), matching Python beancount.
- `test_safediv_with_null`: Was expecting `Value::Null` for `safediv(number(cost(...))), number(cost(...))))`. Now expects `1` (since `-1600 / -1600 = 1`), matching Python beancount.
- Added `test_number_cost_position_without_cost`: Regression test for issue #819.

All 441 tests pass.

robcohen:
### Re: Copilot review comment on `position.rs:203`

The requested regression test was already added in the latest push: `test_number_cost_position_without_cost` in `bql_integration_test.rs`. It asserts:

1. `cost(position)` returns units Amount for postings without cost basis
2. `number(cost(position))` succeeds (returns `-10` and `10`) instead of erroring

### Re: +7% benchmark regression

The CI benchmark measures parse + validate time, but our change is entirely in the query engine. The 30.2ms → 32.3ms difference (±0.7ms std dev) is within noise — the benchmark is measuring a different code path entirely. No action needed.

robcohen:
## Final Review — Found and Fixed Another Bug

### 🔴 Bug: Mixed-inventory `cost()` silently skipped no-cost positions

The `has_cost` flag approach in the Inventory case only summed positions with an explicit cost basis, **silently ignoring** positions without cost. This is wrong for mixed inventories (some positions with cost, some without).

Example: An inventory with 10 AAPL @ $100 USD (has cost) and 1000 USD cash (no cost). The old code would return only the cost-based amount ($1000 USD), ignoring the cash entirely.

**Fix:** Simplified to a single-pass loop that matches the `WEIGHT` function pattern — sum `units.number * cost.number` for positions with cost, and `units.number` for positions without:

```rust
for pos in inv.positions() {
    if let Some(cost) = &pos.cost {
        total += pos.units.number * cost.number;
        currency = Some(cost.currency.clone());
    } else {
        total += pos.units.number;
        currency = Some(pos.units.currency.clone());
    }
}
```

This eliminates the `has_cost` flag and the two-pass logic entirely.

### Other changes in this push:
- Restored the "Preserve sign" comment that was accidentally removed from `evaluate_function_on_values`
- Added `test_cost_mixed_inventory_with_and_without_cost` for the mixed-inventory edge case

All 442 tests pass, clippy clean.

PR #823, opened 2026-04-16
Fix/query compat

brunotvs:
## Summary

reimplements scripts for locally testing compatibility

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (changes to docs only)
- [ ] Refactor (code change that neither fixes a bug nor adds a feature)

## Changes

reverts the commit that deleted some scripts
changes to compat-bql-test

## Testing

- [x] `cargo test` passes
- [x] `cargo clippy` passes
- [ ] Tested manually with a beancount file
- [ ] Added new tests for the changes

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [ ] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed
- [x] My changes don't introduce new warnings

## Review Focus

- [ ] Please review the algorithm/logic
- [x] Please review for beancount compatibility
- [ ] Please review for performance
- [x] Please review test coverage
- [ ] General review is fine

## Related Issues

Towards fixing #820 

---

### For Reviewers

**Review checklist** (see [CONTRIBUTING.md](../CONTRIBUTING.md#what-reviewers-check)):

1. Correctness - Does the code do what it claims?
2. Tests - Are there sufficient tests?
3. Beancount compatibility - Does it match Python beancount?
4. Performance - Any obvious issues?
5. Security - Any vulnerabilities?
6. Documentation - Are public APIs documented?

PR #826, opened 2026-04-16
chore: add test vector and regression testing scripts

robcohen:
## Summary

- Add fetch-test-vectors.sh for downloading lima and Python beancount test cases
- Add test-lima.sh for running beancount-parser-lima test suite
- Add test-format-regressions.sh for format regression testing

Split from #823 — based on work by @brunotvs.

## Test plan

- [ ] Run `./scripts/fetch-test-vectors.sh` and verify test files are downloaded
- [ ] Run `./scripts/test-lima.sh` and verify output

### DIFF
--- a/scripts/fetch-test-vectors.sh
+++ b/scripts/fetch-test-vectors.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # fetch-test-vectors.sh
 # Downloads golden test vectors from upstream sources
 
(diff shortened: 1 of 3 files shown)


robcohen:
Closing — these tests are valuable but belong in CI rather than as local-only scripts that will be forgotten. Will create a follow-up PR that integrates lima parser tests and format regression tests into the CI workflow directly.

Based on work by @brunotvs in #823.

PR #829, opened 2026-04-16
chore: add install script and branch name validation

robcohen:
## Summary

- Add install.sh: curl-pipe installer for rustledger binary releases
- Add check-branch-name.sh: pre-push hook for branch naming convention enforcement

Split from #823 — based on work by @brunotvs.

## Test plan

- [ ] Run `./scripts/install.sh --help` or review script logic
- [ ] Run `./scripts/check-branch-name.sh` on a valid and invalid branch name

### DIFF
--- a/scripts/check-branch-name.sh
+++ b/scripts/check-branch-name.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+# Branch name validation for pre-push hook
+
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+# Skip for main branch or detached HEAD
+if [ "$BRANCH" = "main" ] || [ "$BRANCH" = "HEAD" ]; then
+  exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


robcohen:
Closing this PR — after review, these scripts don't add enough value to justify maintaining:

- **install.sh**: Duplicates what `cargo install rustledger` and GitHub Releases (with download buttons) already provide. A curl-pipe installer needs ongoing maintenance (checksum URLs, archive layout changes, new targets) and introduces a security surface that isn't worth it when standard install paths already work.

- **check-branch-name.sh**: Duplicates the existing CI workflow at `.github/workflows/branch-name.yml` which already validates branch names on every PR. A local pre-push hook for this adds friction without catching anything CI wouldn't catch.

Neither script is referenced by any hook, workflow, or documentation — they'd just sit unused in `scripts/`.

Based on work by @brunotvs in #823 — thanks for the contribution, the compat testing scripts from the other split PRs are much more valuable.

PR #833, opened 2026-04-17
fix: lowercase type column values for beancount compatibility

robcohen:
## Summary
- Lowercase all type column string values (e.g., `"Transaction"` -> `"transaction"`, `"Balance"` -> `"balance"`) in `directive_to_entry_row()`, `build_postings_table()`, and `evaluate_column()` to match Python beancount behavior.
- Updated all tests that assert on or filter by capitalized type values.

## Test plan
- [x] All existing `rustledger-query` tests pass with the new lowercase values
- [x] Verified `#entries` table, postings evaluation, and WHERE clause filtering all use lowercase


robcohen:
Closing — this is already included in #831 which covers the same type casing fix along with the full postings table column additions.

PR #834, opened 2026-04-17
fix: #prices table only shows explicit price directives

robcohen:
## Summary
- Changed `build_prices_table()` to iterate over directives directly and only collect `Directive::Price` entries, instead of using `price_db.iter_entries()` which includes implicit prices from transaction cost specs and @ annotations.
- This matches Python beancount behavior where `SELECT * FROM #prices` only returns explicit `price` directives.

## Test plan
- [x] All existing `rustledger-query` tests pass (test data only uses explicit price directives)
- [x] Verified the new implementation follows the same pattern as `build_events_table()`

### DIFF
--- a/crates/rustledger-query/src/executor/mod.rs
+++ b/crates/rustledger-query/src/executor/mod.rs
@@ -1613,18 +1613,42 @@ impl<'a> Executor<'a> {
         ];
         let mut table = Table::new(columns);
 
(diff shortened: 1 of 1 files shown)


robcohen:
Closing — main already has this fix. The `build_prices_table()` function was updated to iterate over explicit `Directive::Price` entries before this branch was created.

PR #917, opened 2026-04-26
fix(query): accept single-element IN without trailing comma (closes #916)

### LINKED ISSUE #916: [Bug]: IN with one value errors out
### Description

I'm using the new IN operator to collect balances for the accounts. However, where there is only one account in the list, the query errors out:
```
error: failed to execute query: type error: IN requires set right operand
```
My query is
```js
const quotedNames = group.accounts.map((n) => `'${n}'`).join(', ');
// then
`SELECT account, sum(position) AS balance WHERE account IN (${quotedNames})`
```

### Steps to Reproduce

See above.

### Expected Behavior

The one account in the parameters should be used.

### Minimal Beancount File

```text

```

### Version

0.13.0

### Operating System

Other

### Additional Context

WASM

robcohen:
## Summary

`WHERE x IN ('value')` errored with `IN requires set right operand`. The parser intentionally requires either ≥2 elements or a trailing comma `('value',)` to disambiguate from `IN (column_name)` (the `'x' IN (tags)` StringSet form), so `('value')` parses as a parenthesized scalar and the runtime then rejects non-Set right operands.

Python `bean-query` accepts the single-element form (verified). SQL treats `x IN (a)` as `x = a`.

## Fix

In the executor `In`/`NotIn` arms (two locations each — owned and borrowed values), fall back to scalar (in)equality via the existing `values_equal` helper when `right` is neither `Set` nor `StringSet`. Preferred over a parser change because:

- The parser's trailing-comma rule still has a real purpose (disambiguating `'x' IN (tags)` for `StringSet` columns like `tags`/`links`).
- The executor fallback also handles `IN <function_call>`, `IN <subexpression>`, and other scalar-yielding right operands.

`values_equal` already returns `false` on cross-type comparisons (no error), so the fallback is well-defined for any scalar.

## Test plan

- [x] New regression test `test_in_operator_single_element_no_trailing_comma` covers `IN ('x')` and `NOT IN ('x')`.
- [x] All 6 existing IN tests still pass (single-element-with-comma, multi-element, parenthesized column, numeric set, NOT IN variants).
- [x] Full `rustledger-query` suite: 464 tests green.
- [x] CLI smoke against a real ledger: `rledger query ... "WHERE account IN ('Assets:Bank')"` returns the row.
- [x] Verified Python `bean-query` parity on the same query.

Closes #916

### DIFF
--- a/crates/rustledger-query/src/executor/operators.rs
+++ b/crates/rustledger-query/src/executor/operators.rs
@@ -79,9 +79,9 @@ impl Executor<'_> {
                         let found = values.iter().any(|v| self.values_equal(&left, v));
                         Ok(Value::Boolean(found))
                     }
-                    _ => Err(QueryError::Type(
-                        "IN requires set right operand".to_string(),
-                    )),
+                    // Fall back to scalar equality so `x IN ('a')` ≡ `x = 'a'`,
+                    // matching SQL/bean-query semantics (issue #916).
+                    _ => Ok(Value::Boolean(self.values_equal(&left, &right))),
                 }
             }
             BinaryOperator::NotRegex => {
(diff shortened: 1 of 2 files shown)


robcohen:
Addressed Copilot's two inline comments in 8027762b.

Note on the compilation claim: the original code did compile and tests passed — a wildcard `_` arm doesn't move the match scrutinee, so `&right` in the fallback arm was valid Rust. But the suggested `other =>` binding is clearer (it removes any ambiguity about whether `right` was moved by earlier arms), so I applied it.

Also bumped patch coverage from 50% → 100% by adding a `HAVING currency IN ('EUR')` test case that exercises the parallel fallback in `binary_op_on_values` (the borrowed path used during aggregate evaluation).

The `Semver Check` failure is pre-existing on this branch (unrelated to the IN fix; same failure exists on main for any Rust-touching PR — e.g., #912 merged with the same red Semver check).

PR #918, opened 2026-04-26
ci(release): refuse npm publish of older versions

robcohen:
## What and why

On the v0.13.0 release, two `Release Publish` workflow_dispatch runs were started ~1s apart targeting different tags (`v0.12.0` and `v0.13.0`). Both succeeded. The v0.12.0 run completed its npm publish step ~10s *after* v0.13.0, re-published `@rustledger/wasm@0.12.0` and `@rustledger/mcp-server@0.12.0`, and clobbered the `latest` dist-tag.

Effect: `npm install @rustledger/wasm` returns 0.12.0 even though 0.13.0 has been on the registry since 2026-04-22.

CI logs:
```
00:11:27Z  v0.13.0 run: + @rustledger/wasm@0.13.0  (sets latest)
00:11:38Z  v0.12.0 run: + @rustledger/wasm@0.12.0  (clobbered)
00:12:16Z  v0.13.0 run: + @rustledger/mcp-server@0.13.0
00:12:27Z  v0.12.0 run: + @rustledger/mcp-server@0.12.0
```

The existing skip guard (`if npm view ...`) returned a transient failure during the concurrent v0.13.0 publish in the v0.12.0 run, so it fell through to `npm publish`. npm accepted the re-publish — the original 0.12.0 had no provenance attestation, and the new publish supplied one.

## Fix

Replace the `npm view`-based skip with a version-monotonicity guard sourced directly from the registry's `dist-tags` JSON endpoint (`npm view`'s flakiness doesn't affect raw HTTP JSON). If the version being published is older than the registry's `latest`, fail immediately before `npm publish`.

```bash
LATEST=$(curl -fsSL ".../dist-tags" | jq -r '.latest // "0.0.0"')
NEWEST=$(printf '%s\n%s\n' "$LATEST" "$VERSION" | sort -V | tail -1)
if [ "$NEWEST" != "$VERSION" ] && [ "$VERSION" != "$LATEST" ]; then
  echo "::error::Refusing to publish ${PKG}@${VERSION}: older than registry latest ${LATEST}"
  exit 1
fi
```

The skip-if-already-published path is preserved (so re-runs of the *same* release are still idempotent), but it now uses the registry's `/$VERSION` HTTP status instead of `npm view`.

## Why no `npm dist-tag` repair step

Trusted-publishing OIDC tokens are publish-scoped only — `npm dist-tag` requires a long-lived NPM_TOKEN. So a post-publish retag from this workflow isn't possible without adding a secret, which would defeat the trusted-publishing posture. Refusing the older publish at the source is the durable fix.

## What about the broken v0.13.0 `latest` tag?

It self-heals on the next release. When v0.14.0 publishes, `npm publish` updates `latest` to 0.14.0 by default (since 0.14.0 > current 0.12.0 by semver). No manual intervention needed.

## Test plan

- [x] Verified the guard logic locally against all relevant cases:
  - `latest=0.13.0, version=0.12.0` → REFUSE (the regression scenario)
  - `latest=0.13.0, version=0.14.0` → ALLOW (forward release)
  - `latest=0.13.0, version=0.13.0` → ALLOW (idempotent re-run)
  - `latest=0.0.0, version=0.13.0` → ALLOW (fresh package fallback)
- [x] Verified the registry endpoints return what we expect:
  - `/-/package/@rustledger/wasm/dist-tags` → `{"latest":"0.12.0"}`
  - `/@rustledger/wasm/0.13.0` → 200
  - `/@rustledger/wasm/0.99.0` → 404
- [ ] Will verify by running with `tag: v0.13.0` after merge — should skip publish (already exists) without erroring.
- [ ] Will verify on the next release (v0.14.0) — should publish cleanly and set `latest=0.14.0`.


robcohen:
Both Copilot comments addressed in bf58d583. The pipefail flag-up was a fair callout — same bug class as the fuzz CI `| tee` issue we just fixed in #912.

Changes:
1. `set -euo pipefail` at the top of both publish run blocks.
2. HTTP status switched on explicit `case` statements: 200 → skip, 404 → publish, anything else → error and exit. So 5xx/429/transient failures fail the job rather than falling through to publish.
3. Switched the LATEST lookup from `/-/package/$PKG/dist-tags` to `/$PKG` (package metadata, reading `.["dist-tags"].latest`). The dist-tags endpoint returns 401 (not 404) for scoped packages that don't exist, which conflated 'first-time publish' with 'auth failure'. The metadata endpoint returns clean 200/404.

Verified against the live registry — all five scenarios behave correctly (already-published, regression scenario, next release, fresh scoped, fresh non-scoped). Details in the commit message.

PR #921, opened 2026-04-26
ci: stop swallowing real errors across the workflow tree

robcohen:
## Why

Pass over every workflow file removing or hardening `|| true` / `2>/dev/null` / `continue-on-error` patterns that masked real failures.

The pre-existing patterns were a mix of:
1. Genuine bugs hiding broken behavior — **fixed**.
2. Idempotent-cleanup constructs that look identical — **kept**.
3. GNU/BSD compat fallbacks (`stat -c%s … || stat -f%z …`) — **kept**.
4. Explicit conditional probes (`if foo 2>/dev/null; then`) — **kept**.

This change targets only category 1.

## Real bugs fixed

| File | Pattern | Why it was wrong |
|------|---------|------------------|
| `release-publish.yml` | `copr-cli build-package … --nowait \|\| true` | Trigger failures silently swallowed. COPR keeps building previous version with no signal. |
| `release-publish.yml` | `continue-on-error: true` on Homebrew bump | Masked all failures, not just legitimate "PR already exists" — auth/network errors looked identical to success. |
| `compat.yml` | `bean-example … 2>/dev/null \|\| true` | Silent failures shrank the synthetic test corpus → false-green compat metrics. |
| `compat.yml` | `cp …badge files… 2>/dev/null \|\| true` (×2 locations) | Silent failures left badges stale. Replaced with `[ -f ]` guard + unsuppressed `cp`. |
| `compat.yml` | `pip install third-party-plugins \|\| true` | Silent install failures meant compat tests using those plugins got skipped. |
| `tla.yml` | `python3 trace_to_rust_test.py … 2>/dev/null \|\| true` | Masked script bugs and produced empty test files. Replaced with `if … then ... else ::warning::` for unsupported specs. |
| `kani.yml` | `cargo kani … \| tee … \|\| true` | Verification FAILUREs hidden — only appeared in summary while job stayed green. Removed `\|\| true`, added `set -o pipefail`, made Summary `if: always()` so a failed run still produces the human-readable summary. |
| `release-test.yml` | `… \| head -1 \|\| true` | `\|\| true` was a no-op (`head` always exits 0); removed and added a comment that the `2>/dev/null` is intentional polling-noise suppression, not error swallowing. |
| `flake.nix` | `tlc -help 2>/dev/null \| head -1 \|\| echo 'not available'` | `tlc -help`'s first stdout line is blank, so the devshell banner showed `TLA+:` with no value. Replaced with an explicit `command -v tlc` check + `grep -i version`. |

## Kept (legitimate)

- `ci.yml` `Install nextest` `continue-on-error` — paired with a `cargo install` fallback step keyed off `steps.install-nextest.outcome`.
- `wasm.yml` `test-browser` `continue-on-error` — job is `if: false`.
- `release-test.yml` `brew untap … \|\| true` — idempotent cleanup; brew untap on something not tapped is a normal failure mode.
- `npm view` / `dnf info` / `cargo search` polling loops with explicit `exit 1` after timeout — transient flakes are the point of the loop.
- Orphan-branch publish patterns in `bench.yml` / `compat.yml` — `git fetch || true` etc. are intentional setup for "branch may not exist yet" first runs.
- GNU/BSD fallbacks like `stat -c%s … || stat -f%z …`.
- Conditional probes `if curl … 2>/dev/null; then`.
- `curl … || echo '[]' > history.json` — empty JSON array is the valid "no history yet" sentinel.
- Diagnostic `tail … || true` writing to `$GITHUB_STEP_SUMMARY` (fuzz output may not exist on a clean run).

## Test plan

- [x] All modified YAML still parses (`python -c "import yaml; yaml.safe_load(...)"`)
- [x] Pre-push hook (`cargo check --workspace --all-features --all-targets`) passes
- [x] TLA+ devshell banner now shows `TLA+: TLC - provides model checking … Version 2.19 of 08 August 2024` instead of blank
- [ ] CI on this PR will exercise compat.yml; if any of the third-party `pip install` packages have routinely flaked we'll see it here and can decide whether to harden with retries
- [ ] Next compat run will exercise the cleaned-up `bean-example` and badge `cp` paths

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -48,13 +48,14 @@ jobs:
       - name: Install dependencies
         run: |
           pip install beancount beanquery matplotlib pyyaml
-          # Third-party plugins used by test files
-          pip install beanahead tariochbctools || true
-          # beancount-share requires beancount_plugin_utils as a dependency
-          pip install beancount_plugin_utils beancount-share || true
-          # Install reds-plugins from GitHub to get unreleased plugins like box_accrual
-          pip install git+https://github.com/redstreet/beancount_reds_plugins || true
-          pip install git+https://github.com/Evernight/beancount-lazy-plugins || true
+          # Third-party plugins used by test files. Compat tests rely on
+          # them — a silent install failure means tests using these
+          # plugins get skipped and we get false-green CI. Fail loudly
+          # if any of these can't install; re-run on transient PyPI flakes.
+          pip install beanahead tariochbctools
+          pip install beancount_plugin_utils beancount-share
+          pip install git+https://github.com/redstreet/beancount_reds_plugins
+          pip install git+https://github.com/Evernight/beancount-lazy-plugins
       # Cache downloaded test files to avoid re-cloning repos every run
       - name: Cache compatibility test files
         id: cache-compat-files
@@ -94,7 +95,10 @@ jobs:
               end_date=$(date +%Y-%m-%d)
               start_date=$(date -d "$end_date - 1 year" +%Y-%m-%d 2>/dev/null || date -v-1y +%Y-%m-%d)
               output="tests/compatibility/synthetic/bean-example/example_seed${seed}.beancount"
-              bean-example --seed "$seed" --date-begin "$start_date" --date-end "$end_date" --output "$output" 2>/dev/null || true
+              # Fail loudly on bean-example errors. Previously we swallowed
+              # them, which silently shrank the synthetic test corpus and
+              # gave false confidence in compat metrics.
+              bean-example --seed "$seed" --date-begin "$start_date" --date-end "$end_date" --output "$output"
             done
             echo "Generated $(find tests/compatibility/synthetic/bean-example -name '*.beancount' 2>/dev/null | wc -l) bean-example files"
           fi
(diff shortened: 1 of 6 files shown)


robcohen:
All three Copilot inline comments addressed in d7540373 — all three were correct:

**1. release-test.yml polling loop (line 101):**

Copilot's diagnosis was right. GitHub Actions default bash on Linux is `bash --noprofile --norc -eo pipefail {0}` — pipefail IS on, contrary to what I assumed in the original commit. So my removal of `|| true` from `cargo search ... | grep ... | head -1` would have killed the polling loop on iteration 1 whenever the version wasn't yet visible (grep exits 1 → pipe exits 1 → `set -e` kills the script). Captured `cargo search` separately and restored `|| true` only on the grep step where no-match is the actual polling sentinel, not an error.

**2. tla.yml trace-to-test step (line 230):**

Copilot caught that `scripts/trace_to_rust_test.py` expects JSON input produced by a `scripts/tla_trace_to_json.py` converter — and that converter does not exist in this repo. So the entire 'Generate Rust tests from traces' step was structurally broken; the previous `2>/dev/null || true` had been masking a reliable script failure for who knows how long. My follow-up `if … then … else ::warning::` would have just made the always-failing path produce `::warning::` annotations on every TLA+ failure, which is noise without value.

Removed the dead step entirely. Raw .trace files are still uploaded as an artifact for offline analysis until someone ports the converter.

**3. compat.yml unpinned VCS deps (line 58):**

Right concern — `pip install git+https://...` tracks HEAD by default and turns unrelated upstream commits into spurious red CI. Pinned both to current HEAD SHAs:

- redstreet/beancount_reds_plugins @ `41af12d` (also tag `0.4.0`)
- Evernight/beancount-lazy-plugins @ `1a685ed` (no tags)

(Copilot's suggested SHAs were placeholders, but the principle was correct.)

PR #920, opened 2026-04-26
docs(release): rewrite RELEASING.md for manual flow + drop release-plz

robcohen:
## Summary

Replaces the release-plz-driven release process with a documented manual flow: bump versions → open `chore: release vX.Y.Z` PR → merge → `gh release create` triggers build/publish.

Removes all dead release-plz references throughout the repo. The Mergify auto-merge / Semver Check cleanup (#919) and the npm monotonicity guard (#918) already landed; this is the last piece of the release-pipeline cleanup arc.

## Changes

| File | Change |
|------|--------|
| `RELEASING.md` | Full rewrite: documents manual bump → PR → release flow. Keeps useful sections (binary matrix, package channels, trusted publishing, troubleshooting) and references the #918 npm guard. |
| `.github/workflows/release-plz.yml` | **Deleted** |
| `release-plz.toml` | **Deleted** |
| `.github/workflows/ci.yml` | `release-plz-*` branch detector → `release/*` (new convention) |
| `.github/workflows/aur.yml` | Same branch detector switched + comment updated |
| `.github/workflows/branch-name.yml` | Dropped the `release-plz-*` bypass; the validator's general pattern already accepts `release/*` |
| `.github/workflows/release-build.yml` | Version-mismatch error message no longer instructs people to run `release-plz set-version`; points at RELEASING.md instead |

## Test plan

- [x] `python -c "import yaml; yaml.safe_load(...)"` on every modified workflow
- [x] `cargo check --workspace --all-features --all-targets` (pre-push hook)
- [x] No remaining grep hits for `release.?plz` outside the deliberate mention in RELEASING.md
- [ ] CI on this PR will exercise `branch-name.yml` and `ci.yml` against a `chore/*` branch — confirms the validator still accepts the existing patterns

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #927, opened 2026-04-27
chore: release v0.14.1

robcohen:
Patch release to redrive the v0.14.0 publish flow with the mcp-server TypeScript fix from #926 included in the tagged source tree.

## Why a patch

The v0.14.0 tag at `d655dbb8` pre-dates the type-fix from #926. `release-publish.yml`'s mcp job checks out the tag, so re-triggering against `tag=v0.14.0` kept hitting the same `tsc` errors. v0.14.1 captures all the fixes and lets the publish flow run end-to-end.

## What's in the bump

Standard surfaces from RELEASING.md:

- Workspace `[workspace.package].version` + 10 `[workspace.dependencies]` pins
- 14 crate `Cargo.toml`s
- `packages/mcp-server/package.json` (`version` + `@rustledger/wasm` dep)
- `packaging/rpm/rustledger.spec` (`Version`, `Source0`, `%setup`)
- `Cargo.lock` (auto-regenerated)

No code changes beyond version strings — the actual fixes (#924 ops publish order, #926 mcp-server TS types) already landed on main as part of unblocking v0.14.0.

## Expected publish state after merge + `gh release create v0.14.1`

| Channel | Expected |
|---|---|
| crates.io (all 14) | 0.14.1 (skip-if-published guard now honors trusted publishing for ops) |
| `@rustledger/wasm` | 0.14.1 (monotonicity guard from #918 will accept it) |
| `@rustledger/mcp-server` | **0.14.1** ← the actual goal of this release |
| Docker, AUR, Scoop, COPR, Homebrew | 0.14.1 |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3283,7 +3283,7 @@ dependencies = [
 
 [[package]]
 name = "rustledger"
-version = "0.14.0"
+version = "0.14.1"
 dependencies = [
  "anyhow",
  "clap",
@@ -3323,7 +3323,7 @@ dependencies = [
 
 [[package]]
 name = "rustledger-booking"
-version = "0.14.0"
+version = "0.14.1"
 dependencies = [
  "bigdecimal",
  "proptest",
@@ -3336,7 +3336,7 @@ dependencies = [
 
 [[package]]
 name = "rustledger-core"
-version = "0.14.0"
+version = "0.14.1"
 dependencies = [
  "criterion",
  "jiff",
@@ -3353,7 +3353,7 @@ dependencies = [
 
 [[package]]
 name = "rustledger-ffi-wasi"
-version = "0.14.0"
+version = "0.14.1"
 dependencies = [
  "rustc-hash",
  "rustledger-booking",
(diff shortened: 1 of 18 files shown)


robcohen:
Both Copilot inline comments addressed.

**1. `packaging/rpm/rustledger.spec` (`BuildRequires: rust >= 1.75`)** — fixed in 217c8803. The spec had been pinned at 1.75 since well before the workspace migrated to Edition 2024 (which stabilized in 1.85) and bumped MSRV to 1.94. Bumped to `>= 1.94` to match `workspace.package.rust-version`. Real bug — COPR builds against 1.75 would fail at parse time on edition2024 syntax.

**2. `packages/mcp-server/package-lock.json` referencing `@rustledger/wasm@0.14.0`** — acknowledged as a documented limitation, not fixing in this PR. Per RELEASING.md:

> **Don't try to update `package-lock.json`** — `@rustledger/wasm@X.Y.Z` doesn't exist on npm yet during the bump PR, so `npm install` fails with `ETARGET`. The publish workflow regenerates the lockfile after wasm is published.

That guidance was added in PR #920 specifically because of this chicken-and-egg. The publish workflow's `npm install` step regenerates the lockfile in CI but doesn't commit it back, so the repo lockfile stays slightly stale between releases. Local `npm ci` would fail until a follow-up cleanup.

A fully clean fix would be to refresh the lockfile post-publish (after `@rustledger/wasm@0.14.1` exists), but that's out of scope for this release PR. Filing it as a follow-up.

Issue #929, opened 2026-04-27
[Compat]:  balance queries mismatch


brunotvs:
### Description

rledger and python bean-query results diverge for balances. Unable to use aggregations like ```convert(last(balance), 'USD')```

### Beancount File

```text
bean-example >> ledger.beancout
```

### Python Beancount Output

```text
bean-query ledger.beancount "SELECT year, month, convert(last(balance), 'USD') where account ~ '^(Assets|Liabilites)' ORDER BY year, month"
```

### Rustledger Output

```text
rledger query ledger.beancount "SELECT year, month, convert(last(balance), 'USD') where account ~ '^(Assets|Liabilites)' ORDER BY year, month"
```

### Python Beancount Version

3.2.0

### Rustledger Version

0.14.0

robcohen:
Reproduced and fixed in #940.

**Root cause.** rledger's `balance` column was the per-account running balance (bank-statement-style), but BQL's `balance` is a single cumulative Inventory across *all* WHERE-filtered postings in iteration order. That's [documented behavior](https://beancount.github.io/docs/beancount_query_language/) and [load-bearing in Blais's original 2014 design](https://groups.google.com/g/beancount/c/X-IKy7nvtVA) — it's what lets one query language render both journals and balance reports.

**Reproduction** (anyone can run this):

```bash
bean-example > /tmp/ledger.beancount
QUERY="SELECT year, month, convert(last(balance), 'USD') where account ~ '^(Assets|Liabilites)' ORDER BY year, month"
bean-query  /tmp/ledger.beancount "$QUERY"
rledger query /tmp/ledger.beancount "$QUERY"
```

End of January 2024:
- bean-query: `16100.00 IRAUSD   7952.43 USD   10 VACHR`  ← end-of-month portfolio
- rledger (before #940): `-0.02 USD`  ← whichever single account's last posting won the tie

Every one of the 30 rows disagreed.

**Fix in #940.**
- `balance` is now cumulative across WHERE-filtered postings — matches bean-query row-for-row on the same `bean-example` file.
- A new `account_balance` column exposes the old per-account view for users who want bank-statement semantics. Existing rledger queries that depended on `balance` being per-account need a one-word rename to `account_balance`.

Thanks for the report — this also surfaced a bigger gap in our compat testing. The current "100% BQL compatibility" check only verifies that queries *parse and execute*, not that they return the same rows as bean-query. Adding row-level golden diffs is a separate followup.

PR #941, opened 2026-04-28
ci(compat): expand BQL corpus + extract harness to script with row-level diagnostics

robcohen:
## Summary

Two things in one PR:

1. **Expand the BQL compat corpus** from 3 to 13 queries, weighted toward columns/functions where the two tools have historically diverged in semantics. The previous corpus had no query that touched `balance`, `position`, `weight`, `convert`, `LAST`, or `FIRST` — which is how #929 silently shipped through two releases with the badge reporting 100% BQL match.

2. **Extract the harness from inline-in-YAML to a real script** (`scripts/compat-bql-test.py` + corpus in `tests/compatibility/bql-queries.toml`) and add five signal-quality fixes that the inline version couldn't easily express.

## Signal-quality fixes

| Fix | What it catches |
|---|---|
| **Order-aware matching** | Queries with `ORDER BY` now compare row sequences, not sorted sets. Catches ordering bugs the old `sorted()` matching hid. |
| **Empty-result tracking** | Per-query coverage rate. Queries returning 0 rows on >50% of files get flagged — that's a corpus gap, not a test failure. |
| **Multi-row mismatch detail** | Up to 3 differing rows + row-count diff per mismatch. Old harness logged 1 row, making a wholesale-wrong bug look like a small mismatch. |
| **Corpus-size floor** | `MIN_CORPUS_SIZE = 13`. Script aborts if the corpus shrinks. Prevents silent regression. |
| **`BALANCES` + `JOURNAL` forms** | Different code paths from `SELECT` in both tools. Previously untested. |

## Smoke test (local)

Ran the script against `bean-example` output. Result: **9 of 13 queries report mismatches**, including the #929 reproduction. The old 3-query corpus reported 100% match on the same input file.

The mismatches break down into:
- 1 expected (the #929 `balance` semantic — will start matching once #940 lands)
- 2 cost-rendering bugs (`{42.64 USD}` vs `{ 42.64 USD}` — leading whitespace) — separate followups
- 1 missing cost annotation in rledger output — likely separate followup
- Other formatting/precision items worth filing as issues

## Followups (out of scope for this PR)

- Structured value comparison (parse decimals/inventories before diffing) so we can tolerate display-layer differences without losing semantic signal.
- File the cost-rendering whitespace bug observed during smoke testing.

## Test plan

- [x] Script smoke-tests cleanly against `bean-example` (13/13 queries run, mismatches reported with row-level detail)
- [x] Corpus-size floor catches accidental query removal
- [ ] CI run shows the workflow YAML is now ~200 lines smaller and calls the script
- [ ] CI run picks up the new corpus and reports a real BQL match % (likely lower than 100% — that's the point)

Closes none directly; supports the row-diff infrastructure goal mentioned in #929.


robcohen:
Closing — branch name didn't match repo policy (`ci/` prefix not in allowed list). Reopened with the same content under `chore/expand-bql-compat-corpus` in #TBD.

robcohen:
Reopened as #942 (same diff, valid `chore/` branch prefix).

PR #1075, opened 2026-05-11
fix: box BQL expression parser to avoid Apple linker crash

sciyoshi:
## Summary
- Box the Chumsky BQL expression parser and major precedence layers to prevent enormous concrete parser types from producing multi-megabyte Rust symbols
- Keep parser behavior unchanged while allowing Apple ld to link the CLI
- Avoid an unused_mut warning when python-plugin-wasm is disabled

## Verification
- cargo build -p rustledger --no-default-features
- cargo build -p rustledger --all-features
- cargo test -p rustledger-query


robcohen:
## Deep PR review

Thanks for the fix! Read through both changes against current `main`. Verdict: **correct and well-targeted**, with a few small notes.

### Change 1 — Box the BQL expression parser ✅

**Why this works.** `expr()` chains 7 precedence layers (`primary → unary → multiplicative → additive → comparison → NOT → AND → OR`) plus a `recursive(...)` wrapper. Without `.boxed()`, each layer's concrete chumsky type embeds the previous layer's type, producing a fully-monomorphized type expression that grows roughly as `T_n = Foldl<Then<T_{n-1}, ...>, ...>`. By the OR layer, the type name's mangled symbol can exceed Apple `ld`'s symbol-length limits (~4MB on Darwin). This is a known shape of crash in chumsky parsers and the `.boxed()` solution is the documented fix.

**Audit.** Every layer that grows monomorphically gets a `.boxed()`:
- `primary_expr(expr.clone()).boxed()` ✓
- `unary.boxed()` ✓
- `multiplicative.boxed()` ✓
- `additive.boxed()` ✓
- `comparison.boxed()` ✓
- `not_expr.boxed()` ✓
- `and_expr.boxed()` ✓
- outer `recursive(...).boxed()` ✓

8 boxing sites, consistent with the 7 internal layers + the recursive return. The signature change from `impl Parser<...> + Clone` to `Boxed<'a, 'a, ParserInput<'a>, Expr, ParserExtra<'a>>` is the only public-API change; `Boxed` implements `Parser` and `Clone`, so call sites in `query.rs` should not need adjustment.

**Semantic equivalence.** `.boxed()` is type erasure only — it replaces a concrete `impl Parser<...>` with a `Boxed` that internally holds a `Rc<dyn Parser<...>>`. Parser behavior is unchanged (verified by the existing `cargo test -p rustledger-query` suite). I don't see any chumsky semantic that differs between boxed and unboxed parsers.

**Cost trade-off.**
- **Compile time**: significantly faster for `rustledger-query` and its downstream `rustledger` binary, especially on Apple Silicon where the linker is the bottleneck. Worth measuring before/after `cargo build -p rustledger-query` in the PR description — that's the headline benefit.
- **Binary size**: should shrink slightly (less monomorphization).
- **Parse-time runtime**: one vtable indirection per boxing layer per token consumed. For BQL — short query strings parsed once per shell invocation or once per query — completely unmeasurable.

### Change 2 — `unused_mut` warning fix ✅

Cleaner alternative to `#[allow(unused_mut)]`. The shadow pattern reads:

```rust
let warning_count = ledger.errors.iter().filter(...).count();
#[cfg(feature = "python-plugin-wasm")]
let mut warning_count = warning_count;
```

This makes mutability explicitly conditional, which is honest about the intent. Possible nit: `#[cfg_attr(not(feature = "python-plugin-wasm"), allow(unused_mut))] let mut warning_count = ...` would be a single binding instead of a shadow. Either is fine; current approach is slightly more visible at the use site.

### Notes (none blocking)

1. **Two unrelated changes**. Boxing fixes a linker crash; the `unused_mut` fix is collateral damage from a particular feature combo. They could be separate PRs, but both are small and the build break probably occurs together on macOS-with-default-features. Not worth resplitting.

2. **No repro / issue link**. Body doesn't reference an open issue or include the exact `ld` error. A `clang: error: linker command failed` excerpt + "macOS xx.y, rustc xx.yy.z" in the body would help future archeology when this gets re-broken by chumsky upgrades. (And it would be worth searching the issues — if there isn't one, this PR effectively closes a latent bug that hadn't been filed.)

3. **No new tests**. Correct call — `.boxed()` is mechanical and any regression would surface in the existing BQL parse-test matrix. Adding "test boxed parser still parses X" tests would be cargo-culting.

4. **CLA pending** (procedural). Once signed, the full CI matrix runs and should be green.

### Optional follow-up

Worth adding a short comment above `expr()` explaining the boxing choice:

```rust
/// Parse an expression (with precedence climbing).
///
/// The precedence layers are individually `.boxed()` to avoid Apple `ld`
/// failures on the deeply nested concrete chumsky type — without
/// type erasure, the mangled symbol exceeds Darwin's symbol length
/// limit. See PR #1075.
#[allow(clippy::large_stack_frames)]
fn expr<'a>() -> Boxed<'a, 'a, ParserInput<'a>, Expr, ParserExtra<'a>> {
```

This way the next person looking at this won't "optimize" the boxing away.

### Recommendation

LGTM. Ready to merge once CLA is signed and CI confirms green.

robcohen:
Hi @sciyoshi — thanks again for the fix. One blocker on our side: we can't merge without a signed CLA. The [CLA Assistant](https://cla-assistant.io/rustledger/rustledger?pullRequest=1075) bot already left a comment with the link to sign — once you do, CI will run and (assuming green) we can proceed with merge.

If you'd prefer not to sign, no hard feelings — please open a GitHub issue describing the Apple linker crash (the exact `ld` error output, your macOS + rustc versions, and the build invocation that triggered it). We'll take it from there and implement the fix ourselves on a separate branch.

Thanks again for taking the time to track this down.

robcohen:
Hey @sciyoshi — the parser-boxing change itself looks good and all the functional checks (Test, Check, Clippy, Doctests, Regression, MSRV, etc., 24 of 33) pass. The 6 failures break down like this:

## Real, author-side (please fix)

1. **Validate branch name** — the policy requires `<type>/<description>` with a slash. Your branch is `fix-apple-ld-query-parser` (hyphen). Quickest fix:
   ```bash
   git branch -m fix-apple-ld-query-parser fix/apple-ld-query-parser
   git push origin -u fix/apple-ld-query-parser
   git push origin --delete fix-apple-ld-query-parser
   gh pr edit 1075 --head fix/apple-ld-query-parser  # or close and reopen against new branch
   ```
   (Or just rename in GitHub's UI if simpler.)

2. **Format** — `cargo fmt --check` finds diffs at `crates/rustledger-query/src/parser.rs:533` and `:648`. Please run `cargo fmt --all` and push.

## Aggregator (will auto-resolve)

3. **build** — meta-job that aggregates the matrix; failed only because `Format` failed. Fixes itself once the `cargo fmt` issue above is fixed.

## Repo CI infrastructure, **not your fault** (maintainer to address separately)

4. **Gitleaks Secret Scan** — `missing gitleaks license`. gitleaks-action started requiring a paid license; the repo's `GITLEAKS_LICENSE` secret isn't set. This fails on every PR right now, not just yours.
5. **PR Benchmark** — `HttpError: Resource not accessible by integration` when posting an issue comment. Because your PR is from a fork, `GITHUB_TOKEN` is read-only by design, so the workflow can't post its benchmark comment. The workflow treats that as fatal instead of best-effort.
6. **Run compatibility tests** — same fork-token comment-posting failure as item 5.

None of items 4–6 reflect a problem with your code; they're CI configuration issues we'll fix on the repo side. Once you sort out the branch rename and `cargo fmt`, the only failures left will be the three infra ones, and they shouldn't block the merge.

Thanks for the fix!

robcohen:
Thanks for the fix, @sciyoshi. Re-opened your work as #1097 with the conformant branch name (`fix/apple-ld-query-parser` — slash, not hyphen, per the repo's Validate-branch-name policy) and `cargo fmt` applied so CI runs cleanly. Your authorship is preserved on the cherry-picked commit.

Both unrelated infrastructure failures we hit on this PR — the gitleaks-license issue (#1088) and the fork-PR comment-posting failures (#1093) — have since been resolved on main via #1094 and #1096, so #1097 should sail through CI without the spurious red Xs you saw here.

Closing this one as superseded.

PR #1076, opened 2026-05-11
perf: trivial quick wins bundle (blake3 cache, parking_lot, FxHasher, wasm fat-LTO)

### LINKED ISSUE #1065: perf: trivial quick wins (blake3 cache, parking_lot, FxHasher in BQL, wasm fat-LTO)
A bundle of four trivial swaps surfaced by an audit of the perf surface area on `main`. Each is a one-to-few-line change with an existing workspace dependency already pulled in. Filing as one issue so they ship together as a single "perf-quick-wins" PR.

## 1. Cache: `sha2::Sha256` → `blake3`

`blake3 = "1"` is already in workspace deps (`Cargo.toml:104`), but `rustledger-loader::cache` still uses `sha2::Sha256`:

- `crates/rustledger-loader/src/cache.rs:36` — `use sha2::{Digest, Sha256};`
- `crates/rustledger-loader/src/cache.rs:256` — `let mut hasher = Sha256::new();`

Blake3 is 10–20× faster than SHA-256 in software. The cache validity check hashes file metadata (mtime+size) so the per-call cost is small, but free wins are free wins.

**Fix:** swap to `blake3::Hasher::new()` and remove `sha2` from `crates/rustledger-loader/Cargo.toml:42` + the `cache` feature deps at line 19.

## 2. `SyncStringInterner`: `std::sync::Mutex` → `parking_lot::Mutex`

`parking_lot = "0.12"` is already in the workspace (`Cargo.toml:139`). The sync interner still uses the OS mutex:

- `crates/rustledger-core/src/intern.rs:398` — `inner: std::sync::Mutex<StringInterner>,`
- `crates/rustledger-core/src/intern.rs:405` — `inner: std::sync::Mutex::new(StringInterner::new()),`

`parking_lot::Mutex` avoids OS-level syscalls on uncontended/contended access alike, and the API surface is `lock()` → `MutexGuard` so the call sites should be drop-in.

## 3. BQL `hash_row` / `hash_single_value`: `DefaultHasher` → `FxHasher`

`rustc-hash` is already pulled in (used by `FxHashMap` / `FxHashSet` everywhere in this crate). The two BQL grouping helpers are the exceptions:

- `crates/rustledger-query/src/executor/types.rs:213-218`
- `crates/rustledger-query/src/executor/types.rs:223-228`

Both build a `std::collections::hash_map::DefaultHasher` (SipHash 1-3, cryptographic strength). GROUP BY / DISTINCT keys don't need DoS-resistance; they need speed. Swap to `rustc_hash::FxHasher`.

## 4. `wasm-release` profile: add `lto = "fat"`

`[profile.wasm-release]` at `Cargo.toml:205-208` inherits from `[profile.release]` (which is `lto = "thin"` to dodge the Apple linker bug) and never overrides. WASM doesn't have the Apple linker concern, so fat LTO is free:

```toml
[profile.wasm-release]
inherits = "release"
lto = "fat"
opt-level = 3
panic = 'abort'
```

## Expected impact

None of these moves a benchmark dramatically on its own; together they're a "no reason not to" cleanup. Effort: maybe 30 minutes including running the test suites.

## Audit context

These are items 5a, 2c, 7b, 10c from a recent perf audit. Larger items from the same audit are filed as separate issues — see "perf:" issues opened alongside this one.

robcohen:
## Summary

Closes #1065.

Four no-architecture, mostly-one-line swaps. Each is too small to merit its own PR; bundled here so they ship together as a single "perf-quick-wins" change.

| # | Change | Files |
|---|---|---|
| 1 | Cache hash: `sha2::Sha256` → `blake3::Hasher` | `crates/rustledger-loader/src/cache.rs`, `crates/rustledger-loader/Cargo.toml` |
| 2 | `SyncStringInterner`: `std::sync::Mutex` → `parking_lot::Mutex` | `crates/rustledger-core/src/intern.rs`, `crates/rustledger-core/Cargo.toml` |
| 3 | BQL `hash_row` / `hash_single_value`: `DefaultHasher` → `FxHasher` | `crates/rustledger-query/src/executor/types.rs` |
| 4 | `wasm-release` profile: add `lto = "fat"` override | `Cargo.toml` |

### Why bundled?

Each item is < 10 lines of meaningful change and uses dependencies already in the workspace (`blake3 = "1"`, `parking_lot = "0.12"`, `rustc-hash`). Bundling avoids 4 separate PR review cycles for what amounts to "swap the import."

### Notes

- `sha2` stays in the workspace because other crates use it for non-cache purposes (`synthetic/manifest`, `ffi-wasi`, `wasm`, `plugin`). Only `rustledger-loader` drops the dep.
- `parking_lot::Mutex` is non-poisoning, so the call sites simplify from `.lock().unwrap()` to `.lock()`. No semantic change — none of the previous unwraps would have triggered (the mutex is short-lived around interner ops).
- The cache binary format already used a 32-byte hash, so BLAKE3 (also 32 bytes) is a drop-in replacement at the struct level. Any cache file written before this PR will be invalidated on first load (hash mismatch → re-parse), which is the same as any other cache-format change.

## Test plan

- [x] `cargo test -p rustledger-core -p rustledger-loader -p rustledger-query --all-features` — all green
- [x] `cargo clippy --all-features -- -D warnings` — clean
- [ ] CI checks (compatibility, regression, MSRV)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3363,6 +3363,7 @@ version = "0.14.1"
 dependencies = [
  "criterion",
  "jiff",
+ "parking_lot",
  "proptest",
  "rkyv 0.8.16",
  "rust_decimal",
@@ -3414,6 +3415,7 @@ dependencies = [
 name = "rustledger-loader"
 version = "0.14.1"
 dependencies = [
+ "blake3",
  "glob",
  "jiff",
  "rayon",
@@ -3427,7 +3429,6 @@ dependencies = [
  "rustledger-plugin",
  "rustledger-plugin-types",
  "rustledger-validate",
- "sha2 0.11.0",
  "tempfile",
  "thiserror 2.0.18",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -204,6 +204,7 @@ lto = "fat"
 # WASM-optimized profile (use with: cargo build --profile wasm-release)
 [profile.wasm-release]
 inherits = "release"
+lto = "fat"          # Override `release`'s "thin" — WASM doesn't hit the Apple linker bug
 opt-level = 3        # Speed over size (~21% larger, ~60% faster)
 panic = 'abort'      # Smaller binary, no unwinding
 
--- a/crates/rustledger-core/Cargo.toml
+++ b/crates/rustledger-core/Cargo.toml
@@ -29,6 +29,7 @@ serde.workspace = true
 serde_json.workspace = true
 rkyv = { workspace = true, optional = true }
 rustc-hash.workspace = true
+parking_lot.workspace = true
 
 [dev-dependencies]
 proptest.workspace = true
--- a/crates/rustledger-core/src/intern.rs
+++ b/crates/rustledger-core/src/intern.rs
@@ -392,33 +392,36 @@ impl CurrencyInterner {
 
 /// Thread-safe string interner using a mutex.
 ///
-/// Use this when interning strings from multiple threads.
+/// Use this when interning strings from multiple threads. Backed by
+/// `parking_lot::Mutex` which avoids the OS-level syscall that
+/// `std::sync::Mutex` requires on lock/unlock and is non-poisoning, so
+/// `lock()` returns the guard directly without an `.unwrap()`.
 #[derive(Debug, Default)]
 pub struct SyncStringInterner {
-    inner: std::sync::Mutex<StringInterner>,
+    inner: parking_lot::Mutex<StringInterner>,
 }
 
 impl SyncStringInterner {
     /// Create a new thread-safe interner.
     pub fn new() -> Self {
         Self {
-            inner: std::sync::Mutex::new(StringInterner::new()),
+            inner: parking_lot::Mutex::new(StringInterner::new()),
         }
     }
 
     /// Intern a string (thread-safe).
     pub fn intern(&self, s: &str) -> InternedStr {
-        self.inner.lock().unwrap().intern(s)
+        self.inner.lock().intern(s)
     }
 
     /// Get the number of unique strings.
     pub fn len(&self) -> usize {
-        self.inner.lock().unwrap().len()
+        self.inner.lock().len()
     }
 
     /// Check if empty.
     pub fn is_empty(&self) -> bool {
-        self.inner.lock().unwrap().is_empty()
+        self.inner.lock().is_empty()
     }
 }
 
--- a/crates/rustledger-loader/Cargo.toml
+++ b/crates/rustledger-loader/Cargo.toml
@@ -16,7 +16,7 @@ readme = "README.md"
 default = ["cache", "full"]
 
 # Enable rkyv-based binary caching for faster loads
-cache = ["dep:rkyv", "dep:sha2", "rustledger-parser/rkyv"]
+cache = ["dep:rkyv", "dep:blake3", "rustledger-parser/rkyv"]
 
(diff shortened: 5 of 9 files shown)


robcohen:
## Deep PR review — self + Copilot follow-ups

Pushed `5c4220b6` addressing both Copilot inline findings. Posting my own deep review for the record.

### Copilot findings — both addressed

1. **`cache.rs:209` — bump `CACHE_VERSION`.** ✅ 3 → 4. Old caches now fail the cheap header version check at `cache.rs:397` instead of paying the rkyv deserialize cost only to fail the hash compare. The header layout is unchanged (both SHA-256 and BLAKE3 are 32-byte outputs, so `CacheHeader::SIZE == 52` either way), so this is a soft invalidation — old cache files get rejected on first load after upgrade and re-written on save.
2. **`types.rs:219,229` — redundant `Hasher` imports.** ✅ Dropped both inner `use std::hash::Hasher;` — confirmed `Hasher` is already in scope at `types.rs:4` via `use std::hash::{Hash, Hasher};`.

### Self-review findings (all OK)

**Correctness**
- ✅ BLAKE3 `Hasher::update(&[u8])` differs from sha2's `update<T: AsRef<[u8]>>` — needed to add `&` prefix on the `to_le_bytes()` calls. Done at `cache.rs:267-272`.
- ✅ BLAKE3's `finalize()` returns `Hash` (not `Output<Sha256>`). Used `*hasher.finalize().as_bytes()` to dereference the `&[u8; 32]` into the owned `[u8; 32]` return type.
- ✅ `parking_lot::Mutex::lock()` returns `MutexGuard<T>` directly (not `LockResult`) — non-poisoning. All four call sites in `SyncStringInterner` simplified from `.lock().unwrap().X()` to `.lock().X()`. No semantic change; any code that hit the `.unwrap()` would have already panicked.
- ✅ `FxHasher` produces different hash values than `DefaultHasher` (SipHash 1-3), but the BQL hashes are internal dedup tokens — not persisted, not cross-process, not part of any wire format. The change is transparent to callers.
- ✅ `wasm-release` `lto = "fat"` override placed after `inherits = "release"` so it wins.

**Cross-platform safety**
- `parking_lot` works on wasm32 (CI's "Build Package" and "Test (Node.js)" both pass). No new wasm bloat — `parking_lot` was already a transitive dep via other workspace members; this just makes it a direct dep of `rustledger-core`.
- `blake3` has a portable software implementation for wasm32. The CI compatibility job passed end-to-end on Linux, and the WASM package builds cleanly.

**Scope hygiene**
- ✅ `sha2` was kept in the workspace deps — other crates use it for non-cache purposes (`synthetic/manifest`, `ffi-wasi`, `wasm`, `plugin`). Only `rustledger-loader` drops the `sha2` dep.
- ✅ Doc comment on `SyncStringInterner` now explicitly mentions the perf rationale + the non-poisoning consequence at call sites. Useful for future-me wondering why there's no `.unwrap()`.
- ✅ All four changes are independent and self-contained — bundling them in one PR for review efficiency, but each could be reverted in isolation without affecting the others.

### CI

All checks green on the previous commit (35 passed). The fix commit triggers another CI run; nothing in the fix should regress.

### Not addressed (won't-do for this PR)

- Removing `sha2` from the workspace — out of scope; other crates legitimately need it.
- Pre-allocating BLAKE3 hasher state — micro-optimization, not measurable for the cache use case.

Ready for merge once CI re-greens.

PR #1197, opened 2026-05-26
fix(booking): respect NONE booking method during reduction (closes #1182)

### LINKED ISSUE #1182: [Bug]: NONE booking type not respected in reduction.
### Description

I've just tested out a large beancount history file and have discovered many errors on check due to my using NONE booking type having a bug. (It does not respect the booking type on reduction.)

There is a native alternative approach in rustledger but I'll be opening a discussion on that. (The difference between compatibility and adoption of rustledger)

An example that works fine with beancount but not with rustledger (and causes many same challenges in my books) is attached.

### Steps to Reproduce

Run rledger check on the example ledger and run beancount check.

### Expected Behavior

It accumulates bookings just fine.

### Minimal Beancount File

```text
option "booking_method" "NONE"

2000-01-01 open Assets:Cash
2000-01-01 open Assets:Broker

2024-11-01 * "Broker" "Buy"
  Assets:Broker  10 X {{ 100 GBP }}
  Assets:Cash  -100 GBP

2024-11-01 * "Broker" "Buy"
  Assets:Broker  10 X {{ 120 GBP }}
  Assets:Cash  -120 GBP

2024-11-01 * "Broker" "Sell"
  Assets:Broker  -15 X {{ 210 GBP }}
  Assets:Cash  210 GBP
```

### Version

0.15.0

### Operating System

macOS

### Additional Context

_No response_

robcohen:
## Summary

`option "booking_method" "NONE"` was ignored on the reduction path. Every sign-flipping cost-bearing posting was treated as a reduction, FIFO/LIFO/etc. lot-matching ran against existing lots, and the user-written `{{ total }}` was rewritten to the matched lot's per-unit. Downstream the residual calc weighed the posting by that derived per-unit instead of the user's stated total, producing a phantom E3001 imbalance. The reporter's fixture errored with `residual 50 GBP`; bean-check accepts it cleanly.

## Fix

Two parts — the bug surfaced in both the booking engine and the validator's parallel lot-match pass:

1. **Booker** — gate `is_reduction` on `method != BookingMethod::None` in `book::book`, `BookingEngine::apply`, and the validator's `update_inventories`. NONE means every posting is an augmentation; mixed-sign positions accumulate without lot matching. Mirrors Python beancount.
2. **Plumbing** — `validate_open` was falling back to `BookingMethod::default()` (STRICT) for accounts without an explicit method on their `open`, regardless of the file-level `option "booking_method"`. Plumb the resolved `effective_method` (hoisted once in `loader::process`) into a new `ValidationOptions::default_booking_method` so both the validator session and the booking engine see the same value.

## Test plan

- [x] New regression fixture `booking_none_total_cost_1182.beancount` reproduces the reporter's exact 3-transaction sequence; integration test `test_booking_none_total_cost_no_lot_matching_1182` asserts no errors of any kind.
- [x] `bean-check` accepts the fixture (Python-compat sanity check).
- [x] Existing `test_booking_method_none` (which used `@`-priced postings, not `{{ total }}` costs) still passes — no behavioral change for the previously-covered shape.
- [x] `cargo test -p rustledger-booking -p rustledger-validate -p rustledger-loader --all-features` — all pass.
- [x] `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean.

Closes #1182

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -216,7 +216,21 @@ impl BookingEngine {
                     // Check if these units reduce existing cost-bearing inventory lots.
                     // Only positions with a cost basis are considered; simple (no-cost)
                     // positions are ignored to avoid misclassifying augmentations.
-                    let is_reduction = inv.is_reduced_by(units, ReductionScope::CostBearingOnly);
+                    //
+                    // Under `option "booking_method" "NONE"` (issue #1182),
+                    // reduction matching is skipped entirely: NONE means
+                    // "accumulate positions without booking against
+                    // existing lots." Otherwise the booker would replace
+                    // the user-written `{{ total }}` cost spec with a
+                    // FIFO-matched per-unit (line ~282 below), and the
+                    // residual calculation downstream would weigh the
+                    // posting by the matched lots' costs instead of the
+                    // user's stated total — producing a phantom
+                    // E3001 imbalance for ledgers that round-trip
+                    // cleanly through Python beancount.
+                    let method = self.method_for(&posting.account);
+                    let is_reduction = method != BookingMethod::None
+                        && inv.is_reduced_by(units, ReductionScope::CostBearingOnly);
 
                     if is_reduction {
                         // Use reduce (not try_reduce) to actually update the working inventory.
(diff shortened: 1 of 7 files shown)


robcohen:
@copilot please review

PR #1198, opened 2026-05-26
feat(query): accept string amount in CONVERT() (closes #1179)

### LINKED ISSUE #1179: [Feature]: Converter (add string input to convert())
### Problem Statement

I really like that the Convert() function has a parameter for destination currency. I would like to expand this to create a full-fledged currency converter.

### Proposed Solution

What is missing is a string input, i.e. `'100 USD'` as the source.
A query like

```sql
SELECT CONVERT('100 USD', 'EUR') as conversion
```

would return a converted value based on existing prices.
Practically, a string arm in addition to position, number, etc. that parses a value.

### Alternatives Considered

I'm currently creating a fake book with prices and a fictional account just to get a conversion amount in an ad-hoc currency converter.

### Area

BQL Query Engine

### Beancount Compatibility

- [x] This is a new feature not in Python Beancount

robcohen:
## Summary

`CONVERT('100 USD', 'EUR')` now works as a rustledger extension to the BQL CONVERT function. Lets users write ad-hoc currency conversions on top of an existing price database without anchoring the conversion to a posting's value. Python beancount has no equivalent — the issue's compat checkbox confirms this is new ground.

Reporter's example shape:
\`\`\`sql
SELECT CONVERT('100 USD', 'EUR') WHERE account = 'Assets:Bank'
\`\`\`

## Implementation

- **`impl FromStr for Amount`** in rustledger-core. Strict parser: exactly `<number> <currency>` with whitespace between tokens. Rejects currency-first, thousands separators, scientific notation, lowercase commodities, leading-digit commodities, empty/single tokens. Lossless round-trip with `Display`.
- **`AmountParseError` / `AmountParseErrorReason`** carrying the offending input and failure category, so callers can produce actionable error messages.
- **New `Value::String` arm in `eval_convert`** that parses via `Amount::from_str`, then funnels through the existing `convert_amount` helper. Same semantics as the `Value::Amount` arm: same-currency shortcut, no-rate-found returns original unchanged.

The 3-arg form `CONVERT('100 USD', 'EUR', 2024-01-15)` works the same way — picks the rate at the specified date.

## Test plan

- [x] 10 unit tests in `rustledger-core::amount` pinning `Amount::from_str` strictness (round-trip, canonical forms, currency-first rejection, single-token rejection, extra-token rejection, scientific rejection, thousands-sep rejection, lowercase rejection, digit-first commodity rejection, error message includes input).
- [x] 6 BQL integration tests covering the reporter's example, same-currency shortcut, no-rate path, negative amounts, explicit-date form, and typed error messages for every malformed input shape.
- [x] `cargo test -p rustledger-core -p rustledger-query --all-features` — all pass.
- [x] `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean.

Closes #1179

### DIFF
--- a/crates/rustledger-core/src/amount.rs
+++ b/crates/rustledger-core/src/amount.rs
@@ -188,8 +188,9 @@ impl fmt::Display for Amount {
     }
 }
 
-/// Error produced by [`Amount::from_str`] when the input doesn't match
-/// the `<number> <currency>` shape that [`fmt::Display`] emits.
+/// Error produced by the [`FromStr`](std::str::FromStr) impl on
+/// [`Amount`] when the input doesn't match the `<number> <currency>`
+/// shape that [`fmt::Display`] emits.
 ///
 /// Carries the offending input so callers can surface an actionable
 /// message ("you wrote X, expected Y") rather than a generic parse
@@ -203,7 +204,8 @@ pub struct AmountParseError {
     pub reason: AmountParseErrorReason,
 }
 
-/// Distinguishes the failure modes of [`Amount::from_str`].
+/// Distinguishes the failure modes of [`Amount`]'s
+/// [`FromStr`](std::str::FromStr) impl.
 ///
 /// Separate from [`AmountParseError`] so callers can match on the
 /// category (e.g. for distinct error codes) without parsing the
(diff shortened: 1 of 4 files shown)


robcohen:
@copilot please review

Issue #1218, opened 2026-05-28
Design spike: generate TypeScript types from Rust DTOs (item 2 of #1200)


robcohen:
## Background

Tracking issue #1200 originally bundled three follow-ups under one umbrella. Items 1 (cross-binding equivalence harness) and 3 (wire-format audit) landed via PRs #1204, #1209-#1212, #1215-#1217. Item 2 — generating TypeScript types from the Rust DTOs — was deferred per the [revised priority discussion](https://github.com/rustledger/rustledger/issues/1200#issuecomment-...) because it has substantial open design questions that don't fit as a checkbox under a tracking issue.

This issue is the design spike for that work.

## The problem

`crates/rustledger-wasm/` ships **three hand-maintained TypeScript surfaces**:

- `beancount.d.ts` (494 lines)
- `beancount_wasm.d.ts` (398 lines)
- `typescript_custom_section` block inside `src/lib.rs` (wasm-bindgen embeds it in the generated `.d.ts` that ships with the npm package)

Each gets out of sync with the Rust DTOs independently. PRs #1209, #1210, #1211, #1212, #1215, #1216 all had to update all three surfaces in lockstep. PR #1210 was an audit pass that found 7 directive variants entirely missing from one surface and `Posting.price` missing from another.

This isn't going to stop on its own. Every new wire-format field needs three TS updates and an audit on whether all three actually got updated.

## What "done" looks like

The Rust DTO is the source of truth; the `.d.ts` surfaces are a derived artifact. CI fails if the checked-in TS drifts from what the generator emits for the current Rust state. Adding a new wire field touches Rust only.

### 1. Which generator?

| Option | Pros | Cons |
|---|---|---|
| [`ts-rs`](https://docs.rs/ts-rs/) | per-struct derive macro; wasm-bindgen-agnostic; `rustledger-plugin-types` can participate (it has no wasm-bindgen dep) | per-struct output needs collation into the final `.d.ts` files |
| `tsify` / `wasm-bindgen-derive` | tighter wasm-bindgen integration; one declarative annotation | couples to wasm-bindgen — `rustledger-plugin-types` can't use it |
| `specta` + bindgen | type-graph-based, supports multiple target languages (TS, Rust, Go) | bigger dep; less Rust ecosystem mindshare |

Need a small prototype against `DirectiveJson` and `PostingJson` to see what each produces and where the friction is.

### 2. How are the three TS surfaces collated?

Generators emit per-struct types. The current files are organized differently:

[...]

PR #1220, opened 2026-05-28
chore(wasm): spike ts-rs TypeScript generation + ADR-0004 (#1218)

### LINKED ISSUE #1200: Prevent wire-format drift across bindings (cross-binding tests, TS generation, audit)
## Background

#1168 was a real gap: the WASM bindings shipped without directive metadata for the entire lifetime of the crate. The Custom directive's positional `values` were also dropped. JS consumers had no way to read anything the host already parsed and tracked.

Three structural reasons this slipped through:

1. **No cross-binding equivalence test.** The same DTO pattern lives in 4 places — `rustledger-ffi-wasi::types::output`, `rustledger-wasm::types`, `rustledger-plugin-types`, `rustledger-plugin::python::compat` — and three of them carried `meta`, one didn't. Each crate's tests pin its own wire format in isolation; nothing asserts "every binding emits the same shape."
2. **`directive_to_json` is hand-written per binding.** Each crate translates `Directive → JSON DTO` independently. When `Metadata` was added to core, FFI-WASI got threaded through but WASM was missed. No compile error, because the WASM DTO didn't have a `meta` field to require.
3. **Hand-maintained TypeScript types.** `beancount.d.ts` and `beancount_wasm.d.ts` are kept in sync with the Rust types manually. Drift is invisible until someone files an issue.

This tracking issue covers three follow-ups that would prevent the next instance of this pattern.

## 1. Cross-binding metadata equivalence test

**Highest-value, ~1-day fix.** Construct a fixture `Directive` with every `MetaValue` flavor (`String`, `Bool`, `Amount`, `Account`, `Currency`, `Tag`, `Link`, `Date`, `Number`, `None`) and assert that both `rustledger-ffi-wasi`'s `directive_to_json` AND `rustledger-wasm`'s `directive_to_json` produce structurally equivalent JSON. Same fixture, same JSON, no drift.

Best home: a new `crates/rustledger-wire-format-tests` integration crate that depends on both bindings, OR a single test file mirrored into both crates with `include_str!` of a shared fixture. The former is cleaner; the latter is simpler.

Scope:
- Pin metadata equivalence (this issue's motivation).
- Pin `Custom.values` equivalence (also dropped in WASM pre-#1199).
- Pin the `kind`-tagged `CostNumber` shape already covered by `plugin-types/tests/cost_number_wire_format.rs` — add the WASM side to the same fixture.

## 2. Generate TypeScript types from the Rust DTOs

**Eliminates `.d.ts` drift by construction.** Today `beancount.d.ts` + `beancount_wasm.d.ts` are hand-written. Two contenders:

- **[`ts-rs`](https://docs.rs/ts-rs/)** — derive macro that emits `.d.ts` from `#[derive(TS)]` on the DTO structs. Build-time, no runtime cost. Most idiomatic option.
- **`tsify` / `wasm-bindgen-derive`** — similar pattern, more tightly coupled to wasm-bindgen.

Acceptance: the existing `.d.ts` files become generated artifacts (or get a `# Generated; do not edit` header) and a CI check fails if they drift from the source-of-truth Rust types. The TS file becomes a derived artifact of the Rust DTOs, not a parallel hand-maintained surface.

## 3. Audit remaining wire fields for similar drops

Quick grep-driven sweep across all four bindings (`ffi-wasi`, `wasm`, `plugin-types`, `python/compat`) to find host fields that are dropped from one or more wire shapes. Candidates to check (pre-#1199 these were the obvious ones; there may be more):

- `Posting.flag` — posting-level flag (`!`)
- `Posting.comments` / `Posting.trailing_comments`
- Source spans (`Spanned<Directive>.span`, `Spanned<Posting>.span`)
- `Document.tags` / `Document.links` (recently added in core, may not have propagated)
- `Open.booking` representation — currently `format!("{b:?}")` in WASM vs typed string in FFI-WASI
- `Transaction.trailing_comments`

For each: is the field present on the wire? Does it round-trip? If absent, is that intentional (with a comment) or accidental? File sub-issues for each accidental drop.

## Priority

1 first — without that test, items 2 and 3 catch the present moment but don't prevent the next drift. 2 second — it's the structural fix. 3 third — it's a one-shot audit that 1+2 prevent from being needed again.

## Related

- #1168 — the #1199 fix that prompted this tracking issue
- #1178 — the CostSpec refactor that established the `kind`-tagged shape and added `plugin-types/tests/cost_number_wire_format.rs` (the test pattern item 1 generalizes)

robcohen:
Re-opening of #1219 after a branch rename (`spike/...` failed the workspace's allowed-types regex). Same content, plus the deep-review fixes applied as a follow-up commit:

- `bindings/` path docstring corrected to `bindings/bindings/` in both the spike file and ADR.
- Cargo.toml feature comment now spells out why `ts-rs-spike` and the target-gated `ts-rs` dep are deliberately split.
- ADR explicitly notes the spike is a snapshot of the production DTOs and won't track future changes.

## Summary

Spike + design doc for #1218. Item 2 of the closed #1200 tracking issue, deferred for design because it had substantial open questions.

This PR lands:
- `crates/rustledger-wasm/examples/tsrs_spike.rs` — prototype that mirrors the four key wire-format DTOs (MetaValueJson, TypedValueJson, PostingJson, DirectiveJson + supporting types) with `#[derive(TS)]`. Run with `cargo test --example tsrs_spike -p rustledger-wasm --features ts-rs-spike`.
- `docs/reference/adr/0004-ts-types-from-rust-dtos.md` — proposed ADR with concrete recommendations.

`bindings/` output is gitignored; regenerable from the spike.

## Spike findings

**ts-rs got these right:**
- DirectiveJson discriminated union — narrows correctly on `switch (d.type)`.
- CostNumberJson with `tag = "kind"` — same, narrowing-friendly.
- MetaValueJson untagged union — bit-identical to hand-written.
- Doc comments → JSDoc.
- Cross-file imports compose.

**One solvable friction:**
- `Option<T> + skip_serializing_if = "Option::is_none"` emits `field: T | null` by default. The `#[ts(optional)]` attribute fixes this to `field?: T`. Verified in the spike.

**One structural limitation:**
- TypedValueJson's TS-side narrowing is lost. The Rust DTO is `struct { value_type: String, value: MetaValueJson }`; ts-rs faithfully emits the wide TS. The hand-written narrowed `type TypedValue = { type: "string"; value: string } | { type: "amount"; value: AmountValue } | ...` is more precise than the Rust struct can express. ADR proposes hand-tuning this one type via a post-process step.

## ADR-0004 recommendation

Adopt ts-rs with these constraints:
1. `#[ts(optional)]` on skip-if-none fields.
2. Replace `beancount.d.ts` + `beancount_wasm.d.ts` with a generated `bindings/index.d.ts` (post-process concat).
3. Keep `typescript_custom_section` hand-maintained in Phase 1; replace with `include_str!` of the bundle in a follow-up.
4. Hand-tune TypedValueJson narrowing in the post-process step.
5. Python `.pyi` stubs explicitly out of scope.

## Verification

- [x] `cargo check --workspace --all-targets` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo deny check` — clean (ts-rs 12.0.1 has no advisories)
- [x] `ts-rs-spike` feature gates the optional dep; default build unaffected

Relates to #1218. Supersedes #1219.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,12 @@
 /target/
 **/*.rs.bk
 
+# Generated TypeScript bindings (ts-rs spike #1218). The spike example
+# `crates/rustledger-wasm/examples/tsrs_spike.rs` exports per-type .d.ts
+# into this directory; regenerable via `cargo test --example tsrs_spike
+# -p rustledger-wasm --features ts-rs-spike`.
+crates/rustledger-wasm/bindings/
+
 # Editor files
 *.swp
 *.swo
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4032,6 +4032,7 @@ dependencies = [
  "serde-wasm-bindgen",
  "serde_json",
  "sha2 0.11.0",
+ "ts-rs",
  "wasm-bindgen",
  "wasm-bindgen-test",
 ]
@@ -4817,6 +4818,28 @@ dependencies = [
  "tracing",
  "tracing-core",
  "tracing-log",
+]
+
+[[package]]
+name = "ts-rs"
+version = "12.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "756050066659291d47a554a9f558125db17428b073c5ffce1daf5dcb0f7231d8"
+dependencies = [
+ "thiserror 2.0.18",
+ "ts-rs-macros",
+]
+
+[[package]]
+name = "ts-rs-macros"
+version = "12.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "38d90eea51bc7988ef9e674bf80a85ba6804739e535e9cab48e4bb34a8b652aa"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+ "termcolor",
 ]
 
 [[package]]
--- a/crates/rustledger-wasm/Cargo.toml
+++ b/crates/rustledger-wasm/Cargo.toml
@@ -22,7 +22,10 @@ plugins = ["dep:rustledger-plugin"]
 ## Include BQL query completions (bqlCompletions)
 completions = []
 ## Spike (#1218) -- enables the `tsrs_spike` example. Pulls in ts-rs as
-## an optional host-only dep. Not part of any default build.
+## an optional host-only dep (declared under `[target.'cfg(not(...))'.dependencies]`
+## below, so `cargo build --target wasm32-unknown-unknown --features ts-rs-spike`
+## hard-fails on the missing `ts-rs` crate -- intentional, the spike is
+## host-only by design). Not part of any default build.
 ts-rs-spike = ["dep:ts-rs"]
 
 # The `tsrs_spike` example needs the optional ts-rs dep; declare the
--- a/crates/rustledger-wasm/Cargo.toml
+++ b/crates/rustledger-wasm/Cargo.toml
@@ -62,9 +62,9 @@ js-sys = "0.3"
 # Spike (#1218): ts-rs for TypeScript type generation prototype.
 # Host-only optional dep, opt-in via the `ts-rs-spike` feature above.
 # Used by the `tsrs_spike` example to export per-struct .d.ts to
-# `target/ts-rs-spike/` so we can eyeball what ts-rs produces for
-# our existing wire-format DTO shapes. Not pulled into normal builds
-# or the wasm32 target.
+# `crates/rustledger-wasm/bindings/bindings/` (gitignored) so we can
+# eyeball what ts-rs produces for our existing wire-format DTO
+# shapes. Not pulled into normal builds or the wasm32 target.
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 ts-rs = { version = "12.0.1", optional = true }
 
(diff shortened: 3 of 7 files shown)


robcohen:
Addressed Copilot's 5 inline findings in 7723e9c3:

- ✅ #1 PostingCostJson Option fields → added `#[ts(optional)]`. Verified: `PostingCostJson.ts` now `{ number?: CostNumberJson, currency?: string, date?: string, label?: string, }`.
- ❌ #2 Transaction.payee/narration → **declined**. The production code at `src/types.rs:159-160` has NO `skip_serializing_if` on these fields, so they DO emit as `null` on the wire. The spike's `payee: string | null` faithfully matches actual wire behavior. Adding `#[ts(optional)]` would make the spike misrepresent the wire — opposite of the goal.

  However, the audit-side finding stands separately: the hand-written `beancount_wasm.d.ts` declares `payee?: string` (optional, never null), which IS a drift from the wire. Filed as #1221 — a Path A (add `skip_serializing_if` on Rust) vs Path B (fix the .d.ts) decision, separate from this PR.

- ✅ #3 Balance.tolerance → added `#[ts(optional)]`. Verified: `tolerance?: string` instead of `tolerance: string | null`.
- ✅ #4 Cargo.toml stale `target/ts-rs-spike/` path → corrected to `crates/rustledger-wasm/bindings/bindings/`.
- ✅ #5 Docstring `cargo run` → `cargo test`, with parenthetical explaining ts-rs's auto-test export mechanism.

Re-ran the spike to verify all four applied findings produce the expected output.

PR #1243, opened 2026-05-29
fix(plugin): set async_stack_size so Python WASM engine can initialize

armonge:
## Summary

`PythonRuntime::new()` currently fails on every call with:

```
<unknown>: error[E8002]: Python runtime unavailable: WASM runtime error: max_wasm_stack size cannot exceed the async_stack_size
```

surfaced to the user once per `.py` plugin listed in the ledger. The root cause is that `runtime.rs` bumps `max_wasm_stack` to 16 MiB (necessary for CPython's recursive AST visitor) but never sets `async_stack_size`. Because `wasmtime-wasi` enables wasmtime's `async` feature transitively, wasmtime enforces `max_wasm_stack <= async_stack_size` at `Engine::new` time, and the default `async_stack_size` is 2 MiB. 16 MiB > 2 MiB → engine creation fails.

This makes every Python plugin path unusable on releases shipping wasmtime ≥ 44 (i.e. v0.13.0 onward, including v0.15.0 and current `main` on wasmtime 45).

## Repro

```bash
cat > demo.beancount <<'BC'
plugin "demo.py"
2024-01-01 open Assets:Cash EUR
BC
cat > demo.py <<'PY'
def plugin(entries, options_map, config=None):
    return entries, []
PY
rledger check demo.beancount
# <unknown>: error[E8002]: Python runtime unavailable: WASM runtime error:
#   max_wasm_stack size cannot exceed the async_stack_size
```

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes

- Extract the wasmtime `Config` build into a small `engine_config()` helper so the production path and a regression test exercise the same code.
- Set `async_stack_size` to `max_wasm_stack + 1 MiB` (16 + 1 = 17 MiB) to give host frames headroom while satisfying the wasmtime constraint.
- Add `test_engine_config_satisfies_async_stack_constraint` that calls `Engine::new(&engine_config())` and asserts success. Removing the `async_stack_size` line again (or pushing `max_wasm_stack` past it in the future) will make this test fail.

## Testing

- [x] `cargo test -p rustledger-plugin --features python-plugins` passes (47 unit tests including the new regression test)
- [x] `cargo clippy -p rustledger-plugin --features python-plugins --all-targets -- -D warnings` clean
- [x] `cargo fmt` clean
- [x] Tested manually: built `cargo build --release --bin rledger` against a real ledger that uses two `.py` plugins — engine now initializes and the Python WASM module compiles successfully (~30 s, cached). Without this patch the same binary errored out before the runtime even started.
- [x] Added new test for the change

## Checklist

- [x] I have added tests that prove my fix works
- [ ] I have updated documentation if needed — no user-facing docs cover this constant

## Review Focus

- [x] Please review the algorithm/logic — specifically the choice of 1 MiB host-frame headroom. Wasmtime's docs don't pin down an exact lower bound; 1 MiB is conservative for typical Engine internals but I'm open to a different number if you have one calibrated.
- [x] General review is fine

## Related Issues

No existing issue — happy to file one separately if useful. The bug has likely been present since the Python WASI runtime landed (61d75c4, 2026-01-30), latent until wasmtime's stricter check in 44.


robcohen:
Hello. My apologies on the delay for this PR, I've been travelling on vacation and internet has been spotty. Seems like a reasonable PR to me, i'll do a review today.

robcohen (commented):
Verified the diagnosis end-to-end against `wasmtime-45.0.0/src/config.rs:2493`:

```
if self.max_wasm_stack > self.async_stack_size {
    bail!("max_wasm_stack size cannot exceed the async_stack_size");
}
```

The fix is correct, well-explained, and the regression test (`test_engine_config_satisfies_async_stack_constraint`) pins exactly the right invariant: it'll fail if either the `async_stack_size` call is dropped or `max_wasm_stack` is bumped past it. ✅

Two observations, one substantive:

## 1. (substantive) 1 MiB host-frame headroom is less than wasmtime's stock 1.5 MiB

Wasmtime's defaults give `2 MiB - 512 KiB = 1.5 MiB` of guaranteed host-function stack. The PR sets `17 MiB - 16 MiB = 1 MiB`, which is *less* generous than the default. From `wasmtime-45.0.0/src/config.rs:812-815`:

> The amount of stack space guaranteed for host functions is `async_stack_size - max_wasm_stack`, so take care not to set these two values close to one another; doing so may cause host functions to overflow the stack and abort the process.

You explicitly asked about this in "Review Focus", so: I'd bump `ASYNC_STACK_HEADROOM` to **2 MiB** (or even 4 MiB). The cost is zero in practice. `runtime.rs` uses only the sync `wasmtime_wasi::p1` API (no `.await`, no async blocks anywhere in the file), so the async stack is never actually allocated; this is purely satisfying the config validator. Bigger headroom doesn't hurt sync execution.

```rust
const ASYNC_STACK_HEADROOM: usize = 2 * 1024 * 1024;  // match wasmtime's stock 1.5 MiB+ host headroom
```

## 2. (doc nit) Attribution of where `async` comes from

The rustdoc on `engine_config` says:

> the `async` feature is compiled in (which it is, transitively, via `wasmtime-wasi`)

`wasmtime-wasi-45/Cargo.toml` actually opts *out* of wasmtime's default features (`default-features = false`, `features = ["runtime", "std"]`). The `async` feature reaches us via the workspace's own `wasmtime = "45.0.0"` dep using default features (and wasmtime's `default = [..., "async", ...]`). Cargo unions feature sets, so the result is the same, but the comment slightly mis-credits the source.

Suggested rewrite:

> When the `async` feature is compiled in (which it is: `wasmtime`'s default features include `"async"`, and the workspace dep uses defaults), wasmtime enforces ...

## Verified locally

- `cargo test -p rustledger-plugin --features python-plugins --lib test_engine_config_satisfies_async_stack_constraint` -> 1 passed
- `cargo clippy -p rustledger-plugin --features python-plugins --all-targets -- -D warnings` -> clean
- Function extraction is clean, constants are well-named, the test pins the contract at the right boundary.

Otherwise LGTM. Happy to approve once the headroom bump (or your preferred number) lands. Either way, this is a critical fix; the current main is unusable for anyone with a Python plugin.

robcohen:
Plan: I'll merge this PR as-is. The fix is correct and this is a critical unblock for everyone running Python plugins on wasmtime >= 44. I don't want the small headroom adjustment to gate the unblock.

Follow-up PR will:

1. **Bump `ASYNC_STACK_HEADROOM` from 1 MiB to 2 MiB.** Matches wasmtime's stock default difference, and the wasmtime docs explicitly warn against setting `max_wasm_stack` and `async_stack_size` too close. Zero runtime cost: the Python runtime uses only sync `wasmtime_wasi::p1`, so the async stack is never actually allocated. This is purely satisfying the config validator.
2. **Fix the rustdoc attribution** to credit `wasmtime`'s default features (which include `"async"`) rather than `wasmtime-wasi` (which actually opts out of defaults).

Thanks for the clean fix and the thoughtful "Review Focus" callout. That's exactly the kind of question a reviewer wants flagged.

PR #1304, opened 2026-06-11
fix(lsp): full-ledger completion detail (#1297) + tag/link completion (#1268)

### LINKED ISSUE #1268: [Feature]: Tag completion
### Problem Statement

https://github.com/nathangrigg/vim-beancount offers completion for tags:

Image

Rustledger LSP does not (even though it offers completions for accounts):

Image

Minimal file to reproduce:

```text
2024-01-01 open Assets:Bank:Checking USD
2024-01-01 open Expenses:Stuff USD

2024-01-15 * "Central Perk" #coffee
  Assets:Bank:Checking  -5 USD
  Expenses:Stuff

2024-01-16 * "Central Perk"
  Assets:Bank:Checking  -5 USD
  Expenses:Stuff
```

Current version: v0.13.0-312-g9cc17b7c

### Proposed Solution

Implement tag completion in the LSP 🤷 

### Alternatives Considered

_No response_

### Area

Other

### Beancount Compatibility

### LINKED ISSUE #1297: [Bug]: Completion detail only references current file, ignoring journalFile setting
## Description

The completion detail popup (showing account balance and transaction count) only references the **currently edited file**, even when `journalFile` is correctly configured to point to the main ledger file.

## Environment

- rustledger LSP (rledger-lsp)
- Neovim with blink.cmp
- Multi-file ledger setup (monthly journal files included via `include` directives in `main.beancount`)

## Configuration

```lua
init_options = {
  journalFile = "/path/to/main.beancount",
},
settings = {
  journalFile = "/path/to/main.beancount",
},
```

## Expected Behavior

Completion detail should show balance and transaction count aggregated across the **full ledger** (as resolved via `journalFile`), consistent with how `hover` (K) already works.

## Actual Behavior

Completion detail only reflects data from the file currently being edited. When editing a monthly journal file (e.g. `2024-01.beancount`), the detail shows only that month's data instead of the full ledger totals.

## Notes

- `hover` (K) correctly uses the full ledger — so the `journalFile` setting is being read for some features but not completion detail
- Workaround: none currently available; merging all files into a single file causes performance issues (BigFile detection, LSP timeout) on large ledgers (~83K lines)
- First observed: 2026-05-08

robcohen:
## Summary

Two related LSP completion fixes.

### #1297: completion detail respects `journalFile`

The completion detail popup (account balance, transaction count, price history) was computed only from the currently edited file, even when a `journalFile` was configured. While editing a monthly journal, the detail showed only that month's totals instead of whole-ledger totals, which is inconsistent with how `hover` already behaves.

`handle_completion_resolve` now operates on a directive slice, and the call site feeds it the full ledger's directives (all included files) when a `journalFile` is loaded, falling back to the current file otherwise. This mirrors how `handle_completion_request` already consults `ledger_state`.

### #1268: tag and link completion

The LSP completed accounts, currencies, and payees but not tags. Typing `#tag` (or `^link`) on a transaction header or in `pushtag`/`poptag` produced no completions.

- New `Tag` / `Link` completion contexts, detected when the token under the cursor begins with the sigil, guarded against strings, comments, and already-finished tokens.
- Known tags and links are gathered from both the current file and the full ledger (`LedgerState` now tracks them).
- `#` and `^` registered as completion trigger characters.
- The sigil is kept in the display `label` but excluded from `insert_text`/`filter_text`, so the already-typed sigil is not doubled into `##coffee`.

Note: the issue only asks for tags, but link completion shares the entire code path, so it is included. Happy to split it out if reviewers prefer.

## How to test

In a multi-file ledger wired up with `journalFile`:

1. Open an included monthly file and trigger completion on an account. The detail popup now shows whole-ledger balance and transaction count, matching `hover`.
2. On a transaction header type `#`. Known tags are offered; accepting one yields `#tag` with no doubled sigil. Same for `^` and links.

## Verification

Run in the Nix dev shell (`nix develop`):

- `cargo check -p rustledger-lsp --all-features`
- `cargo clippy -p rustledger-lsp --all-features --all-targets -- -D warnings`
- `cargo fmt -p rustledger-lsp -- --check`
- `cargo test -p rustledger-lsp --all-features`

233 lib tests plus the integration suites pass, including 9 new tests: tag/link detection on headers and `pushtag`, string/comment guards, sigil-free insert text, and a #1297 regression test proving aggregation across a multi-file directive set.

Closes #1297
Closes #1268

### DIFF
--- a/crates/rustledger-core/src/extract.rs
+++ b/crates/rustledger-core/src/extract.rs
@@ -8,7 +8,7 @@
 //! (extract, hover, completion, …) benefits.
 
 use crate::Directive;
-use crate::visit::{visit_accounts, visit_currencies};
+use crate::visit::{visit_accounts, visit_currencies, visit_links, visit_tags};
 
 /// Common default currencies included in completions.
 pub const DEFAULT_CURRENCIES: &[&str] = &["USD", "EUR", "GBP"];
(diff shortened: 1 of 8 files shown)


robcohen:
Applied two improvements from a deep self-review (architecture-focused):

- **Reuse/altitude**: moved tag/link extraction into the `rustledger-core` `visit` module (the canonical no-catch-all enumeration used for accounts/currencies/payees) as `visit_tags`/`visit_links` + `extract_tags`/`extract_links`. Both the LSP per-file path and `ledger_state` now delegate to it, removing the duplicated hand-rolled walk. This also fixes a latent gap: tags/links on `Document` directives and inside metadata are now surfaced.
- **Correctness**: replaced the naive `count('"') % 2` + `contains(';')` guard in tag detection with `in_code_position`, an escape-aware scan matching the lexer string rule, so a `;` or `#` inside a narration is no longer misclassified (e.g. `"a;b" #tag` now offers a tag).

Deferred to a follow-up PR (pre-existing design, larger refactor): CST-driven `detect_context` and structured `kind`-based completion resolve dispatch (the `match kind` path in resolve is currently dead because completion only attaches `{uri}`).

MunsuBahk:
Thank you for the thorough fix! I reported #1297 after noticing the completion detail only showed data from the current file, and it's great to see it resolved so cleanly.

I've filed a few bug reports in other projects before, but this is the first time I've actually received a response and seen a fix land — it means a lot. Really appreciate the work.

Issue #1342, opened 2026-06-14
[Bug]: BQL parser stack-overflows / DoSes on deeply-nested parentheses


robcohen:
Found while investigating an intermittent `Fuzz Query (fuzz_query_parse)` CI failure (slow-unit artifacts). These are a **real denial-of-service / stack-overflow bug**, not benign slow inputs.

## Repro

A ~2 KB BQL string of deeply-nested `(` (e.g. 1023 of them) sent to `rustledger_query::parse`:

- **SIGSEGV** (stack overflow) on a default 8 MB stack, and
- on a larger (256 MB) stack, parses in **super-linear time**: measured **2.7 s / 5.3 s / 10.5 s** for three ~2 KB fuzzer inputs.

The recursive (chumsky) grammar in `crates/rustledger-query/src/parser.rs` descends one stack frame per open paren (parenthesised expressions, function args, subqueries) with **no depth bound**.

## Impact

`parse()` is reachable from `rledger query` and any embedder. An untrusted/large query can crash the process (stack overflow) or hang it (multi-second parse for ~2 KB). The query fuzzer surfaced it as `slow-unit-*` artifacts, which the per-target crash gate (`ls -A artifacts && exit 1`) then reports intermittently.

## Fix

Bound parenthesis nesting up front. PR #__ adds a `MAX_NESTING_DEPTH = 128` pre-flight check (O(n), skips parens inside string literals) so over-deep input fails fast with a clean `SyntaxError` before the recursive parse.

After the fix the three fuzzer slow-units fail fast in **3-27 µs** on the default stack (no SIGSEGV).

## Notes

- Unrelated to the in-flight PRs (#1338/#1341); this is a pre-existing query-parser bug.
- Follow-up worth considering: the `fuzz.yml` "Check for crashes" step treats any artifact (including `slow-unit-*`) as a crash. Once this DoS is fixed the slow-units disappear, but the gate could still be refined to distinguish genuine `crash-*`/`oom-*`/`timeout-*` from `slow-unit-*`.

PR #1344, opened 2026-06-14
ci(fuzz): don't fail the crash gate on benign slow-units

robcohen:
Follow-up to #1342.

## Problem

Each of the three fuzz jobs (`parser` matrix, `query`, `booking`) gates on:

```bash
if [ -d "fuzz/artifacts/<target>" ] && [ "$(ls -A fuzz/artifacts/<target>)" ]; then
  echo "::error::Fuzzer found crashes!"; exit 1
fi
```

This fails on **any** artifact. But libFuzzer writes `slow-unit-*` files for inputs that exceed its slow-execution threshold yet **still complete** — these are not crashes, and because slowness is timing-dependent on the runner, the gate went **intermittently red** (e.g. `fuzz_query_parse` on #1338, which touches no query code).

## Fix

Distinguish genuine failure artifacts (`crash-*` / `oom-*` / `timeout-*` / `leak-*`) from `slow-unit-*`:

- **Real crash artifact** → `::error::` + `exit 1` (unchanged, still uploaded via the `failure()` step).
- **Only slow-units** → `::warning::` listing them, build **passes**.
- **Nothing** → passes.

Applied to all three "Check for crashes" steps.

## Verification

YAML validated; the gate shell logic unit-tested across cases:

| artifacts | result |
|---|---|
| empty | PASS |
| `slow-unit-*` only | **PASS + warning** (was FAIL) |
| `crash-*` | FAIL |
| `slow-unit-*` + `crash-*`/`oom-*` | warn + FAIL |

## Scope

`fuzz.yml` only. Independent of #1343 (which removes the *specific* slow-units by fixing the parser DoS) — this hardens the gate so the whole **class** of slow-unit false-positives can’t recur, while still surfacing slow inputs as warnings for investigation.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -109,13 +109,26 @@ jobs:
         if: steps.check.outputs.should_run == 'true'
         run: |
           cd crates/rustledger-parser
-          if [ -d "fuzz/artifacts/${{ matrix.target }}" ] && [ "$(ls -A fuzz/artifacts/${{ matrix.target }})" ]; then
-            echo "::error::Fuzzer found crashes!"
-            ls -la fuzz/artifacts/${{ matrix.target }}/
-            exit 1
-          else
-            echo "No crashes found"
+          dir="fuzz/artifacts/${{ matrix.target }}"
+          if [ ! -d "$dir" ] || [ -z "$(ls -A "$dir" 2>/dev/null)" ]; then
+            echo "No crash artifacts found"
+            exit 0
           fi
+          # Real failures are crash-/oom-/timeout-/leak- artifacts.
+          # `slow-unit-*` means an input was slow but completed — surface
+          # it as a non-fatal warning instead of failing the build (#1342).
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed the review (Copilot):

- **`ls` in command substitution** (3×): replaced the `[ -z "$(ls -A …)" ]` emptiness guard with find-only detection (`find … 2>/dev/null` is empty on a missing dir), so there is no command-substitution-under-`set -e` concern and no `ls` output parsing. Cleaner across all three steps.
- **"filter positively on crash-/oom-/timeout-/leak-"** (3×): **respectfully declined** — failing on *any* non-`slow-unit-*` artifact is intentional. It is the safe **superset** of libFuzzer's real-failure prefixes (the list also includes `mismatch-`, and future/unknown types), so a positive allowlist would silently let new failure-artifact types pass. Updated the in-code comment to make that intent explicit (it previously read like the four prefixes were exhaustive, which is what prompted the flag). Verified `mismatch-*` still fails the gate; `slow-unit-*`-only passes with a warning.

Issue #1238, opened 2026-05-28
cargo-mutants: periodic mutation-testing audit on rustledger-core + rustledger-parser


robcohen:
## Background

`cargo-mutants` performs mutation testing: it makes small modifications to source code (replace `+` with `-`, replace `<` with `<=`, etc.) and checks whether the test suite catches the change. Mutations that survive (tests still pass) are evidence that the tests aren't actually exercising the mutated logic.

Repo today has good unit test coverage and growing property test coverage (issue #1235), but coverage measures **whether a line was hit**, not **whether the test would fail if the line were wrong**. Mutation testing closes that gap.

## What "done" looks like

A `cargo mutants` job that runs against a curated set of crates (start small — `rustledger-core`, `rustledger-parser`) and produces a report of surviving mutants. The report is reviewed periodically; surviving mutants either get a test or get an annotation explaining why they're un-killable (rare; usually means the code is dead and should be removed).

This is **not** a PR gate. Mutation testing runs are slow (10-60 min for a single crate) and produce noise. It's a quarterly audit tool, like `cargo udeps` or `cargo-bloat`.

### 1. Where in CI?

- A scheduled weekly job that posts results as an issue or dashboard
- An on-demand `gh workflow run mutants` for an author to invoke before a refactor
- A `just mutants` recipe for local use

Probably (1) + (3). Skip (2) until there's demand.

### 2. Mutant filtering

Some mutants are trivially un-killable (e.g. mutating a debug-print format string). Configuration to filter these out keeps the report signal-to-noise high. `cargo-mutants` supports config files for this.

### 3. Mutation operators to include

Default operators are arithmetic and comparison swaps. Worth considering whether to also enable boolean-literal swaps, return-value mutations, etc. — these tend to find more bugs but produce more false positives.

## Why "honorable mention"

Mutation testing has high *intellectual* value but middling *practical* leverage on a codebase that already has strong unit + integration + property tests. It's worth doing once to find the gaps, then revisiting every few months — not worth a hard CI gate.

## Related

- Issue #1235 — property tests at pipeline boundaries. Mutation testing is a natural follow-up to property tests: it verifies that the property tests are actually catching the mutations they should.
- ADR-0001 (testing standards)

robcohen:
Closing — the audit tooling this issue asked for is all in place, and the first real audit has been run and acted on.

**"Done" criteria:**

| Wanted | Status |
|---|---|
[...]

PR #1345, opened 2026-06-14
feat(just): local mutation-testing recipes (#1238)

robcohen:
Partially addresses #1238 (the "(3) local recipe" piece).

## What changed after review

A deep review caught that **`.github/workflows/mutation.yml` already exists** on `main` ("Mutation Testing" — monthly matrix over 6 crates, `--all-features`, mutation-score + per-crate threshold enforcement). My original `mutants.yml` duplicated it with weaker conventions, so I **removed it**.

What was genuinely missing — and is all this PR now adds — are the **local `just` recipes** (no mutation recipe existed in the justfile):

- `just mutants` — audit the curated crates (`rustledger-core` + `rustledger-parser`).
- `just mutants-crate <pkg>` — audit a single crate.
- Both pass extra cargo-mutants args through, e.g. `just mutants --file src/inventory/booking.rs`, and append `-- --all-features` so **local results match what `mutation.yml` runs in CI** (e.g. the `fuzz`-gated constructors in `rustledger-core`).

Net diff: **`justfile` only.**

## Verification

`cargo mutants --package rustledger-core --list` (via the recipe, with `--all-features`) enumerates the mutant set with `.cargo/mutants.toml` applied — confirms the recipes, package selection, and passthrough resolve.

## Issue status

#1238s scheduled-audit half (option 1) is already covered by `mutation.yml`; this closes the local-recipe half (option 3). Leaving the issue open is reasonable if the maintainers still want the "post results as a tracking issue/dashboard" follow-up — happy to adjust the `Closes`/`Refs` as preferred.


robcohen:
All three Copilot comments were on `.github/workflows/mutants.yml`, which has been **removed** in the review revision (it duplicated the existing, more capable `.github/workflows/mutation.yml`):

- *duplicate-workflow* (line 15) → resolved by deleting the workflow.
- *`--all-features`* (line 49) → folded into the surviving `just` recipes (`-- --all-features`) so local audits match `mutation.yml`.
- *truncate `missed.txt`* (line 76) → moot (no workflow summary here).

This PR is now `justfile`-only (the two local recipes), retitled accordingly.

Issue #1290, opened 2026-06-09
[Compat]: formatter whitespace disagrees with bean-format


lutzky:
### Description

Context: I use beancount for my CI and main workflow, and want to introduce rledger as an editor-built-in replacement. Therefore, `rledger format` must agree with `bean-format`. See also #1252 - this issue remains after #1284, but we'll use the same example.

`rledger format` removes some empty lines, which doesn't cause any problems in my particular case. However, the problem is that it disagrees with `bean-format` over how many spaces some lines should have.

If I take the beancount file below as `bf.bean`, and run:

```bash
$ rledger format bf.bean > rl.bean
$ bean-format rl.bean > bf2.bean
$ rledger format --check --diff bf2.bean
```

```diff
--- bf2.bean
+++ bf2.bean (formatted)
@@ line 8 @@
-  Assets:Money  -1024 USD
+  Assets:Money           -1024 USD
@@ line 12 @@
-  Assets:Money    0.1 USD
+  Assets:Money             0.1 USD
@@ line 19 @@
-  Assets:Money   -0.1 USD
+  Assets:Money            -0.1 USD
```

That is, `bean-format` and `rledger format` are arguing with each other.

Ostensibly, this is related to the `-W` and `-w` flags for `rledger format` as mentioned in [docs/commands/format.md](https://github.com/rustledger/rustledger/blob/main/docs/commands/format.md). However, as of 3160e89d9b9dbc5277e2a196e5d0bd8e888a327f, these have been removed! Found like so:

```bash
git bisect run sh -c "cargo run --bin rledger -- format --help | grep -q num-width"
```

Even if it's possible to fix this with `-W` and `-w`, there three issues:

1. As mentioned, it has been removed
2. It does not appear to be possible to change these settings for the LSP version
3. The defaults between `rledger format` and `bean-format` should match, at least to the extent that they won't be "arguing with each other" as explained above.

### Beancount File

```text
; This comment gets to stay

2024-07-19 open Assets:Money ; This comment is deleted

; Notably, the following comment is load-bearing for CI:

2024-07-19 open Expenses:Thingamabobs ; cSpell: word Thingamabobs

2024-07-20 * "Commas should stay"
  Assets:Money  -1,024 USD
  Expenses:Thingamabobs

2024-07-20 * "bean-format does not remove plus signs, neither should rledger format"
  Assets:Money    +0.1 USD
  Expenses:Thingamabobs

; Behold, the incredible disappearing pushtag!

[...]

PR #1347, opened 2026-06-14
feat(lsp): align inferred-amount inlay hints with the format column (#1346)

### LINKED ISSUE #1346: [Feature]: LSP amount hints should align with formatting
### Problem Statement

Here's a screenshot from `format_compat/cases/realistic_multi_feature/expected.bean`:

Image

Note that the amount hints (not sure what they're called - the greyed-out `-1000.00 USD`, `-4.50 USD` and `-52.30 USD` do not line up with the actually-entered amounts. If I explicitly enter all amounts, the alignment is much nicer:

Image

### Side-note about whole-file alignment

I'm not sure why the alignment goes so far to the right... from #1324:

> Why normalize. rledger and bean-format deliberately disagree on exactly one axis: the column model. bean-format aligns every amount to one whole-file column; rledger aligns per transaction, so a long account name elsewhere in the file doesn't reflow every other transaction. The script normalizes away that alignment padding (leading indentation preserved, so a real indent regression still fails) before comparing.

...though this seems to have `rledger` aligning across multiple transacations 🤷  (though not quite whole-file - `balance` is still aligned differently).

I'd personally prefer it if rledger at least had an *option* to agree with `bean-format` about this, as it would make my life ever-so-slightly easier with CI 😅 

## Option 1 - only align hints

This only works for transactions 2,3 in the file - render the hint at line 20 so that it aligns with line 19 (2 columns right), and the hint at line 26 so that it aligns with line 24 (1 column right). Lines 15-16 will remain unfortunately out-of-alignment.

## Option 2 - align everything

A-priori have `rledger format` align things pretending that the implicit numbers are there. That is, have `rledger format` output this:

```
2020-01-02 * "Opening" #setup
  Assets:Checking           1000.00 USD
  Equity:Opening-Balances
```

...because it's pretending to see this:

```
2020-01-02 * "Opening" #setup
  Assets:Checking           1000.00 USD
  Equity:Opening-Balances  -1000.00 USD
```

...and then also do Option 1.

### Alternatives Considered

_No response_

### Area

### Beancount Compatibility

- [x] This is a new feature not in Python Beancount

robcohen:
Closes #1346 (Option 1).

## Problem (reported by @lutzky)

The LSP renders a greyed-out **inferred-amount inlay hint** for elided posting amounts (the auto-computed balancing leg). It was placed at a fixed 2-space gap after the account, so it never lined up with the sibling explicit amounts:

```
2020-01-03 * "Groceries"
  Expenses:Food:Groceries    4.50 USD
  Assets:Checking  -4.50 USD        ← hint floats here, not under 4.50
```

## Fix

Right-justify the hint's number into the formatter's number field — `ParseResult::alignment` (`number_col` / `number_width`), the same file-wide alignment `rledger format` and the range-formatter already use. The hint now sits where the real amount would, so its number/currency line up with explicit amounts:

```
2020-01-03 * "Groceries"
  Expenses:Food:Groceries    4.50 USD
  Assets:Checking            -4.50 USD   ← aligned to the format column
```

This is **Option 1** from the issue:
- A number-bearing sibling drives the column → the hint aligns to it.
- When the *elided* posting has the longest account, the formatter excludes it from the column (it can't drive alignment it doesn't participate in — see #1290), so the hint clamps to the 2-space minimum rather than overlapping the account. This is the documented Option-1 limitation (the issue's "lines 15-16 remain out of alignment").

## Tests

Two new tests pin the behavior by asserting the hint's number lands at `number_col + number_width - num_len` (right-justified slot): the aligning case and the long-elided-account fallback. Full `rustledger-lsp` suite (236) green; clippy clean.

## Notes / follow-ups

- **Option 2** (the issue's "align everything" — feed the *inferred* amounts into the column computation too) is deliberately out of scope. It would mainly help the one corner where an inferred number is wider than every explicit one (e.g. a leading `-` on the balancing leg pushes its currency 1 column past the explicit currency). Worth a follow-up if desired.
- Inlay rendering is client-side; this is verified via the computed position/label (the same way the existing inlay tests are). A quick visual confirm in an editor is recommended before release.

### DIFF
--- a/crates/rustledger-lsp/src/handlers/inlay_hints.rs
+++ b/crates/rustledger-lsp/src/handlers/inlay_hints.rs
@@ -166,6 +166,33 @@ pub fn handle_inlay_hints(
             let indent = line.len() - line.trim_start().len();
             let end_col = indent + trimmed.len();
 
+            // Align the greyed-out amount with the column `rledger
+            // format` uses for explicit amounts (issue #1346), instead
+            // of a fixed 2-space gap that left the hint out of line with
+            // sibling amounts. The formatter right-justifies the number
+            // in a `number_width` field starting at `number_col`
+            // (`ParseResult::alignment`, the same file-wide alignment
+            // the parser pre-computed); reproduce that so the hint sits
+            // where the real amount would.
+            let num_text = amount.number.to_string();
+            let align = parse_result.alignment;
+            let label = if align.number_col > 0 {
+                // Start column of the right-justified number text.
+                let num_start =
+                    align.number_col + align.number_width.saturating_sub(num_text.chars().count());
+                // Gap from the account end to the number, never below
+                // the conventional 2-space minimum. A long amount-less
+                // account whose end passes `num_start` can't be aligned
+                // — the formatter excludes such postings from the column
+                // for the same reason (#1346 Option 1 limitation).
+                let gap = num_start.saturating_sub(end_col).max(2);
+                format!("{}{} {}", " ".repeat(gap), num_text, amount.currency)
+            } else {
+                // No number-bearing postings to align to: fall back to
+                // the conventional 2-space gap.
+                format!("  {} {}", num_text, amount.currency)
+            };
+
             // Store data for resolve - include account for rich tooltip
             let data = serde_json::json!({
                 "kind": "inferred_amount",
@@ -176,11 +203,14 @@ pub fn handle_inlay_hints(
 
             hints.push(InlayHint {
                 position: Position::new(posting_line, end_col as u32),
-                label: InlayHintLabel::String(format!("  {} {}", amount.number, amount.currency)),
+                label: InlayHintLabel::String(label),
                 kind: Some(InlayHintKind::TYPE),
                 text_edits: None,
                 tooltip: None, // Resolved lazily
-                padding_left: Some(true),
+                // The leading gap is baked into the label so the number
+                // lands exactly at the formatter's column; no extra
+                // client-side left padding (which would shift it by 1).
+                padding_left: Some(false),
                 padding_right: None,
                 data: Some(data),
             });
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed the review findings (Copilot + a deep self-review). All three were real:

**1. Combined-gap clamp dropped right-justification (High).** `gap.max(2)` clamped the *whole* gap, so when a long elided account forced the field pad to its 2-space minimum while the justify pad was nonzero, the number landed one column left of the explicit amounts. Now mirrors `emit_posting` exactly: `field_pad.max(2) + justify_pad`. New test `..._field_and_justify_pad_combine_at_clamp_boundary_1346` pins it (the buggy code gave 2 spaces; correct is 3).

**2. `position.character` was a UTF-8 byte length (High).** Wrong under the default UTF-16 negotiation — misplaced the hint on any non-ASCII line. Now the account-end byte offset is converted through `LineIndex` (`position_to_offset` + `offset_to_position`) so the position is in the negotiated encoding. (Pre-existing on `main`; fixed here since the code was being touched.)

**3. Gap math mixed bytes and chars (High).** `end_col` was byte-based while `number_col`/`number_width` are char-based. Now computed in **both** units, kept separate: bytes for the position, chars for the visual gap.

**4. Test brittleness (Low).** The expected number length is derived from the hint label instead of a hard-coded `5`.

New regression coverage: the clamp-boundary case, and a multi-byte account (`Assets:Café`) asserting both the UTF-16 `position.character` (13, not the byte length 14) and char-grid currency alignment. Full `rustledger-lsp` suite (239) green; clippy clean.

robcohen:
Ran a fresh deep review of the post-fix code (after `d4744e5f`). **Verdict: merge with nits — no new bugs.**

The three review fixes were verified correct, empirically, across CRLF, BOM, non-ASCII, tab indents, no-trailing-newline, width-saturation, and the clamp boundary:
- `gap = field_pad + justify_pad` mirrors `emit_posting` exactly (justify pad correctly *not* clamped).
- The UTF-16 position is right (`hint_line == posting_line` always; BOM needs no adjustment here because `directives` spans are already in the original-source frame, unlike range-formatting).
- The byte/char split is sound; `line[..indent_bytes]` is always a valid boundary (posting indent is ASCII).

Both new tests genuinely pin their fixes (they fail if the fix is reverted).

**One documented caveat (pre-existing, by-design, not introduced here):** the hint aligns to the *canonical* formatter column, so on a non-canonically-formatted line (e.g. 3-space indent, tabs, extra inner spaces) it sits where `rledger format` *would* place the amount — which only matches the on-screen explicit amount once the file is formatted. Added an in-code comment spelling this out (`d4744e5f`+1). The same `number_col`-based math predates this PR, so it’s not a regression.

No blockers remain from my side.

Issue #1309, opened 2026-06-11
[testing] I ran mutation testing with fable this morning


matthiasdebernardini:
hey I'm keeping my own fork of this so I can have my agent friencly version #1291 , 

Do you want the skill I used to run this? This is what it found, I'm using cargo mutants.

I'm also going to run property testing with fable.

<img width="1718" height="331" alt="Image" src="https://github.com/user-attachments/assets/b3b461f1-78d3-4d46-a223-5009f78d2ae9" />

robcohen:
Sure, I'd be happy to look at it

robcohen:
This is a great audit — thank you. Yes please, I'd love the killing tests, and yes to the `.cargo/mutants.toml` cleanup. Let me make sure we don't duplicate work.

I've opened #1329 with the first slice:

- **`is_balanced` is fixed.** You were exactly right — `replace is_balanced -> bool with true` passed the entire suite because both existing tests only assert the balanced case. Added a negative test and an exact-tolerance-boundary test; `cargo mutants --re is_balanced` now reports 5/5 caught.
- **Honest scoring.** Excluded `kani_proofs.rs` via `exclude_globs` (it's `#![cfg(kani)]`, so `cargo test` never compiles it and every mutant there is missed by construction — that's your ~195 noise). `cargo mutants --list` now shows 0 kani mutants. This should pull the adjusted ratio back up to roughly what you measured.
- Added `rustledger-parser` to the monthly scheduled audit (it was only reachable via the manual `all` before).

So you can **skip cluster 2's `is_balanced` and the mutants.toml work** — both are in #1329. That leaves the three big domain clusters, which are squarely your area and where the real correctness value is:

1. **Lot-reduction cost-basis** (inventory, ~71) — the HIFO/avg/FIFO-LIFO sign-flip and guard-inversion survivors. This is the scariest cluster and the one I'd prioritize; partial-lot reductions never being asserted with exact values is a real hole.
2. **Residual / interpolation math** (booking, ~61) — the rest of cluster 2 beyond `is_balanced`.
3. **`calculate_tolerances`** (validate, ~50) — the "replace the whole map with one default entry" survivor.

If you take 1 and 3, I'm happy to take the remaining booking residual/interpolation mutants in 2 so we converge faster — or slice it however you prefer. Either way, one focused PR per cluster keeps review tractable.

[...]

PR #1348, opened 2026-06-14
test(core): kill lot-reduction mutants in inventory booking (#1309)

robcohen:
Cluster 1 of 3 from the #1309 cargo-mutants audit (@matthiasdebernardini). The "scariest" cluster per the triage: lot-reduction cost-basis in `rustledger-core/src/inventory/booking.rs`, which had **no direct unit tests** for the reduction methods — so 56 mutants survived (sign flips, cost-basis arithmetic, guard inversions, filter `&&`/`==`, partial-lot takes).

## What

A new `reduction_tests` module (28 exact-value tests) covering both the read-only `try_reduce_*` preview path and the mutating `reduce_*` commit path:

- **FIFO / LIFO / HIFO / Average / Strict / StrictWithSize / merge / None**
- exact **cost basis** for partial multi-lot reductions
- **lot selection order** (oldest / newest / highest-cost first)
- **filter isolation** — currency, empty lots, and sign, so a mutated predicate can't pull in the wrong lot
- **insufficient-units accounting** (`requested - remaining`)
- **merge** currency/sign filtering + non-merged lots preserved
- **post-reduction position state** asserted directly (the `units()` cache had masked the `new_units` mutant)

## Result

**56 to 10 surviving mutants** (154 caught — 94% of viable). The 10 remaining are **provably equivalent and un-killable**:

| Mutant | Why equivalent |
|---|---|
| `split(x * signum())` `*` to `/` (x6) | `signum()` is ±1, so `x * (±1)` == `x / (±1)` |
| `!has_any_cost \|\| cost_currency.is_none()` flipped to `&&` | operands are logically equivalent (`has_any_cost` iff `cost_currency.is_some()`) |
| three `&&` in `try_reduce_strict` filter | on multiple matches the STRICT path delegates to `try_reduce_ordered`, which re-applies the same filter — the mutation can't change observable output |

cargo-mutants can't skip these per-line (its `exclude_re` matches the function-level mutant text), so they're documented here rather than annotated.

## Verification

`cargo test -p rustledger-core` green; clippy clean; mutation re-run confirms 10/164 (all equivalent). Verified locally via `nix develop`.

Part of #1309 — clusters 2 (booking residual/interpolation) and 3 (validate `calculate_tolerances`) follow in separate PRs.

### DIFF
--- a/crates/rustledger-core/src/inventory/booking.rs
+++ b/crates/rustledger-core/src/inventory/booking.rs
@@ -972,3 +972,470 @@ impl Inventory {
         })
     }
 }
+
+#[cfg(test)]
+mod reduction_tests {
+    //! Direct unit tests for the read-only `try_reduce_*` booking paths.
+    //!
+    //! These pin exact cost-basis, lot selection, and guard behavior so
+    //! the lot-reduction mutants surfaced by the #1309 audit are killed
+    //! (the public mutating `reduce_*` path was covered indirectly, but
+    //! the `try_reduce_*` preview path had no direct assertions).
+    use crate::{Amount, BookingMethod, Cost, CostSpec, Inventory, Position, naive_date};
+    use rust_decimal::Decimal;
+    use rust_decimal_macros::dec;
+
+    fn d(n: i64) -> Decimal {
+        Decimal::from(n)
+    }
+
+    /// A cost-bearing lot of `units` STK at `cost` USD, dated 2024-01-`day`.
+    fn lot(units: i64, cost: i64, day: u32) -> Position {
+        Position::with_cost(
+            Amount::new(d(units), "STK"),
+            Cost::new(d(cost), "USD").with_date(naive_date(2024, 1, day).unwrap()),
+        )
+    }
+
+    fn mk(lots: impl IntoIterator<Item = Position>) -> Inventory {
+        let mut i = Inventory::new();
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed a deep-review finding: my "10 remaining are provably equivalent" claim was **overstated**. The three `&&` connectors in `try_reduce_strict`'s filter are in fact killable — against an inventory holding only a same-sign (short) lot, correct code returns `NoMatchingLot` (0 matches) while each `&& -> ||` mutant wrongly admits the lot (`currency == STK`, or the always-true `matches_cost_spec` on the default spec, satisfies the disjunction) and succeeds via `try_reduce_from_lot`.

Added `strict_rejects_when_only_same_sign_lot_present`, which kills all three. Also corrected the "filter isolation" fixtures: a zero-units lot is dropped by `Inventory::add`, so the `!is_empty()` clause is unreachable for add-built inventories — removed the dead empty lot and the misleading `_and_empty` test names.

**Updated result: 56 → 7** (157 caught). The 7 are now genuinely all equivalent: `split(x * signum())` `*`→`/` (×6, `signum()` is ±1 after the non-empty filter) and the redundant `!has_any_cost || cost_currency.is_none()` `||`→`&&` (the operands are logically identical). Pushed in `702ae466`.

PR #1349, opened 2026-06-14
test(booking): kill residual/interpolation mutants (#1309 cluster 2)

robcohen:
Cluster 2 of 3 from the #1309 cargo-mutants audit (@matthiasdebernardini): residual / interpolation math in `rustledger-booking/src/interpolate.rs`. Despite 29 existing tests, 28 mutants survived in unasserted arithmetic and edge paths.

## What

10 new exact-value tests:
- **unit-price** and **total-price** interpolation (units × price; total)
- **multi-posting residual sum** (-(Σ explicit))
- **single elided leg splitting two currency residuals** into two postings
- **sub-cent residual preservation** — `round_interpolated` must not round a non-zero residual to `0.00` and silently leave the txn unbalanced
- **post-fill residual returns to zero**
- **CurrencyOnly elided fill** zeroes the residual

## Result

**28 → 17 surviving mutants** (57 caught). The core, financially-meaningful arithmetic — cost basis, price conversion, residual totals, precision preservation — is now covered.

The 17 remaining:
- **4 provably equivalent**: `x * signum()` `*` → `/` (signum is ±1, so ×≡÷).
- **13 in niche edge paths**, lower risk and killable only with more targeted inputs:
  - total-price **scale tracking** (`scale > 0`) — affects display precision only, not the balance;
  - the **NumberOnly** / **incomplete-price** residual arms (reachable via unusual/malformed inputs);
  - the **PerUnitFromTotal** booked-cost form (post-booking);
  - the single-vs-multi-currency **split guard** (`> 1`);
  - **prune-shift `filled_indices`** bookkeeping (index ordering after a zero-fill is pruned — not amounts).

I documented these rather than over-chase; the edge-path ones are a reasonable follow-up if a higher mutation score is wanted here.

## Verification

`cargo test -p rustledger-booking` green; clippy clean; mutation re-run confirms 17/74. Verified locally via `nix develop`.

Part of #1309 — cluster 1 is #1348; cluster 3 (validate `calculate_tolerances`) follows.

### DIFF
--- a/crates/rustledger-booking/src/interpolate.rs
+++ b/crates/rustledger-booking/src/interpolate.rs
@@ -1840,14 +1840,21 @@ mod tests {
     #[test]
     fn interpolate_number_only_infers_currency_and_balances() {
         // A NumberOnly leg (`-100`, currency missing) infers its currency
-        // from the sibling USD posting and contributes `number` to the
-        // residual (kills that arm's `residual += *number`).
+        // from its OWN price annotation (the arm only consults the
+        // posting's own cost/price, never siblings — a bare NumberOnly
+        // with no cost/price would route to the unassigned path instead).
+        // The `@ 1 USD` price is a currency hint with a unit multiplier,
+        // so the leg contributes `-100` to the residual via that arm's
+        // `residual += *number` — which this test kills.
         let txn = Transaction::new(date(2024, 1, 1), "number-only")
             .with_synthesized_posting(Posting::new("Expenses:X", Amount::new(dec!(100), "USD")))
-            .with_synthesized_posting(Posting::with_incomplete(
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed a deep-review finding: `interpolate_number_only_infers_currency_and_balances` did not kill what its docstring claimed. A bare `NumberOnly` leg has no currency to infer (the arm consults only the posting's own cost/price, never siblings), so it routed to the unassigned-fill path — not the NumberOnly residual arm (line 333). Gave the leg an `@ 1 USD` price (currency hint, unit multiplier) so it now reaches line 333; confirmed the `+=` mutants there die (file 17 → 15). Pushed `a8afd210`.

PR #1353, opened 2026-06-14
test(validate): kill calculate_tolerances mutants (#1309 cluster 3)

robcohen:
Cluster 3 of 3 from the #1309 cargo-mutants audit (@matthiasdebernardini): `calculate_tolerances` (and its `decimal_quantum` helper) in `rustledger-validate/src/validators/transaction.rs`. The file had **no unit tests**, so the "replace the whole map with one default entry" survivor plus all the tolerance arithmetic survived (22 mutants in these two functions).

## What

`calculate_tolerances` is a pure `(txn, options) -> HashMap<Currency, Decimal>` function, so a new `tolerance_tests` module asserts the returned map exactly:

- **`decimal_quantum`** by scale (`0.01` / `0.001` / `1`)
- **base tolerance** = `quantum × multiplier`, per-currency max; integer (scale-0) amounts contribute nothing (no entry)
- **cost-inferred** = `units_quantum × multiplier × cost_per_unit`
- **price-inferred** = `units_quantum × multiplier × price`
- **per-currency `inferred_tolerance_default`** acts as a floor
- **wildcard `*`** default applies to all currencies

## Result

The `calculate_tolerances` + `decimal_quantum` mutants drop from **22 → 0**. File overall **26 → 4** (59 caught).

The 4 remaining are in **unrelated validators** in the same file — `validate_transaction_structure`, `validate_account_lifecycle`, `validate_transaction_balance` (simple comparison operators) — outside this cluster's scope. A reasonable follow-up, noted for visibility rather than chased here.

## Verification

`cargo test -p rustledger-validate` green; clippy clean; mutation re-run confirms 0 survivors in the targeted functions. Verified locally via `nix develop`.

Completes the #1309 cluster sweep: cluster 1 (#1348, inventory lot-reduction), cluster 2 (#1349, residual/interpolation), cluster 3 (this).

(Supersedes #1350 — auto-closed by the branch rename off `test/`.)

### DIFF
--- a/crates/rustledger-validate/src/validators/transaction.rs
+++ b/crates/rustledger-validate/src/validators/transaction.rs
@@ -649,12 +649,18 @@ mod tolerance_tests {
                 .with_currency("USD"),
         );
         let t = calculate_tolerances(&mk_txn(vec![p]), &opts);
+        // USD from the cost; STK from the units-quantum base (0.01 * 0.5).
+        // Assert the whole map so an unexpected/missing entry is caught.
         assert_eq!(t.get(&cur("USD")), Some(&dec!(0.01)));
+        assert_eq!(t.get(&cur("STK")), Some(&dec!(0.005)));
+        assert_eq!(t.len(), 2);
     }
 
     #[test]
     fn tolerance_price_inferred_is_units_quantum_times_mult_times_price() {
-        // infer_from_cost: 10.00 STK @ 3.00 USD -> 0.005 * 3.00 = 0.015.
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed the review: the cost and price tests now assert the full map — the STK units-quantum base entry (0.01×0.5=0.005) plus `t.len()` — matching the "assert exactly" intent; floor/wildcard tests assert `t.len() == 1`; and the price test's comment is corrected (it exercises the price path, gated by `infer_tolerance_from_cost`). Pushed.

PR #1399, opened 2026-06-15
feat(ffi): emit per-position cost in Inventory query output

### LINKED ISSUE #1398: FFI-WASI: emit per-position cost in Inventory query output
## Summary

The FFI-WASI query output serializes an `Inventory` as units-only — each position carries `{"units": {"number", "currency"}}` with **no cost basis**. Consumers therefore cannot reconstruct per-lot cost detail from an inventory column. Request: optionally emit the per-position `cost` in the Inventory payload so embedders can preserve lots.

## Current behavior (verified against v0.16.0 wasm)

For two lots of the same commodity bought at different costs:

```
2024-01-01 * "buy1"
  Assets:Broker   2 ITOT {10 USD, 2024-01-01}
  Assets:Cash    -20 USD
2024-02-01 * "buy2"
  Assets:Broker   3 ITOT {11 USD, 2024-02-01}
  Assets:Cash    -33 USD
```

`SELECT account, sum(position) ... GROUP BY account` returns:

```json
{"positions": [
  {"units": {"currency": "ITOT", "number": "2"}},
  {"units": {"currency": "ITOT", "number": "3"}}
]}
```

The two lots are present as separate entries, but the cost bases (`10 USD` / `11 USD` and the dates) are dropped — only `units` is serialized. (`cost(sum(position))` and `value(sum(position))` correctly return `53 USD` / `55 USD` as scalar amounts, so the engine *has* the cost internally; it just isn't in the inventory wire format.)

## Request

Include the per-position cost in the Inventory serialization, e.g.:

```json
{"units": {"currency": "ITOT", "number": "2"},
 "cost":  {"number": "10", "currency": "USD", "date": "2024-01-01", "label": null}}
```

`cost` optional/null for positions held at cost `NONE`, to stay backward compatible. This mirrors the `CostNumber` shape already used elsewhere in the FFI types.

## Motivation

rustfava (the primary FFI consumer) flattens inventory columns to `{currency: units}` and explicitly cannot offer lot-level views (cost columns in a balances grid, holdings-by-lot reports) without this. Tracked downstream at rustledger/rustfava#155 (criterion 3) — that issue chose to document the units-only limitation now and unblock the lot-level path here when a consumer needs it.

## Notes

- Part of the broader FFI wire-format stability work — see #1395.
- Additive/optional field; existing units-only consumers keep working.

robcohen:
Closes #1398.

## What

`Inventory` and `Position` query result values now include an optional per-position `cost` object when the holding was booked at cost:

```json
{"units": {"number": "2", "currency": "ITOT"},
 "cost":  {"number": "10", "currency": "USD", "date": "2024-01-01", "label": null}}
```

`cost` is omitted entirely for holdings held without cost (booking method `NONE`).

## Why

The cost was already in hand — `rustledger_core::Position` carries `cost: Option<Cost>` and the query executor's `Value::Position` holds it — but `convert.rs::value_to_json` serialized `units` only. Consumers (notably rustfava) therefore couldn't reconstruct per-lot cost basis from an inventory column and had to collapse to units. See the downstream discussion in rustledger/rustfava#155.

## Shape note

A *booked* position carries a realized `core::Cost` (a flat per-unit `{number, currency, date, label}`), **not** the tagged `CostNumber` cost *spec* used on input postings — so the wire shape is a flat object, not a tagged enum. (This corrects the suggestion in the issue body.)

## Versioning

Additive, optional field → backward compatible. Units-only consumers ignore it. Per the `api_version` policy (major-on-break), this is a minor bump: **`API_VERSION` 2.0 → 2.1**, version-history entry added, plus the CHANGELOG and the `openrpc.json` example.

## Changes

- `convert.rs`: new `position_to_json` helper used by both the `Position` and `Inventory` arms; emits `cost` when present.
- `lib.rs`: `API_VERSION` → `2.1` + version-history entry.
- `CHANGELOG.md`, `openrpc.json`: doc/version sync.
- 3 unit tests in `convert::tests` (units-only, full cost, null date/label).

## How to test

```
cargo test -p rustledger-ffi-wasi          # 61 lib tests incl. convert::tests, all green
cargo clippy -p rustledger-ffi-wasi --all-targets   # clean
```

Self-contained in `rustledger-ffi-wasi`; no `core`/`query` changes.

### DIFF
--- a/crates/rustledger-ffi-wasi/CHANGELOG.md
+++ b/crates/rustledger-ffi-wasi/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Features
 
+- `Inventory` and `Position` query result values now include an optional
+  per-position `cost` object when the holding was booked at cost, using the
+  same wire shape as a directive `PostingCost` (`number` is a tagged
+  `CostNumber`, always `per_unit` for a booked position; absent `date`/`label`
+  omitted). Additive and backward compatible — units-only consumers ignore it
+  — so `api_version` is bumped to `2.1` (minor). Enables embedders to build
+  cost-/lot-aware views from query output. (#1398)
 - `ParseErrorEntry.kind_code` exposes the stable numeric discriminant
   from `rustledger_parser::ParseError::kind_code()`. Codes 1-26
   documented in `openrpc.json`'s `ParseErrorEntry` schema. Notably,
--- a/crates/rustledger-ffi-wasi/openrpc.json
+++ b/crates/rustledger-ffi-wasi/openrpc.json
@@ -57,7 +57,7 @@
           "result": {
             "name": "Result",
             "value": {
-              "api_version": "2.0",
+              "api_version": "2.1",
               "entries": [],
               "errors": [],
               "options": {},
--- a/crates/rustledger-ffi-wasi/src/convert.rs
+++ b/crates/rustledger-ffi-wasi/src/convert.rs
@@ -329,6 +329,42 @@ pub fn directive_to_json(directive: &Directive, line: u32, filename: &str) -> Di
     }
 }
 
+/// Serialize a position's units and, when present, its realized cost.
+///
+/// The `cost` field reuses the same wire shape as a directive [`PostingCost`]
+/// — `number` is a tagged [`CostNumber`], `currency`/`date`/`label` follow,
+/// and absent optional fields are omitted — so `cost.number` is uniform across
+/// the API and consumers switch on `kind` everywhere. A booked position always
+/// carries a concrete per-unit cost, so the kind is always `per_unit`. The
+/// `cost` field is omitted entirely for holdings held without cost (booking
+/// method `NONE`), leaving units-only consumers unaffected.
+fn position_to_json(p: &rustledger_core::Position) -> serde_json::Value {
+    use crate::types::output::CostNumber;
+
+    let mut obj = serde_json::json!({
+        "units": {
+            "number": p.units.number.to_string(),
+            "currency": p.units.currency
+        }
+    });
+    if let Some(cost) = &p.cost {
+        // Embed the typed CostNumber so the tagged shape stays in sync with
+        // PostingCost rather than being hand-rolled here.
+        let mut cost_obj = serde_json::json!({
+            "number": CostNumber::PerUnit { value: cost.number.to_string() },
+            "currency": cost.currency,
+        });
+        if let Some(date) = cost.date {
+            cost_obj["date"] = serde_json::Value::String(date.to_string());
+        }
+        if let Some(label) = &cost.label {
+            cost_obj["label"] = serde_json::Value::String(label.clone());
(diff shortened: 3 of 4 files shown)


robcohen:
## Self-review notes (adversarial pass)

Checked the three ways this could be subtly wrong; all clear, one test gap closed:

**1. Does the `api_version` 2.0 → 2.1 bump break a consumer?** No. The change is additive (optional field, units-only consumers ignore it). I also checked the primary consumer, rustfava: its `SUPPORTED_API_VERSION` constant is **never referenced** (dead code, and stale — it reads `"1."` while talking to a 2.x server), so there's no version gate to trip. Filed-worthy on the rustfava side, but irrelevant to this PR's safety.

**2. Is there a second position serializer that now diverges?** No. The only query-result inventory/position serializer is `value_to_json` (both arms updated via the shared `position_to_json` helper). The other `positions()` iterations in `commands/clamp.rs` consume positions *internally* to synthesize summary transactions (summing units for P&L), and the `"units"` literals there are test helpers — neither serializes query output, so no divergence.

**3. Golden/snapshot breakage?** None. ffi-wasi has no golden snapshot of inventory output; the only `positions`-bearing schema is `rustledger-wasm/bindings/` (a different crate/serialization path, unaffected). Full `rustledger-ffi-wasi` suite stays green.

**Shape correctness:** realized `core::Cost` is per-unit (`cost.rs`: "Cost per unit"), date renders ISO via jiff, and `Decimal::to_string()` matches the existing `units` formatting.

**Test gap closed:** the first revision only unit-tested the `position_to_json` helper. Added integration tests through `value_to_json` — including the literal #1398 scenario (two same-currency lots at different costs → two positions, each with its own cost, `len == 2` confirming distinct lots survive) and the scalar `Position` arm. 5 `convert::tests`, 63 lib tests total, clippy + fmt clean.

robcohen:
## Re-review revision: align cost shape with PostingCost

A second adversarial pass flagged a wire-consistency issue: the first revision serialized a query position's `cost.number` as a **plain string**, while a directive `PostingCost` serializes `cost.number` as a **tagged `CostNumber`** (`{"kind":"per_unit","value":...}`) — and `CostNumber`'s own doc tells consumers to *"switch on `kind`"*. Same field path, two shapes; a consumer's cost-number parser would break across contexts.

Fixed: a position's `cost` now reuses the exact `PostingCost` wire shape —

```json
"cost": {
  "number": {"kind": "per_unit", "value": "150.00"},
  "currency": "USD",
  "date": "2024-01-15",
  "label": "lot-a"
}
```

- `number` embeds the typed `CostNumber::PerUnit` (always `per_unit` for a booked position), so the tagged shape stays in sync with `PostingCost` rather than being hand-rolled.
- Absent `date`/`label` are now **omitted** (matching `PostingCost`'s `skip_serializing_if`), not `null`.
- Net effect: `cost.number` is monomorphic across the whole API — one parser everywhere.

Tests updated to assert the tagged shape (incl. the two-cost-lot inventory and field omission). 63 lib tests green, clippy + fmt clean. CHANGELOG and the `api_version` 2.1 version-history note updated.

PR #1418, opened 2026-06-17
fix(options): accept beancount 3.x use_precise_interpolation (#1416)

### LINKED ISSUE #1416: [Compat]: error[E7001]: Invalid option "use_precise_interpolation"
### Description

Beancount has recently introduced the new option `use_precise_interpolation`. See this [announcement](https://groups.google.com/g/beancount/c/uRxBIpvhkyg/m/UfzWOprbAwAJ). When present, the option is flagged as "invalid option" by rustledger and (presumably) not implemented.

### Beancount File

```text
option "use_precise_interpolation" "TRUE"
```

### Python Beancount Output

```text
$ bean-check foo.beancount
$
# no output, i.e., no error
```

### Rustledger Output

```text
$ rledger check foo.beancount 
error[E7001]: Invalid option "use_precise_interpolation"
✗ 1 error
$
```

### Python Beancount Version

3.2.3

### Rustledger Version

0.16.0

robcohen:
Closes #1416.

## Bug
`option "use_precise_interpolation" "TRUE"` (new in beancount 3.x) → `error[E7001]: Invalid option`, while `bean-check` accepts it silently.

## Cause
The loader's `KNOWN_OPTIONS` allowlist (`options.rs`) didn't include the new option, so the validator flagged it.

## Fix (option B — model the flag)
Mirror the existing tolerance/interpolation flags (`use_legacy_fixed_tolerances`, `experiment_explicit_tolerances`): add `use_precise_interpolation` to `KNOWN_OPTIONS`, add a parsed `use_precise_interpolation: bool` field + match arm.

**Why it's a no-op on results:** rustledger already interpolates missing amounts with exact `rust_decimal` arithmetic — it's effectively always in beancount's "precise" mode and has no legacy-imprecise path. So the flag is accepted and recorded but doesn't change booking output. This is the documented *"stricter / more correct than Python"* stance from the compat policy (we don't implement the legacy imprecise mode).

## Verification
- `rledger check` on `option "use_precise_interpolation" "TRUE"` → **✓ No errors found** (was E7001).
- Regression test `test_use_precise_interpolation_accepted` (no warning; field set).
- clippy + fmt clean.

### DIFF
--- a/crates/rustledger-loader/src/options.rs
+++ b/crates/rustledger-loader/src/options.rs
@@ -27,6 +27,7 @@ const KNOWN_OPTIONS: &[&str] = &[
     "infer_tolerance_from_cost",
     "use_legacy_fixed_tolerances",
     "experiment_explicit_tolerances",
+    "use_precise_interpolation",
     "booking_method",
     "render_commas",
     "display_precision",
@@ -132,6 +133,12 @@ pub struct Options {
     /// Enable experimental explicit tolerances in balance assertions.
     pub experiment_explicit_tolerances: bool,
 
+    /// Beancount 3.x `use_precise_interpolation` flag, parsed for compatibility.
+    /// rustledger always interpolates with exact `rust_decimal` arithmetic, so
+    /// this is effectively always-on; the field records the user's declared
+    /// value but does not change booking results (issue #1416).
+    pub use_precise_interpolation: bool,
+
     /// Default booking method.
     pub booking_method: String,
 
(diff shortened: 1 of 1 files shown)


robcohen:
Re: Copilot — fair consistency point. Keeping the lenient parse here intentionally: the field is a documented no-op (rustledger always interpolates with exact decimals), so the parsed value never affects results — only records the user's declaration. Strict value-validation/E7002 for the compat flags is a reasonable small follow-up but out of scope for this E7001 fix.

Issue #1384, opened 2026-06-15
Migrate rustledger-ffi-wasi from WASI p1 (JSON-RPC) to p2 / Component Model (WIT)


robcohen:
## Status: spike complete → this is now the production migration plan

The spike is **done and the decision is made**. A WIT contract → `wasm32-wasip2` component → wasmtime host round-trip works end-to-end, including a real tagged **variant** (`cost-number`, the exact shape that breaks rustfava) — so the risky part, type modeling, is de-risked, not just the plumbing. Decision: **dual-ship**.

- Spike artifacts: branch `spike/ffi-wit-component`, tag `ffi-wit-spike-1384` (commit `1b2db5a0`), PR #1396 (closed unmerged, kept as reference per spike best-practice).
- Consumer audit (done): the **only** consumer of `rustledger-ffi-wasi-<ver>.wasm` is **rustfava** (`engine.py` downloads it and drives JSON-RPC over the wasmtime CLI) — which is also what breaks on every wire-format drift.

This issue now tracks building it.

## Why now

The JSON wire shape is hand-maintained across mirrored DTOs (`types/output.rs` ⇄ `types/input.rs` ⇄ `convert.rs`) and policed by "any silent shape change is a wire break" round-trip tests. It still drifts, and rustfava breaks on it (the v0.11→v0.16 cost-shape change; tracked at #1395). Most recent concrete footgun: **#1399** — adding per-position `cost` to query output required getting `cost.number` to match the tagged `CostNumber` shape used for directive postings *by hand and by review*. Under a WIT contract that consistency is enforced by generated bindings — the bug is unrepresentable. A `.wit` becomes the single source of truth; embedders get type-checked bindings instead of hand-parsing JSON into `Decimal`s.

## Architecture (dual-ship)

1. Keep the wasip1 JSON-RPC module unchanged — don't break the current rustfava integration.
2. Add a wasip2 WIT component as a parallel release artifact.
3. Migrate rustfava to the component.
4. Deprecate the wasip1 surface after a window.

Host cost is small: `wasmtime` 45 (already a workspace dep) has full Component Model support; `wasm32-wasip2` emits a component directly (no `cargo-component`); `wit-bindgen` guest macro + `wasmtime::component::bindgen!` host both read one `.wit`. (Spike findings.)

### Phase 0 — Decision (this issue) ✅
- [x] Feasibility proven (string + tagged variant round-trip).
- [x] Consumer audit — rustfava only.
[...]

PR #1421, opened 2026-06-17
feat(ffi-component): stateful resource session ledger handle (#1384)

robcohen:
Phase 1 of the rustfava↔rustledger boundary redesign (rustfava #173, part of #1384). De-risked first by the `spike/ledger-resource` spike; this is the production contract + impl.

## What
A Component-Model **`resource session`** in the WIT: `constructor(source)` / `from-file(path, …)` build a handle that holds the booked ledger *inside the component*, and the methods run against it:

| method | returns |
|---|---|
| `info()` | `load-result` (entries/errors/options/plugins/includes) — materialized once for the host |
| `query(q)` | `query-result` — BQL against the held ledger |
| `filter(begin, end)` | `list<directive>` within `[begin, end)` |
| `clamp(begin, end)` | `list<directive>`, opening-balance + summary synthesized |

## Why
Embedders (rustfava) today re-parse `source` — or re-serialize already-loaded entries back to beancount text — on every query/clamp, and shuttle directive lists across the FFI boundary per op. The handle removes both: **parse + book once, operate server-side.**

Critically, `clamp`/`filter` run on the held **core** directives. `clamp` calls `rustledger_ops::clamp` directly rather than round-tripping `WIT → core → WIT`, which is faster *and* sidesteps the shape-translation drift that round-trip caused on the host (rustfava #173 — the cost-basis / typed-value bug class).

## How
Reuses the existing converters wholesale via a normalized `SessionState` that flattens the two load paths (`load_source`'s `LoadResult` vs `load_file`'s `FileLoad`) into one held representation, so the methods don't care which produced it. The free `load`/`query`/`clamp` functions stay for back-compat during the dual-ship window.

## Tests
New host parity test `session_clamp_preserves_cost_basis` (construct → info → query → clamp; asserts the per-unit cost survives clamp) alongside the existing 13 — **14/14 green**. clippy (`-D warnings`, wasip2) + fmt clean.

Next (rustfava #173): the rustfava adapter consumes this handle, then the full-suite gate, then the default flip.

### DIFF
--- a/crates/rustledger-ffi-component-tests/tests/parity.rs
+++ b/crates/rustledger-ffi-component-tests/tests/parity.rs
@@ -494,3 +494,67 @@ fn load_file_runs_auto_accounts_synth() -> Result<()> {
     assert_generated_opens(&loaded.entries, "component load_file");
     Ok(())
 }
+
+const LEDGER_WITH_COST: &str = "\
+2023-01-01 open Assets:Cash USD
+2023-01-01 open Expenses:Food USD
+2023-06-15 * \"old\"
+  Expenses:Food  5 USD
+  Assets:Cash
+2024-02-10 * \"Coffee\"
+  Expenses:Food  7 USD {2 USD}
+  Assets:Cash  -14 USD
+";
+
+/// The stateful `resource session` (#173): construct once, then info/query/
+/// clamp run against the held ledger. Crucially `clamp` operates on the held
+/// core directives, so cost basis survives with no WIT->core->WIT round-trip.
+#[test]
+fn session_clamp_preserves_cost_basis() -> Result<()> {
+    use rustledger::ledger::types::{CostNumber, Directive};
+
+    if !component_path().exists() {
+        eprintln!("skip: component wasm not built");
+        return Ok(());
+    }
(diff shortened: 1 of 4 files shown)


robcohen:
Thanks @copilot — addressed in fd35930a:

- **`#173` ambiguity** (several comments): qualified to `rustfava#173` in the WIT + code comments, since bare `#173` is an unrelated PR in this repo.
- **`SessionState::query` re-running `merge_with_padding`**: now cached in a `OnceCell`, computed on first query.
- **`SessionState::filter` materializing all directives then filtering**: keeping as-is for now — `filter` is a lossless date predicate, and reusing the existing WIT-level `filter` avoids duplicating the open/close/commodity boundary logic on core directives. The win the resource targets (no host round-trip, no re-parse) already holds; a core-side filter is a self-contained later optimization.
- **api-version bump for the new surface**: intentionally staying at `2.1`. The `resource session` is purely additive and back-compat, the version string tracks the shared wire contract with `rustledger-ffi-wasi` (also 2.1), and a host detects the resource by using it rather than by version. Bumping only the component would desync the two surfaces' advertised version.

PR #1423, opened 2026-06-17
feat(ffi-component): builder.query-entries — query a loaded directive set

robcohen:
Adds `builder.query-entries(entries: list<directive>, query: string) -> query-result` — the typed counterpart to the existing `filter`/`clamp`. The embedder passes the directives it already holds and gets BQL results, with **no re-parse of source and no re-render to beancount text**.

## Why
rustfava runs queries over *filtered* entry subsets (rustfava #173). Its options were both bad: re-parse the original source (ignores the filter) or re-render the subset to beancount text and re-parse it (fragile — invalid rendered text crashes the parser, the actual bug). `query-entries` gives it the same typed entry-set path it already uses for `clamp`, so **query/filter/clamp are symmetric over the WIT contract** — no source text round-trip anywhere.

## Impl
Mirrors `clamp`'s `WIT → core` conversion (`loaded_directive_to_input` + `input_entry_to_directive`), then pad-expands (`merge_with_padding`) and runs the query — matching what the source-based `query` does.

## Tests
New host parity test `query_entries_matches_source_query` asserts it matches the source `query`'s row count. clippy (`-D warnings`, wasip2) + fmt clean; 15/15 host parity tests pass.

Next: cut a release so rustfava can consume it, then reroute `query_shell` through it and delete `_entries_to_source` (rustfava #173).

### DIFF
--- a/crates/rustledger-ffi-component-tests/tests/parity.rs
+++ b/crates/rustledger-ffi-component-tests/tests/parity.rs
@@ -558,3 +558,37 @@ fn session_clamp_preserves_cost_basis() -> Result<()> {
     handle.resource_drop(&mut store)?;
     Ok(())
 }
+
+/// `builder.query-entries` (rustfava#173): query an already-loaded directive
+/// set directly, matching the source-based `query` — the typed alternative to
+/// re-rendering entries to source.
+#[test]
+fn query_entries_matches_source_query() -> Result<()> {
+    if !component_path().exists() {
+        eprintln!("skip: component wasm not built");
+        return Ok(());
+    }
+    let (mut store, inst) = instantiate()?;
+    let q = "SELECT account, position";
+    let loaded = inst
+        .rustledger_ledger_ledger()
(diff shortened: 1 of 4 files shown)


robcohen:
Thanks @copilot:

- **`filter_map(...).ok()` dropping unconvertible directives**: intentional and consistent with the sibling `clamp`/`filter` (same pattern). The inputs are directives the component itself produced and the host round-trips back, so `input_entry_to_directive` should not fail in practice; the `.ok()` is defensive, and matching `clamp` keeps the three entry-set ops uniform. Surfacing a per-directive conversion error would be a separate change applied to all three.
- **WIT package version at 2.1.0**: kept intentionally, same call as #1421 (which added the `resource session` at 2.1.0). `query-entries` is purely additive and back-compat; the runtime `api-version` string handles negotiation, and it tracks the shared contract with `rustledger-ffi-wasi`. We'd bump deliberately across both surfaces, not per additive function.

PR #1445, opened 2026-06-20
fix(parser): decode string-literal escapes at parse

robcohen:
Found by dogfooding. String literals stored their **raw escaped bytes**, so a narration `"Dinner at \"The Place\""` kept the backslashes — queries displayed `Dinner at \"The Place\"` and `WHERE narration = "..."` never matched. Closes #4 (dogfooding tracker).

## Approach
The architectural key: rledger's **formatter is CST-based** (`format_source` works on the green tree / raw source), so decoding the *core* value is cleanly separate from formatting — the journal keeps raw source, only the semantic value decodes. Beancount's de-escaping rules were verified directly (incl. the unknown-escape case).

## Changes
- **`StringLit::text_decoded`** decodes escapes (`\"`→`"`, `\\`→`\`, `\n`,`\t`,`\r`; unknown `\x`→`x`, matching Beancount's lexer). All CST→core string extractions use it — narration, payee, meta values, note, document, event, query, custom, option, booking.
- **Directive `Display` impls re-escape** via `format::escape_string`, so the `-f beancount` query output and directive printing stay valid Beancount. The CST formatter is untouched, so the `format_compat` "preserve raw string interior" contract still holds.
- **`CACHE_VERSION` 9 → 10**: the stored string bytes changed, so old parse caches regenerate instead of serving stale escaped strings (this staleness was caught during testing).

## Verification
| input | before | after |
|---|---|---|
| `SELECT narration` of `"a \"b\""` | `a \"b\"` | `a "b"` |
| `WHERE narration = "a \"b\""` | 0 rows | matches |
| `-f beancount` output | — | re-escaped (valid Beancount) |

Beancount oracle confirmed: `\x`→`x` (drop), `\r`→CR, `\t`→tab, `\n`→newline. New test `string_escapes_are_decoded`; `format_compat` (55 tests) green — the CST formatter is unaffected; parser + core suites green.

This touches the parser, so it warrants the fuzz + compat scrutiny in CI.


robcohen:
Addressed all three Copilot findings in 0b767cf0:
- **cache.rs FIXTURE_VERSION**: bumped 9→10 (v10 does not touch the `CostNumber` archived layout, so the byte fixtures stay valid). This was the actual CI failure.
- **directive.rs / escape_string**: now re-escapes `\t`→`\\t` and `\r`→`\\r` so Display no longer emits raw control bytes inside quotes; added a unit test.
- **cache.rs v8 note**: the "bump to v10" guidance now points to v11.
Plus 97a5a6ac regenerates the parser-corpus baseline manifest (17 escaped-string fixtures, parser hashes only).

PR #1446, opened 2026-06-20
fix(check): match bean-check on close-with-balance; move advisory to lint

robcohen:
## What

`rledger check` emitted **E1004 / AccountCloseNotEmpty** when an account was closed while still holding a non-zero balance. Python beancount's `bean-check` (3.2.3) never emits this — `close` records that an account is inactive from a date; it does not require a zero balance. So `check` was stricter than beancount on valid input, including our own `valid-ledger.beancount` fixture (which the `check_clean` snapshot test was asserting *with* a warning).

Per the **Python Compatibility Policy** (`check` mirrors `bean-check`), this moves the advisory out of `check` into a dedicated lint rather than dropping it — the "you closed an account with money still in it" nudge is genuinely useful, just not a `check`-level concern.

## How

- `ErrorCode::is_advisory_only()` marks **E1004** as advisory-only.
- `check` skips advisory-only codes in both its rendered report and its warning count (`is_advisory_only_code`), so `valid-ledger.beancount` now checks clean — matching `bean-check`.
- The detection is **unchanged** in `rustledger-validate`, so the LSP/editor still surfaces the advisory.
- New **`rledger lint closed-nonempty`** reuses the canonical `process(validate: true)` pipeline and keeps only E1004, with text and JSON output.

## Verify

```
# before: E1004 warning; after: clean (matches bean-check)
rledger check valid-ledger.beancount

# the advisory now lives here
rledger lint closed-nonempty ledger.beancount
# -> ledger.beancount:7: Cannot close account Assets:Cash with non-zero balance
#    1 account(s) closed with a non-zero balance.
```

- `check_clean_{text,json}` snapshots updated → "✓ No errors found", `warning_count: 0`.
- Regression tests in `closed_nonempty.rs`: reports a balance-bearing close; silent on a zero-balance close.
- `cargo test -p rustledger -p rustledger-validate`, clippy, and fmt all green.

### DIFF
--- a/crates/rustledger-validate/src/error.rs
+++ b/crates/rustledger-validate/src/error.rs
@@ -175,6 +175,16 @@ impl ErrorCode {
     }
 }
 
+/// Whether a rendered diagnostic code string (e.g. `"E1004"`) is advisory-only.
+///
+/// The string-keyed counterpart to [`ErrorCode::is_advisory_only`], for
+/// consumers (the CLI `check`/`lint` split) that only carry the code string.
+/// Keeping it here means the set of advisory-only codes lives in one place.
+#[must_use]
+pub fn is_advisory_only_code(code: &str) -> bool {
+    code == ErrorCode::AccountCloseNotEmpty.code()
+}
+
 /// Severity level for validation messages.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Severity {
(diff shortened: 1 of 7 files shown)


robcohen:
Addressed all three Copilot findings in e1d5bc87:
- **closed_nonempty.rs (clearing errors)**: the lint now bails on a file with load errors instead of silently reporting "no accounts"; added a regression test (missing include → hard error).
- **check.rs / closed_nonempty.rs (hardcoded E1004)**: replaced both literals with the canonical `rustledger_validate::is_advisory_only_code()` and `ErrorCode::AccountCloseNotEmpty.code()` — no more code-string drift.

PR #1447, opened 2026-06-20
fix(validate): resolve relative document paths against the directive's file

robcohen:
## What

A relative `document` path was validated with a **CWD-relative** `exists()` whenever neither `document_base` nor the `documents` option was set. So `rledger check /abs/ledger.beancount` reported a document as present or missing depending on the directory you ran it from:

```
$ cd /tmp && rledger check /abs/top.beancount     # ✗ Document file not found: receipt.pdf
$ cd /abs  && rledger check /abs/top.beancount     # ✓ No errors found
```

Python beancount normalizes the path at parse time against the **directive's own file**, and rledger already does this for `include`. `bean-check 3.2.3` is silent on the repro from any directory.

## How

Resolve the fallback against the directive's own source-file directory:

- **`ValidationOptions::document_source_dirs`** — a per-`file_id` list of source-file directories (`file_id` indexes `source_map`), populated by the loader's `build_validation_options`.
- **`document_file_exists`** — one shared resolution chain (absolute → `document_base` → `documents` dirs → **directive's source dir** → CWD), used by both the pre-pass cache and the validator, so the two can't drift.
- The exists-cache is keyed by **`(path, file_id)`** because resolution now depends on the source file — the same relative path in two differently-located files can resolve to different files.

The stored `doc.path` is unchanged (no round-trip impact); only the existence check resolves correctly.

## Verify

```
mkdir -p /tmp/d && cd /tmp/d
printf '2024-01-01 open Assets:Cash\n2024-01-02 document Assets:Cash "receipt.pdf"\n' > t.beancount
echo x > receipt.pdf
cd / && rledger check /tmp/d/t.beancount        # before: not found | after: ✓
```

- Two regression tests (`loader_test.rs`): relative path resolves against the ledger dir; **a document in an included file resolves against that file's dir** (exercises per-`file_id`).
- A genuinely-missing document still errors.
- Full `rustledger-loader` (66) + `rustledger-validate` (89) suites, clippy, and fmt all green.

Note: the LSP builds `ValidationOptions` through a separate path and is unchanged by this PR (no regression); wiring `document_source_dirs` there is a possible follow-up.


robcohen:
Addressed all four Copilot findings in the latest commit:
- **Cache key allocations**: the exists-cache is now keyed by `(&str, Option<u16>)` borrowed from the `directives` slice; neither the cache build nor the validator lookup allocates/clones a `String` per document.
- **Brittle test assertions**: both regression tests now assert on the stable error code `E8001` (DocumentNotFound) instead of the error-message substring.

PR #1449, opened 2026-06-20
fix(validate): anchor E2003 (unused pad) to the pad source line

robcohen:
## What

An unconsumed `pad` directive emitted its E2003 with **no source location**:

```
$ rledger check t.beancount
<unknown>: error[E2003]: Unused Pad entry
```

while sibling errors (E2001, etc.) render a full miette report with file/line and carets. The project's own self-review checklist requires file-location context on errors.

## Why

E2003 is **deferred** — `check_unused_pads` runs after both validation phases, walking leftover `PendingPad`s. But `PendingPad` kept only the pad's account/date/currencies, never its span or `file_id`, so the deferred error had nothing to anchor to.

## How

- `PendingPad` now carries `location: Option<(Span, u16)>`.
- `validate_pad` takes the pad's `span_info()` from the dispatch and stores it.
- `check_unused_pads` sets `span`/`file_id` on the E2003 from the `PendingPad`.

It now renders with the pad's own line:

```
E2003
  x Unused Pad entry
   ,-[t.beancount:3:1]
```

## Verify

```
printf '2024-01-01 open Assets:Checking USD\n2024-01-01 open Equity:Opening USD\n2024-03-15 pad Assets:Checking Equity:Opening\n' > t.beancount
rledger check t.beancount   # before: <unknown>: ... | after: t.beancount:3:1
```

Regression test `test_unused_pad_error_has_source_location` asserts the E2003 `LedgerError` carries a location pointing at the pad's line. `rustledger-validate` + `rustledger-loader` suites, clippy, fmt all green.

### DIFF
--- a/crates/rustledger-loader/tests/loader_test.rs
+++ b/crates/rustledger-loader/tests/loader_test.rs
@@ -988,6 +988,37 @@ fn test_relative_document_in_included_file_resolves_against_that_file() {
 }
 
 #[test]
+fn test_unused_pad_error_has_source_location() {
+    // Regression: E2003 (Unused Pad) must be anchored to the pad's own line,
+    // not rendered as `<unknown>:`. The pad below is never consumed by a
+    // following balance assertion.
+    let dir = tempfile::tempdir().unwrap();
+    let path = dir.path().join("ledger.beancount");
+    std::fs::write(
+        &path,
+        "2024-01-01 open Assets:Checking USD\n\
+         2024-01-01 open Equity:Opening USD\n\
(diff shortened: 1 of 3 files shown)


robcohen:
Addressed the Copilot finding: a plugin-synthesized pad (file_id == SYNTHESIZED_FILE_ID) now gets the same advisory note on its deferred E2003 that the per-directive patching loop attaches to in-phase errors. The note text is shared via a new SYNTHESIZED_DIRECTIVE_NOTE const so the two paths cannot drift.

PR #1450, opened 2026-06-20
fix(importer): drop doubled Row N: prefix in extract date warnings

robcohen:
## What

Per-row extract warnings for the two date errors carried a **doubled** `Row N:` prefix:

```
$ rledger extract h.csv -a Assets:Bank --date-column NoSuchCol ...
warning: Row 1: Row 1: missing date column
```

`csv_importer::parse_row` pre-attached `Row {n}: ...` to its date errors (missing date column, failed-to-parse date), and its **sole caller** `extract` then wrapped *every* `parse_row` error again with `Row {row_num}: {e}`.

## How

Drop the inner prefixes; the outer wrap already supplies exactly one `Row N:` for every parse_row error. The now-unused `row_num` parameter is removed from `parse_row`.

## Verify

```
printf 'date,desc,amt\n2024-01-01,Coffee,-4.50\n' > h.csv
rledger extract h.csv -a Assets:Bank --date-column NoSuchCol --narration-column desc --amount-column amt
# before: warning: Row 1: Row 1: missing date column | after: warning: Row 1: missing date column
```

`test_csv_import_missing_column_error` now asserts the warning carries exactly one `Row ` prefix; full importer lib suite (137) green.

### DIFF
--- a/crates/rustledger-importer/src/csv_importer.rs
+++ b/crates/rustledger-importer/src/csv_importer.rs
@@ -111,14 +111,7 @@ impl CsvImporter {
                 }
             };
 
-            match self.parse_row(
-                &record,
-                config,
-                csv_config,
-                &amount_format,
-                &header_map,
-                row_num,
-            ) {
+            match self.parse_row(&record, config, csv_config, &amount_format, &header_map) {
                 Ok(Some(txn)) => directives.push(Directive::Transaction(txn)),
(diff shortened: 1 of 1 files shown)


robcohen:
Reworded the comment to refer to parse_row's caller generically rather than naming a specific function, per the suggestion.

PR #1451, opened 2026-06-20
fix(compat): exec rledger by absolute path in bean-* wrappers

robcohen:
## What

The generated `bean-*` wrapper scripts ran `exec rledger {sub} "$@"` (and `@rledger {sub} %*` on Windows), calling rledger by **bare name**. With `--prefix` installing into a directory that isn't on PATH — which the install step itself warns about — every shim died:

```
$ rledger compat install --prefix /tmp/bin
$ env PATH=/usr/bin:/bin /tmp/bin/bean-check main.beancount
exec: rledger: not found        # exit 127
```

## How

Exec the **absolute `current_exe()` path** instead of a bare name (POSIX single-quoted on Unix, double-quoted on Windows). A `rledger compat wrapper` marker line keeps `is_rledger_wrapper` recognizing the file regardless of the binary's name.

```sh
#!/bin/sh
# rledger compat wrapper
exec '/usr/local/bin/rledger' check "$@"
```

## Verify

```
rledger compat install --prefix /tmp/bin
env PATH=/usr/bin:/bin /tmp/bin/bean-check main.beancount   # before: exit 127 | after: ✓ runs
```

Tests updated (`test_wrapper_content_unix` asserts the absolute path is execed and no bare `rledger`; the overwrite test no longer pins the old `rledger check` substring). compat suite (7), clippy, fmt green.

### DIFF
--- a/crates/rustledger/src/cmd/compat.rs
+++ b/crates/rustledger/src/cmd/compat.rs
@@ -38,11 +38,16 @@ fn resolve_target_dir(prefix: Option<&Path>) -> Result<PathBuf> {
     Ok(dir.to_path_buf())
 }
 
-/// POSIX single-quote a path so it survives in the generated `sh` wrapper even
-/// when it contains spaces or shell metacharacters.
+/// Marker line written into every generated wrapper so [`is_rledger_wrapper`]
+/// can recognize one precisely, instead of treating any file that merely
+/// mentions "rledger" as a wrapper.
+const WRAPPER_MARKER: &str = "rledger compat wrapper";
+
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed all five findings: UTF-8-validate the path once and pass a `&str` to `wrapper_content` (no lossy corruption); compute `current_exe()` a single time and derive the dir from it; `is_rledger_wrapper` now matches the `rledger compat wrapper` marker (plus legacy `exec rledger`/`@rledger` for upgrades) rather than any file mentioning rledger; the overwrite-test exec-line assertion is now platform-specific; added a `#[cfg(windows)]` wrapper test and a spaces-in-path test.

PR #1452, opened 2026-06-20
fix(extract): fail loudly when no transactions are extracted

robcohen:
## What

`extract` printed `Extracted 0 transactions` and exited **0** regardless of the result, so a garbage / wrong-type / empty file silently produced nothing:

```
$ printf 'garbage not a csv\nsecond line\n' > foo.txt
$ rledger extract foo.txt -a Assets:Bank; echo $?
Extracted 0 transactions from foo.txt
0
```

A scripted `extract … >> ledger.beancount` therefore appends nothing **silently** on a malformed or wrong-path-but-existing file (only a *nonexistent* path exited non-zero).

## How

Bail with an actionable message when the result is empty (before writing output), so the exit code is non-zero:

```
error: no transactions were extracted from foo.txt
  the file may not match a recognized importer format, be empty, or use unexpected columns
  try: --auto, an explicit importer (--importer), or column flags (--date-column, --amount-column, …)
```

## Verify

```
printf 'garbage not a csv\n' > foo.txt
rledger extract foo.txt -a Assets:Bank; echo $?   # before: exit 0 | after: error + non-zero
```

A valid CSV still extracts normally (regression guarded). Regression test `test_extract_empty_result_exits_nonzero` asserts a non-importable file exits non-zero with the explanation.

### DIFF
--- a/crates/rustledger/src/cmd/extract_cmd/mod.rs
+++ b/crates/rustledger/src/cmd/extract_cmd/mod.rs
@@ -827,6 +827,21 @@ pub fn run_with_writer<W: Write>(args: &Args, file: &Path, out: &mut W) -> Resul
     let formatted = canonicalize_directives(directives.iter(), &fmt_config)
         .map_err(|e| anyhow::anyhow!(e.to_string()))?;
 
+    // Fail loudly on an empty result instead of printing "Extracted 0
+    // transactions" and exiting 0. A garbage / wrong-type / empty file
+    // otherwise makes a scripted `extract … >> ledger.beancount` silently
+    // append nothing.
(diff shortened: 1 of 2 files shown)


robcohen:
Addressed both findings: the empty-result check now counts EXTRACTED transactions right after `importer.extract()` and bails only when that is zero — before `--existing` dedup or `--balance` augmentation can mask the cause; the "Extracted N" line now counts `Directive::Transaction` values, not `directives.len()`. The regression test uses OsStr-safe `.arg(file.path())`. Also merged main (resolving the cli_commands_test.rs append conflict).

PR #1453, opened 2026-06-20
fix(query): name trailing tokens instead of mislabeling as EOF

robcohen:
## What

The query parser reported **"unexpected end of input"** for *any* trailing tokens after a valid prefix:

```
$ rledger query main.beancount "SELECT account, sum(position) GROUP BY account WHERE number > 0"
syntax error at position 47: unexpected end of input        # 47 is the WHERE token
$ rledger query main.beancount "SELECT account FOOBAR"
syntax error: unexpected end of input                       # at FOOBAR
```

The classifier used `e.found().is_none()` to mean EOF — but chumsky's `end()` combinator (which rejects leftover tokens) *also* reports `found() == None`, so trailing tokens were mislabeled even though the error span pointed straight at them.

## How

Classify by the **error span position** instead: a position at or past the end of the input is a genuine EOF; otherwise name the unexpected token at that position.

```
syntax error at position 47: unexpected token 'WHERE'
syntax error at position 15: unexpected token 'FOOBAR'
```

## Verify

```
rledger query f.beancount "SELECT account, sum(position) GROUP BY account WHERE number > 0"   # names WHERE@47
rledger query f.beancount "SELECT account FOOBAR"                                              # names FOOBAR@15
```

Test `trailing_tokens_are_named_not_mislabeled_eof`; full `rustledger-query` suite, clippy, fmt green.

### DIFF
--- a/crates/rustledger-query/src/parser.rs
+++ b/crates/rustledger-query/src/parser.rs
@@ -1082,13 +1082,16 @@ mod tests {
         ] {
             let err = parse(q).expect_err("should be a parse error");
             assert_eq!(err.position, pos, "span should point at the token in {q:?}");
-            match err.kind {
-                ParseErrorKind::SyntaxError(ref m) => assert!(
-                    m.contains(token),
-                    "error should name {token:?}, got {m:?} for {q:?}"
-                ),
-                other => panic!("expected SyntaxError naming {token:?}, got {other:?} for {q:?}"),
-            }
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed both: the classifier now only special-cases the ambiguous `found() == None` situation (EOF vs end()-trailing-tokens) and keeps chumsky's richer message when a concrete token is found; renamed the test to the `test_` prefix.

PR #1454, opened 2026-06-20
fix(add): name the missing-currency error in quick mode

robcohen:
## What

In `add` quick mode, a bare or split amount became a phantom **account** and produced a misleading error:

```
$ rledger add t.beancount -y --quick 'Store' 'Lunch' Expenses:Food 5.00 Assets:Cash
error: Quick mode supports at most one posting without an explicit amount, but 3 postings lack amounts.
```

The account/amount loop calls `parse_amount` on each post-account token and, on failure, silently treats it as the next account. `parse_amount` rejects a currency-less number, so a bare `5.00` (or a `5.00 USD` amount the shell split into two args) slipped through as an account.

## How

Detect an amount-looking token in the account position (`looks_like_amount`) and bail with a message that names the real cause:

```
error: '5.00' looks like an amount but appears where an account name is expected.
  Amounts need an explicit currency (e.g. '5.00 USD') and must be a single argument —
  quote amounts that contain a space (e.g. '5.00 USD') so the shell does not split them.
```

The `--quick` help text now documents the quoting and FILE-before-`--quick` / `--` requirements.

## Verify

```
rledger add t.beancount -y --quick 'Store' 'Lunch' Expenses:Food 5.00 Assets:Cash       # clear error
rledger add t.beancount -n --quick 'Store' 'Lunch' Expenses:Food '5.00 USD' Assets:Cash  # works
```

Test `looks_like_amount_distinguishes_amounts_from_accounts`.

### DIFF
--- a/crates/rustledger/src/cmd/add_cmd/mod.rs
+++ b/crates/rustledger/src/cmd/add_cmd/mod.rs
@@ -74,6 +74,13 @@ pub struct Args {
     pub yes: bool,
 
     /// Quick mode: payee narration account amount \[account \[amount\]\]...
+    ///
+    /// Each amount needs an explicit currency and must be a SINGLE argument, so
+    /// quote it: `'5.00 USD'` (otherwise the shell splits it into `5.00` and
+    /// `USD`). Because `--quick` is variadic it consumes every following
+    /// argument, so pass the FILE before `--quick`, or end the quick args with
+    /// `--`. Example:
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed all three: looks_like_amount now matches sign+decimal forms (-.5/+.5) and its doc matches the numeric-prefix implementation; the error message uses a fixed "5.00 USD" example (no more interpolating a token that may already include a currency); and the --quick help examples use double quotes for cross-shell (incl. cmd.exe) compatibility.

PR #1455, opened 2026-06-20
feat(extract): make the merchant dictionary reachable via --use-merchant-dict

robcohen:
## What

rledger ships a 76-pattern merchant dictionary (`rustledger-ops::merchants`) that maps known merchants to expense categories, but it was **unreachable from the CLI**: `use_merchant_dict` defaulted false and was only settable via the library `CsvConfigBuilder`. So every documented merchant (NETFLIX, AMAZON, SHELL, …) stayed at `Expenses:Unknown`:

```
$ printf 'Date,Description,Amount\n2024-03-08,NETFLIX.COM,-15.99\n' > m.csv
$ rledger extract m.csv -a Assets:Bank --date-column Date --narration-column Description --amount-column Amount
  Expenses:Unknown
```

(Worse: even with `use_merchant_dict` on, the regular extract path only produced enrichment *metadata* — it never applied the matched account.)

## How

- **`match_mapping`** (the regular path's account assignment) now consults `MERCHANT_PATTERNS` (case-insensitive substring), **after** user mappings, when `use_merchant_dict` is set — so the dictionary's account is actually assigned.
- A new **`--use-merchant-dict`** flag (wired into the CLI-flags and `--auto` paths) and an **`importers.toml` `use_merchant_dict`** field expose it.

```
$ rledger extract m.csv -a Assets:Bank --use-merchant-dict --date-column Date --narration-column Description --amount-column Amount
  Expenses:Subscriptions:Streaming
```

## Verify

Regression test `test_merchant_dict_categorizes_account_when_enabled` asserts NETFLIX → `Expenses:Subscriptions:Streaming` with the flag, `Expenses:Unknown` without. Full importer suite (138), clippy, fmt green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3351,6 +3351,7 @@ dependencies = [
  "jiff",
  "ofxy",
  "proptest",
+ "regex",
  "rmp-serde",
  "rust_decimal",
  "rustledger-core",
--- a/crates/rustledger-importer/Cargo.toml
+++ b/crates/rustledger-importer/Cargo.toml
@@ -36,6 +36,7 @@ rustledger-plugin = { workspace = true, features = ["wasm-runtime"], optional = 
 rustledger-plugin-types.workspace = true
 jiff.workspace = true
 rust_decimal.workspace = true
+regex.workspace = true
 anyhow.workspace = true
 format_num_pattern.workspace = true
 csv.workspace = true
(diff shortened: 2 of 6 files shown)


robcohen:
Addressed all three: the merchant dictionary is now matched with a precompiled, case-insensitive, unanchored regex table (built once via LazyLock), mirroring `RulesEngine::load_merchant_dict` — so alternation patterns like `AMAZON|AMZN` match instead of being silently missed by substring matching (test extended to cover this); `match_mapping` no longer lowercases payee/narration when there are no user mappings; and the help text now shows the real account `Expenses:Subscriptions:Streaming`.

Issue #1457, opened 2026-06-20
OFX import: 2.x (XML) unsupported and 1.x header parsing is overly strict


robcohen:
`extract` uses the `ofxy` crate for OFX parsing, which has two real-world gaps that block bank statements the import surface should accept. Both surface as clean errors (no panic), and both are rooted in `ofxy`, not rledger's own code. Filing together since they share a root cause and fix area (`crates/rustledger-importer/src/ofx_importer.rs`).

## 1. OFX 2.x (XML) is unsupported

Modern banks, Quicken, and many European exporters emit OFX 2.x (XML-based, `VERSION="200"+`). `ofxy` rejects it:

```
$ rledger extract statement.ofx -a Assets:Bank
error: Failed to parse OFX content: parse error: Ofxy does not yet support OFX version: 200
```

Minimal repro:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<?OFX OFXHEADER="200" VERSION="200" SECURITY="NONE" OLDFILEUID="NONE" NEWFILEUID="NONE"?>
<OFX><BANKMSGSRSV1><STMTTRNRS><TRNUID>1</TRNUID><STATUS><CODE>0</CODE><SEVERITY>INFO</SEVERITY></STATUS>
<STMTRS><CURDEF>USD</CURDEF><BANKACCTFROM><BANKID>123</BANKID><ACCTID>456</ACCTID><ACCTTYPE>CHECKING</ACCTTYPE></BANKACCTFROM>
<BANKTRANLIST><DTSTART>20240101</DTSTART><DTEND>20240131</DTEND>
<STMTTRN><TRNTYPE>DEBIT</TRNTYPE><DTPOSTED>20240115</DTPOSTED><TRNAMT>-50.00</TRNAMT><FITID>t1</FITID><NAME>COFFEE</NAME></STMTTRN>
</BANKTRANLIST></STMTRS></STMTTRNRS></BANKMSGSRSV1></OFX>
```

## 2. OFX 1.x header strictness

`ofxy` requires the *full* canonical OFX 1.x header block — `CHARSET`, `COMPRESSION`, `OLDFILEUID`, `NEWFILEUID` must all be present. A statement missing any one fails:

```
$ rledger extract statement.ofx -a Assets:Bank
error: Failed to parse OFX content: parse error: headers missing 'CHARSET'
# add CHARSET -> next failure: headers missing 'OLDFILEUID', etc.
```

[...]

PR #1458, opened 2026-06-20
fix(import): normalize OFX header to accept sparse 1.x and 2.x XML files

robcohen:
## What

Addresses **#1457**: `extract` rejected two real-world OFX shapes because the `ofxy` crate enforces a strict, complete OFX 1.x SGML header.

- **1.x missing headers** (item 2): banks omit `CHARSET`/`OLDFILEUID`/`NEWFILEUID` → `parse error: headers missing 'CHARSET'`.
- **2.x XML** (item 1): `parse error: Ofxy does not yet support OFX version: 200`.

## How

A small preprocessing shim, `normalize_ofx_header`, run before `content.parse::<ofxy::Ofx>()`. `ofxy` splits the file at `<OFX>`, so we rewrite only the **header region** and leave the body alone:

- **1.x** → fill only the *absent* required keys, preserving the real `VERSION`/`ENCODING`/etc.
- **2.x** → the `<?xml?>`/`<?OFX?>` prolog has no `KEY:VALUE` headers, so we emit a synthetic 1.x header. Key discovery: **`ofxy`'s body deserializer already handles XML** — closed tags (`<TRNAMT>-50.00</TRNAMT>`), entity decoding (`&amp;` → `&`), and self-closing tags (`<MEMO/>`) all work — so the 2.x body parses unchanged once the header is SGML-shaped.
- Conformant 1.x input is returned borrowed (no change); input without `<OFX>` is left for `ofxy` to reject as before.

The injected values are placeholders `ofxy` reads only structurally (transactions come from the body), so this is lossless.

## Verify

```
# 1.x sparse headers       -> 3 txns (was: "headers missing 'CHARSET'")
# 2.x XML                  -> 1 txn  (was: "does not yet support OFX version: 200")
# 2.x XML w/ &amp; + <MEMO/> -> 1 txn, payee "Johnson & Co"
# full-header 1.x          -> 2 txns (unchanged)
```

Unit tests added for the 1.x-sparse, 2.x-XML, conformant-untouched, and no-`<OFX>` paths. `rustledger-importer` suite + clippy + fmt green.

## Residual risk / long-term

The 2.x path relies on `ofxy`'s body deserializer tolerating XML — an implementation detail, not a documented contract. This is bounded by the pinned `ofxy = "0.2"` and the new 2.x unit tests (which fail loudly if an `ofxy` bump regresses it). The durable direction remains a native OFX reader that drops `ofxy` (and its `chrono` seal, per this module's header comment); the shim makes that non-urgent.

### DIFF
--- a/crates/rustledger-importer/src/ofx_importer.rs
+++ b/crates/rustledger-importer/src/ofx_importer.rs
@@ -18,8 +18,84 @@ use anyhow::{Context, Result};
 use rustledger_core::NaiveDate;
 use rustledger_core::{Amount, Directive, Posting, Transaction};
 use rustledger_ops::enrichment::{CategorizationMethod, Enrichment};
+use std::borrow::Cow;
 use std::fs;
 use std::path::Path;
+
+/// The OFX 1.x SGML header keys `ofxy` requires, with safe canonical defaults.
+/// `ofxy` reads the header only structurally — we extract transactions from the
+/// body — so supplying placeholders for absent keys is lossless here.
+const OFX_HEADER_DEFAULTS: &[(&str, &str)] = &[
+    ("OFXHEADER", "100"),
+    ("DATA", "OFXSGML"),
+    ("VERSION", "102"),
+    ("SECURITY", "NONE"),
+    ("ENCODING", "USASCII"),
+    ("CHARSET", "1252"),
+    ("COMPRESSION", "NONE"),
+    ("OLDFILEUID", "NONE"),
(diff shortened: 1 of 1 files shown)


robcohen:
Superseded by #1459, which replaces `ofxy` with a native dependency-free OFX parser. The native parser fixes both halves of #1457 more robustly (no reliance on `ofxy`'s undocumented XML tolerance) and additionally removes the `chrono` dependency this module carried only for `ofxy`. Closing this shim in favor of that.

PR #1460, opened 2026-06-20
fix(cli): broken-pipe handling for check/completions + query flag-after-query hint

robcohen:
## What

Two CLI robustness papercuts found by dogfooding:

1. **Broken pipe on `check` / `completions`.** Piping their output to a reader that closes early (`| head`, or quitting a pager) produced `error: Broken pipe (os error 32)` (exit 2) for `check`, and a **panic (exit 101)** for `completions` — whereas `query`/`format`/`report` exit cleanly.
2. **`query` swallows flags placed after the query.** The `QUERY` positional is `trailing_var_arg` (to allow unquoted queries), so `rledger query f.bean "SELECT …" --numberify` captured `--numberify` as a query token and failed with a baffling `column 'numberify' not found`.

## How

- Add the existing `is_broken_pipe` guard to the **`check`** dispatch arm (matching query/format/report).
- **`completions`**: generate into a buffer, then write it ourselves and treat `BrokenPipe` as success — `clap_complete::generate` writes straight to the sink and panics on a write error.
- **`query`**: before running, detect a leading `--x` / `-x` (short flag) token in the query and bail with an actionable hint ("Flags must come before the query…"). Negative numbers (`-5`) are not flagged.

## Verify

```
rledger check big.beancount | head            # exit 0, no "Broken pipe"
rledger completions bash | head -c 30         # exit 0, no panic
rledger query f.bean "SELECT account" -m      # error: '-m' looks like a command-line flag …
rledger query f.bean "SELECT account WHERE number > -5"   # runs fine
```

Extended the broken-pipe integration test to cover `check` and `completions`; added a `query` flag-after-query test. clippy + fmt green.

### DIFF
--- a/crates/rustledger/src/bin/rledger.rs
+++ b/crates/rustledger/src/bin/rledger.rs
@@ -324,6 +324,9 @@ fn main() -> ExitCode {
             }
             match rustledger::cmd::check::run(&args) {
                 Ok(code) => code,
+                // A downstream reader (e.g. `| head`) closing the pipe is not an
+                // error — exit cleanly, matching the Query/Format arms.
+                Err(e) if rustledger::pager::is_broken_pipe(&e) => ExitCode::SUCCESS,
                 Err(e) => {
                     eprintln!("error: {e:#}");
                     ExitCode::from(2)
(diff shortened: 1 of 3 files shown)


robcohen:
Good catch — fixed. The short-flag check now requires exactly `-` plus one letter (e.g. `-m`), so BQL unary negation of a column (`SELECT -number`, `WHERE amount > -balance`) is no longer mistaken for a flag; long flags stay `--…`. Added a test asserting `-number` is not flagged.

Issue #877, opened 2026-04-21
fix: zero-value posting pruning hides validation errors


robcohen:
## Description

When an elided posting is interpolated to zero, rustledger prunes it from the transaction. This can hide validation errors — for example, if the elided posting references an account that was never opened, the E1001 error is suppressed because the posting is removed before validation runs.

## Reproducer

```beancount
option "operating_currency" "USD"
2024-01-01 open Assets:Bank USD

2024-01-15 * "Zero balance"
  Assets:Bank  0.00 USD
  Expenses:NeverOpened
```

Expected: E1001 error for `Expenses:NeverOpened` (account not opened).
Actual: Only E3004 warning (transaction has one posting) — the unopened account error is hidden.

## Root cause

Zero-value interpolated postings are pruned in `interpolate.rs` before validation runs. Validation should occur on the full set of postings (including zero-value ones), or at minimum the pruning step should validate accounts before removing postings.

## Upstream reference

Beancount issue beancount/beancount#962

PR #1461, opened 2026-06-20
fix(validate): defer explicit-posting open check past account-rewriting plugins

robcohen:
## What

Account-rewriting regular plugins (`rename_accounts`, `split_expenses`, …) run **after** the early validation phase, so rledger's open-existence check (`E1001`) fired on the **pre-rewrite** account name — an account that doesn't exist in the final ledger:

```
plugin "rename_accounts" "{'Expenses:Old': 'Expenses:New'}"
2020-01-01 open Assets:Cash
2020-01-01 open Expenses:New
2020-01-02 * "x"
  Assets:Cash  -10.00 USD
  Expenses:Old  10.00 USD     ; renamed to Expenses:New by the plugin
→ error: Account Expenses:Old was never opened   ❌
```

Found by dogfooding (pass #5).

## How

The pipeline is `sort → synth-plugins → Early → book → regular-plugins → Late → finalize`. The open-existence check for a posting to an unopened account was emitted entirely in **Early**, before regular plugins rewrite accounts.

- **Early** now emits E1001 only for **elided** postings to unopened accounts. This case *must* stay early: booking prunes zero-interpolated postings, which would otherwise hide the error (the documented Python #877 invariant — `interpolate.rs:532`).
- **Explicit** postings' open check is **deferred to Late**, which runs after the regular plugin pass, so it sees the final account names.
- A new `account_not_open_early` set on `LedgerState` dedups, so an elided posting that's *still* unopened after plugins isn't double-reported in Late.

## Scope

Fixes account-rewriting plugins for **explicit** postings (`rename_accounts`; `split_expenses` with explicit amounts). The **elided-shorthand** source case (`split_expenses` with an elided `Expenses:Accommodation` posting) stays tied to the early zero-pruning invariant and is left as a known limitation.

## Verify

- `rename_accounts` + explicit posting → loads clean (was E1001). ✓
- Genuine unopened account (no plugin), explicit or elided → still E1001. ✓
- `#877` elided-zero pruned posting → still caught early. ✓
- Full `rustledger-validate` (89+ tests) + `rustledger-loader` suites pass; new `test_open_check_deferred_past_account_rewriting_plugins`; clippy + fmt green.

### DIFF
--- a/crates/rustledger-loader/tests/loader_test.rs
+++ b/crates/rustledger-loader/tests/loader_test.rs
@@ -2134,7 +2134,7 @@ fn test_open_check_deferred_past_account_rewriting_plugins() {
     );
 }
 
-/// The deferred E1001 dedup is keyed per-posting (file_id, span), not per
+/// The deferred E1001 dedup is keyed per-posting (`file_id`, span), not per
 /// (account, date): two transactions on the SAME date posting to the SAME
 /// unopened account — one elided (reported early), one explicit (reported
 /// late) — must BOTH be flagged, not silently deduped to one.
--- a/crates/rustledger-loader/tests/loader_test.rs
+++ b/crates/rustledger-loader/tests/loader_test.rs
@@ -2088,3 +2088,48 @@ fn test_booking_failure_partitioned_from_late_validation() {
         "all three transactions should appear in final Ledger output (including the failed one); got {txn_count}"
     );
 }
+
+/// Account-rewriting regular plugins (e.g. `rename_accounts`, `split_expenses`)
+/// run AFTER the early validation phase, so the open-existence check (E1001)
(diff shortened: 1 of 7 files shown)


robcohen:
Good catch — fixed. The dedup set is now keyed by the posting's source identity `(file_id, span)` instead of `(account, date)`, so the late phase skips only the *same* posting (a booked-from-elided one), and a different same-day transaction posting to the same unopened account is still reported. Verified: `Assets:Cash -10 / Expenses:Forgot` (elided shorthand) → exactly 1 E1001; two same-day txns (one elided, one explicit) to the same unopened account → 2. It's also fail-safe — if a span ever didn't round-trip through booking, the worst case is a duplicate diagnostic, never a missing one. Added a loader regression test.

PR #1463, opened 2026-06-20
fix(ffi): preserve warning severity + robust include handling in load_source

robcohen:
## What

Follow-up to #1462 (the `load_source` pipeline collapse), addressing Copilot's review — its comments landed just after the auto-merge, so they're handled here.

1. **Warning severity dropped (most impactful).** `ledger_error_to_ffi` hardcoded `severity: "error"`. With `load_source` now routing through the full pipeline, plugin **warnings** — e.g. the `unrealized` plugin's gain/loss notices — flow through `ledger.errors` and were surfacing to FFI consumers as **errors**. Now maps `ErrorSeverity::Warning` → wire `"warning"`.

2. **Glob include paths errored.** The empty-VFS-stub approach keyed on the literal include path, which can't match a glob (`[ab].bean` ≠ literal file `[ab].bean`) → `GlobNoMatch` parse error, breaking the "list, don't resolve, don't error" contract. Replaced stubbing with a **targeted filter** that drops parse-phase resolution failures (file-not-found / glob-no-match) for *declared* include paths — uniform for literal and glob, and narrow enough never to mask a real syntax error.

3. **Fatal-fallback error mis-tagged.** The fatal load/process fallback built a default (parse-phase) `Error`; a fatal process/plugin failure would then suppress `ledger.validate`/`query`. Now tagged `validate`.

## Tests

`tests/load_regular_plugins.rs` (now 5 tests):
- `load_source_preserves_warning_severity` — `unrealized` notice surfaces as `warning`, not `error`.
- `load_source_glob_include_does_not_error` — unresolved glob include produces no error.

## Verify

All 8 `rustledger-ffi-wasi` suites + `rustledger-ffi-component` pass; `cargo clippy --all-features --all-targets` and `cargo fmt --check` clean.

### DIFF
--- a/crates/rustledger-ffi-wasi/src/helpers.rs
+++ b/crates/rustledger-ffi-wasi/src/helpers.rs
@@ -109,12 +109,6 @@ pub fn load_source(source: &str) -> LoadResult {
     // (booking errors surface; semantic validation is `ledger.validate`'s job).
     let mut vfs = rustledger_loader::VirtualFileSystem::new();
     vfs.add_file("<source>", source);
-    for inc in &includes {
-        // Empty stub: include resolution becomes a no-op, so a bare source with
-        // `include` directives lists them (above) without emitting a parse-phase
-        // "file not found" error that would suppress `ledger.validate`.
-        vfs.add_file(inc.path.as_str(), "");
-    }
     let load_opts = rustledger_loader::LoadOptions {
         validate: false,
         ..Default::default()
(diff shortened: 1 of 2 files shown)


robcohen:
Good catch — fixed. The filter now normalizes path separators to `/` on both the error message and the declared include paths before matching, so it works regardless of platform (the loader renders paths via `Path::display()`, which uses `\` on Windows).

PR #1464, opened 2026-06-21
fix(wasm): stop treating plugin warnings as fatal

robcohen:
## What

The `rustledger-wasm` library correctly routes source through the canonical `process()` pipeline, so plugin **warnings** (e.g. the `unrealized` plugin's gain/loss notices) flow through `load.errors` with `Warning` severity. But several entry points treated *any* non-empty error list as fatal — so a warning behaved like an error:

- **`run_validation` skipped validation entirely** on a non-empty `load.errors`. A ledger that emits a warning had its real diagnostics (E1001, balance residuals, …) **silently dropped**. *(the serious bug)*
- **`validateSource` / `validateMultiFile`** computed `valid: errors.is_empty()`, so a **warning-only ledger reported `valid: false`**.
- **`query` / `expandPads` / `runPlugin`** returned early on any non-empty `load.errors`, refusing results for a ledger that merely warns.

Confirmed with `unrealized` + a posting to an unopened account: `load.errors` held the warning, `run_validation` returned 0 errors, and the genuine `E1001` was missed.

## How

Add a `has_fatal(&[Error])` helper (true iff any `Severity::Error`) and gate all of the above on it instead of emptiness — matching `rledger check`, which exits 0 on warning-only input. Warnings are still surfaced (validation results include them); they no longer abort processing or invalidate the ledger.

## Tests

`helpers.rs` unit tests:
- `has_fatal_ignores_warnings`
- `run_validation_not_skipped_by_warning` — validation runs and reports `E1001` even when the ledger also emits a plugin warning.

## Verify

86 `rustledger-wasm` tests pass; `cargo clippy --all-features --all-targets` and `cargo fmt --check` clean.

### DIFF
--- a/crates/rustledger-wasm/src/api.rs
+++ b/crates/rustledger-wasm/src/api.rs
@@ -11,7 +11,7 @@ use rustledger_loader::{FileSystem, LoadError, LoadResult};
 use rustledger_parser::parse as parse_beancount;
 
 use crate::convert::{directive_to_json, value_to_cell};
-use crate::helpers::{extract_options, load_and_book, run_validation, to_js};
+use crate::helpers::{extract_options, has_fatal, load_and_book, run_validation, to_js};
 #[cfg(feature = "completions")]
 use crate::types::{CompletionJson, CompletionResultJson};
 use crate::types::{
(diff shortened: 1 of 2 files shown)


robcohen:
Good catch on all three — fixed. `query`, `expandPads`, and `runPlugin` now carry the non-fatal `load.errors` (warnings) into every result path: success, plus the secondary error paths (query parse/execution error, runPlugin unknown-plugin/conversion error). So warning-only ledgers return their results AND surface the warnings.

PR #1465, opened 2026-06-21
docs(check): document E7003 duplicate-option strictness as a deliberate deviation

robcohen:
## What

Surfaced by a cross-surface warning audit (pass 12): `rledger check` reports a repeated **non-repeatable** option — E7003, e.g. `option "title"` declared twice — as a **hard error** (exit 1), while beancount's `bean-check` silently accepts it (last value wins, exit 0).

Mapping rledger vs beancount across the option-warning codes:

| Code | Case | beancount | rledger |
|---|---|---|---|
| E7001 | unknown option | exit 1 | exit 1 ✓ |
| E7002 | invalid value | exit 1 | exit 1 ✓ |
| E7005 | read-only option | exit 0 | exit 0 ✓ |
| **E7003** | **duplicate non-repeatable** | **exit 0** | **exit 1** (stricter) |

The existing comment (*"In Python beancount, invalid options are errors, not warnings"*) is accurate for E7001/E7002 but was applied uniformly, sweeping in E7003.

## Decision

Keep the stricter behavior — a duplicated non-repeatable option is almost always a copy-paste mistake worth surfacing — but make it a **deliberate, documented deviation** per the CLAUDE.md Python-compatibility checklist (bucket 1), rather than an undocumented divergence.

## How

- Expand the `cmd::check` comment to distinguish E7001/E7002 (match beancount) from E7003 (deliberate deviation), with the rationale.
- Add a rationale note at the E7003 generation site in `options.rs`.
- Pin the behavior with CLI regression tests.

No behavior change.

## Tests

- `cli_commands_test::test_check_duplicate_option_is_error` — duplicate option fails check, output cites E7003.
- `cli_commands_test::test_check_single_option_ok` — a single non-repeatable option passes (no false positive).

## Verify

New CLI tests pass; `rustledger-loader` option tests (31) pass; `cargo clippy --all-features --all-targets` and `cargo fmt --check` clean.

### DIFF
--- a/crates/rustledger-loader/src/options.rs
+++ b/crates/rustledger-loader/src/options.rs
@@ -244,7 +244,12 @@ impl Options {
             return; // Don't apply the value
         }
 
-        // Check for duplicate non-repeatable options (E7003)
+        // Check for duplicate non-repeatable options (E7003).
+        //
+        // Stricter than beancount: `bean-check` silently lets the last value
+        // win (exit 0). We flag the duplicate so the CLI can surface it as an
+        // error (a repeated non-repeatable option is almost always a mistake) —
+        // a DELIBERATE deviation. See `cmd::check` E7003 handling.
         let is_repeatable = REPEATABLE_OPTIONS.contains(&key);
         if is_known && !is_repeatable && self.set_options.contains(key) {
             self.warnings.push(OptionWarning {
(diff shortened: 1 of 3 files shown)


robcohen:
Good catch — reworded. The comment now states that all option warnings (E7001-E7006 from Options::set) are surfaced as hard errors, while keeping the E7001/E7002 (match beancount) vs E7003 (deliberate deviation) distinction.

robcohen:
Already addressed before merge: the rebased branch reworded this comment to state that all option warnings (E7001-E7006 from `Options::set`) are surfaced as errors. See current `check.rs`.

PR #1466, opened 2026-06-21
fix(price): skip self-referential (base == quote) fetch jobs

robcohen:
## What

Found by dogfooding (pass 13): `rledger price --undeclared` emits a **nonsensical `price USD <n> USD`** directive — a currency priced in itself — which would pollute the user's price DB.

```
$ rledger price -f ledger.beancount --undeclared -b --source-cmd '…'
2024-06-01 price AAPL 150.00 USD
2024-06-01 price USD 150.00 USD   ← bogus: USD priced in USD
```

The `--undeclared` ticker-shape heuristic walks transactions and picks up the operating/quote currency (USD, held as cash or seen as a cost currency) as a **base** candidate. Resolving its quote yields the operating currency — itself — producing a self-referential `(USD, USD)` job. bean-price never fetches a currency against itself.

## How

Skip any job whose quote currency equals its base symbol. The price command has **three** independent job-iterating paths, so the guard is applied in each to keep them consistent:
- the network fetch loop,
- the `--dry-run` plan (`dump_fetch_plan`),
- the `--source-cmd` external path (`run_with_external_command`).

The per-job `base == quote` check (rather than excluding the currency entirely) preserves legitimate foreign-currency fetches: EUR held in a USD ledger still produces `(EUR, USD)`.

## Tests

`cli_commands_test::test_price_no_self_quote_directive` — a ledger holding AAPL in USD, via `--source-cmd`, fetches `price AAPL` but emits no `price USD … USD`.

## Verify

New test + 9 existing `price` tests pass; `cargo clippy --all-features --all-targets` and `cargo fmt --check` clean. Confirmed across `-n`, network, and `--source-cmd` paths; `(EUR, USD)` still fetched.

### DIFF
--- a/crates/rustledger/src/cmd/price_cmd.rs
+++ b/crates/rustledger/src/cmd/price_cmd.rs
@@ -357,6 +357,15 @@ pub fn run_with_writer<W: Write>(
             );
 
         for (effective_currency, per_spec_mapping) in per_quote_jobs {
+            // A commodity is never priced in itself. Skip self-referential
+            // (base == quote) jobs — these arise when `--undeclared` picks up
+            // the operating/quote currency (e.g. USD held as cash, or seen as a
+            // cost currency) as a base candidate, whose resolved quote is then
+            // itself. bean-price never fetches a currency against itself; left
+            // unfiltered this emits a nonsensical `price USD … USD` directive.
+            if &effective_currency == symbol {
+                continue;
+            }
             // --clobber: pre-fetch skip when an explicit `price` directive for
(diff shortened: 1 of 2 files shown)


robcohen:
Both addressed: the test now parses each emitted `<date> price <SYMBOL> <num> <CURRENCY>` directive and asserts `symbol != currency` per line (precise — no longer a substring match that would catch `price USDX`), skips only when stderr shows `--source-cmd` is unsupported (no longer masking real failures), and is `#[cfg(unix)]` since the `echo` stub source isn't a standalone executable on Windows.

PR #1467, opened 2026-06-21
fix(query): SQL three-valued NULL semantics in comparisons

robcohen:
## What

Found by a BQL differential pass against beanquery (pass 16): comparisons against a **NULL column value** (e.g. a transaction with no payee) diverged from beanquery two ways:

```
ledger: 1 txn WITH payee "Alpha", 1 txn WITHOUT a payee
                          rledger   beanquery
WHERE payee != "Alpha"      4 (!)      2     ← includes the no-payee rows
WHERE payee > "A"        EMPTY (!)     2     ← whole query lost, even "Alpha" > "A"
WHERE payee < "z"        EMPTY (!)     2
```

1. **`!=` includes NULLs.** `Ne` is `!values_equal`, and `values_equal(NULL, x)` is false, so `NULL != x` → true. A `WHERE payee != "x"` filter wrongly includes payee-less transactions.
2. **Ordered comparisons empty the query.** `compare_values` returns `Err("cannot compare values")` on a NULL operand; that error propagates and empties the **whole** result — so a single payee-less transaction makes `WHERE payee > "X"` return nothing, dropping valid non-NULL matches too.

## How

Per SQL three-valued logic (and beanquery), any comparison with NULL is UNKNOWN → a WHERE clause treats it as not-matched. Return `false` for every comparison operator (`=`, `!=`, `<`, `<=`, `>`, `>=`) when either operand is NULL, in both `evaluate_binary_op` and `binary_op_on_values`. `values_equal` is untouched, so **GROUP BY still groups NULL keys together** (verified).

## Tests

`bql_integration_test::test_null_comparison_excludes_rows` — `!=`/`=`/`>`/`<` against a NULL payee column.

## Verify

`payee != "x"` / `> "A"` / `< "z"` now match bean-query exactly; full `rustledger-query` suite passes (0 regressions, incl. GROUP-BY null grouping); `cargo clippy --all-features --all-targets` and `cargo fmt --check` clean.

### DIFF
--- a/crates/rustledger-query/src/executor/operators.rs
+++ b/crates/rustledger-query/src/executor/operators.rs
@@ -8,6 +8,20 @@ use crate::error::QueryError;
 use super::Executor;
 use super::types::{Interval, PostingContext, Value};
 
+/// Whether `op` is an equality or ordering comparison — the operators for which
+/// a NULL operand yields SQL "UNKNOWN" (treated as not-matched).
+const fn is_comparison(op: BinaryOperator) -> bool {
+    matches!(
+        op,
+        BinaryOperator::Eq
+            | BinaryOperator::Ne
+            | BinaryOperator::Lt
+            | BinaryOperator::Le
+            | BinaryOperator::Gt
+            | BinaryOperator::Ge
+    )
+}
+
 impl Executor<'_> {
     /// Evaluate a binary operation.
     pub(super) fn evaluate_binary_op(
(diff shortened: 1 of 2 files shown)


robcohen:
This is a false positive — `matches!(left, Value::Null)` matches a *fieldless* variant, which inspects rather than moves `left` (no binding is introduced), so `left`/`right` remain usable afterward. The code compiles and the full `rustledger-query` suite passes. No change needed.

robcohen:
Still a false positive (re-flagged after the rebase): `matches!(left, Value::Null)` matches a fieldless variant, which inspects without moving `left`. The branch builds and the full rustledger-query suite passes. No change needed.

PR #1468, opened 2026-06-21
fix(query): COUNT(column) counts only non-NULL values

robcohen:
## What

Found by the pass-18 NULL-edge BQL differential: `COUNT(column)` counted NULLs.

```
ledger: 1 txn WITH payee, 1 txn WITHOUT (4 postings total)
                  rledger   beanquery
COUNT(*)             6          6      ← all rows (correct)
COUNT(payee)         6 (!)      4      ← should skip the payee-less rows
```

`COUNT` always returned `group.len()`, ignoring whether the argument was `*` or a column. SQL/beanquery count only **non-NULL** values for `COUNT(column)`.

## How

Count only rows where the argument evaluates to non-NULL. `COUNT(*)` (no arg or `*`) still counts every row.

## Tests

`aggregate_count_null_test::count_column_excludes_nulls` — `COUNT(*)` = 4, `COUNT(payee)` = 2 on a ledger with a payee-less transaction. (Standalone test file to avoid an append-conflict with the concurrent #1467.)

## Verify

`COUNT(payee)` now matches bean-query; full `rustledger-query` suite passes (0 regressions); clippy + fmt clean.

### DIFF
--- a/crates/rustledger-query/src/executor/aggregation.rs
+++ b/crates/rustledger-query/src/executor/aggregation.rs
@@ -265,14 +265,29 @@ impl<'a> Executor<'a> {
             Expr::Function(func) => {
                 match func.name.to_uppercase().as_str() {
                     "COUNT" => {
-                        // COUNT(*) or COUNT(expr) — validate argument count
                         if func.args.len() > 1 {
                             return Err(QueryError::InvalidArguments(
                                 "COUNT".to_string(),
                                 "expected 0 or 1 argument".to_string(),
                             ));
                         }
-                        Ok(Value::Integer(group.len() as i64))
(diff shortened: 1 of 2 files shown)


robcohen:
Good catch on the second COUNT path. Investigating, I found rledger's subquery aggregation (`FROM (SELECT ...)`) currently returns NULL for the count (a deeper, separate issue from this NULL-count fix — `SELECT count(payee) FROM (SELECT payee)` yields NULL, not a number). Rather than ship a table-path COUNT change that can't be exercised by a passing test, I reverted it and scoped this PR to the verified posting-stream fix. The subquery-count gap is tracked as a separate finding.

PR #1532, opened 2026-06-23
ci(compat): block on BQL query regressions

robcohen:
## What

Makes the **Compatibility** check **block on BQL query regressions**. Today the compat job records results but never exits non-zero — which is how #1506 dropped BQL `JOURNAL` from 90/90 → 12/90 (BQL compat 93%→77%) and still merged.

## How

- `scripts/compat-bql-test.py` gains `--baseline <prev-results.jsonl>`. After the run it compares against the baseline and **exits non-zero if any `(file, query)` pair that *matched* in the baseline now fails** — a true→false regression. The CI annotation lists exactly what regressed (`REGRESSED: <query> | <file>`).
- `compat.yml` fetches main's `compat-bql-results.jsonl` from the `compatibility` branch and passes it as `--baseline`.

**Why per-pair, not percentage:** comparing the overall % is fragile when the corpus or the eligible-file set changes. Per-`(file, query)` only flags an actual result flip — a pair that drops out of the run (coverage change) or a brand-new failure (never passed before) is **not** counted. Verified with a standalone test (catches the flip; ignores coverage-drops and new failures).

## Safety / ordering

- On **PRs**: a regression fails the BQL step → red check → blocks merge. The baseline-commit step is already `if: github.event_name != 'pull_request'`, so PRs never write the baseline.
- On **main / nightly**: a regression fails the job *before* the later "Commit results to compatibility branch" step, so a regression is **not** blessed into the baseline (the next run still compares against the last-good baseline).
- First run after merge: main is at ~93% (post-JOURNAL-fix) vs a ≤93% baseline → only improvements (false→true), no regressions → gate passes and the baseline advances.

## Scope note

This gates **BQL query** regressions (the JOURNAL class). The file-level `regression_count` is also computed but is intentionally **not** gated here — synthetic test files are regenerated each run, so a hard file gate needs determinism handling first. Tracked as a follow-up.

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -154,6 +154,9 @@ jobs:
           mkdir -p .github/badges
           curl -sSf https://raw.githubusercontent.com/${{ github.repository }}/compatibility/.github/badges/compat-check-results.jsonl \
             -o .github/badges/previous-check-results.jsonl 2>/dev/null || echo "" > .github/badges/previous-check-results.jsonl
+          # BQL baseline (main's per-query results) for the regression gate below.
+          curl -sSf https://raw.githubusercontent.com/${{ github.repository }}/compatibility/.github/badges/compat-bql-results.jsonl \
+            -o .github/badges/previous-bql-results.jsonl 2>/dev/null || echo "" > .github/badges/previous-bql-results.jsonl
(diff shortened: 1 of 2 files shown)


robcohen:
Reopening from chore/ branch (ci/ prefix fails the branch-name check).

PR #1549, opened 2026-06-23
refactor(plugin): single-source the python file-vs-module classifier across crates

robcohen:
## What

Architecture-roadmap [M/BR4], follow-on to #1548. Finishes single-sourcing the Python plugin **file-vs-module classifier** by unifying the *fourth* copy — `rustledger_plugin::python::discover_module_source`, the one that actually performs the runtime routing (the point Copilot raised on #1548).

`discover_module_source` keyed file-vs-module on `.py || contains(MAIN_SEPARATOR)`, omitting the forward slash — so on Windows a ref like `plugins/foo` was classified inconsistently with the loader's (already-fixed) up-front rejection.

## How

- New `rustledger_plugin::python::is_python_plugin_file_ref(name)` — the single string-criterion (`.py` extension, or a path separator including `/`).
- `discover_module_source` calls it.
- The loader's `is_python_plugin_file` (added in #1548) now delegates its string check to it (keeping its own `exists()` test).

All four classification sites — loader up-front rejection, loader dispatch, and `discover_module_source` — now share **one** criterion and can no longer disagree about the same reference.

## Behavior

Linux-identical (`MAIN_SEPARATOR == '/'`, so the added `/` is a no-op); the only change is the Windows forward-slash case, which now agrees everywhere.

## Tests

`rustledger-plugin` + `rustledger-loader` suites (with `python-plugins`) pass; clippy `--all-features --all-targets -D warnings` + fmt clean.

### DIFF
--- a/crates/rustledger-loader/src/process.rs
+++ b/crates/rustledger-loader/src/process.rs
@@ -1644,20 +1644,15 @@ fn is_python_module_name(resolved: &std::path::Path, raw_name: &str) -> bool {
 }
 
 /// Classify a Python plugin reference as a FILE path (vs a dotted module name):
-/// a file when it resolves to an existing path, ends in `.py` (case-insensitive),
-/// or contains a path separator. Forward `/` counts even on Windows (where
-/// `MAIN_SEPARATOR` is `\`), so `plugins/foo.py` is never mistaken for a module.
-///
-/// Single source for the up-front #1432 rejection (via `is_python_module_name`)
-/// and the runtime file-vs-module dispatch, which previously used non-equivalent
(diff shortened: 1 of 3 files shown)


robcohen:
Addressed: moved the `Discover and read a Python plugin's source code.` summary back onto `discover_module_source` (it got orphaned above the new classifier when I inserted it), so `is_python_plugin_file_ref`'s doc now starts with its own summary. On the Windows forward-slash point — that is intentional and the doc states it: `/` counts as a separator even on Windows (where `MAIN_SEPARATOR` is `\`), so `plugins/foo.py` is never misread as a module on any platform.

PR #1550, opened 2026-06-23
fix(importer): unify CSV categorization on RulesEngine (honor regex_mappings on the plain path)

robcohen:
## What

Architecture-roadmap [M/BR4] — unify `CsvImporter` categorization on `RulesEngine` for both extraction paths. **Fixes a real CLI bug.**

The plain `extract`/`extract_string` path (what the CLI uses) categorized via a bespoke `match_mapping` matcher that checked exact `mappings` and the merchant dictionary but **never consulted `regex_mappings`** — so a row matched only by a regex rule landed at the *default* account instead of its mapped one. The enriched path used `RulesEngine` (which loads all three). The two were kept in sync only by an "exactly as RulesEngine does" comment, and silently diverged.

## How

- New `build_rules_engine(csv_config)` — the single engine builder (exact mappings + regex mappings + merchant dict), used by **both** paths.
- The plain path builds it once per file and `parse_row` categorizes via `engine.categorize(...)`.
- Deleted the bespoke `match_mapping` and `merchant_regexes()` (their logic is `RulesEngine`'s `load_from_mappings` (substring) + `load_merchant_dict`, which is exactly what the plain path duplicated).

`load_from_mappings` uses the same case-insensitive substring match as the old `match_mapping`, so exact-mapping and merchant-dict behavior is unchanged — the only behavior change is that `regex_mappings` now apply on the plain path too.

## Tests

The audit noted **no test covered `regex_mappings` on the plain path**. Added `test_regex_mappings_categorize_on_plain_path_and_match_enriched`: a `SBUX`-only-regex row now categorizes to `Expenses:Coffee` on the plain path and matches the enriched path. Full `rustledger-importer` suite passes (158 existing tests unchanged); clippy + fmt clean.

### DIFF
--- a/crates/rustledger-importer/src/csv_importer.rs
+++ b/crates/rustledger-importer/src/csv_importer.rs
@@ -29,27 +29,6 @@ use std::path::Path;
 #[derive(Debug, Default, Clone)]
 pub struct CsvImporter;
 
-/// Precompiled merchant-dictionary regexes, built once on first use. Patterns
-/// that fail to compile are skipped, mirroring
-/// [`rustledger_ops::categorize::RulesEngine::load_merchant_dict`]. Each entry
-/// pairs a case-insensitive, unanchored regex with its `&'static` account.
-fn merchant_regexes() -> &'static [(regex::Regex, &'static str)] {
-    use std::sync::LazyLock;
-    static REGEXES: LazyLock<Vec<(regex::Regex, &'static str)>> = LazyLock::new(|| {
-        rustledger_ops::merchants::MERCHANT_PATTERNS
-            .iter()
-            .filter_map(|entry| {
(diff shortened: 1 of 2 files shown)


robcohen:
All three addressed: (1) the contra account is now borrowed as `&str` (matched account or default) — no per-row `String` allocation; (2) added a private `extract_string_with_engine` so the enriched path builds the `RulesEngine` once and reuses it for both extraction and enrichment instead of twice; (3) the merchant-dict regexes are compiled once process-wide via a `MERCHANT_REGEXES` `LazyLock` and `Arc`-cloned per engine — restoring the cache the plain path had before unification, now shared by both paths. importer + ops suites pass; clippy + fmt clean.

PR #1551, opened 2026-06-23
fix(plugin,booking): reject multi-char wire flags; document apply() precondition

robcohen:
## What

Architecture-roadmap [S/3] — close two boundary failure-mode gaps where a precondition violation is currently silent.

**(1) Multi-char wire flag silently truncated.** `data_to_transaction` (`from_wrapper.rs`) converted a multi-character wire flag to its first char (`"txn"` -> `'t'`), even though the same module deliberately *rejects* span overflow and `PerUnitFromTotal` "rather than silently truncating/corrupting", and the empty-flag case already errors. Now a flag whose char-length != 1 is rejected with `ConversionError::InvalidFlag`, symmetric with the empty-flag rejection.

**(2) `BookingEngine::apply()` had an undocumented "must book first" precondition.** It discarded `inv.reduce(...)`'s `Result` and is `pub` with no documented precondition — a caller doing `apply(&original_txn)` on an *unbooked* transaction gets a silently over-sold inventory. Added a `# Precondition` doc and a `debug_assert!` that the reduction succeeded (the audit's minimal option — release behavior unchanged, debug builds catch a violating caller). Not the heavier `BookedTransaction` newtype.

## Tests

`test_multi_char_wire_flag_rejected_not_truncated`: `"txn"` -> `InvalidFlag`, single-char still converts. Full `rustledger-plugin` + `rustledger-booking` suites pass; clippy + fmt clean.

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -447,7 +447,18 @@ impl BookingEngine {
         })
     }
 
-    /// Apply a transaction to the inventories (update balances).
+    /// Apply a transaction's postings to the running inventories (update
+    /// balances).
+    ///
+    /// # Precondition
+    ///
+    /// The transaction MUST already be booked — postings filled with complete
+    /// units and resolved costs, as produced by [`Self::book_and_interpolate`]
+    /// or the free [`book`](crate::book) function. Applying an *unbooked*
+    /// transaction can silently over-sell an inventory: a reduction with no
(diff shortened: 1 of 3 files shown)


robcohen:
Good catch — corrected the comment: `reduce()` can also fail with `InsufficientUnits`/`AmbiguousMatch`/`CurrencyMismatch`, not only a missing lot. All of them indicate the transaction was not booked before `apply()`, so the `debug_assert` (which fires on any error) is unchanged.

PR #1552, opened 2026-06-23
refactor(ops,ffi): restore clamp source provenance via clamp_indexed (BR6 step 1)

robcohen:
## What

Architecture-roadmap [L/BR6] — **step 1** of making `rustledger_ops::clamp` the single clamp engine. This PR restores the source provenance that blocked the unification, without yet touching the JSON engine.

`SessionState::clamp` already routes through the typed `ops::clamp`, but mapped **every** output through `directive_to_json(d, 0, "<clamped>")` — so in-window transactions and carried-forward prices lost their real `filename`/`lineno`, all attributed to a synthetic `<clamped>:0`. That provenance loss is exactly why #1425 backed the stateless `entry.clamp` off the typed path and onto the 846-line JSON `clamp_entries`.

## How

- New `rustledger_ops::clamp::clamp_indexed` returns, per output directive, the index of the **input** it was passed through from (`Some(i)`), or `None` when synthesized (opening-balance / earnings summary). `clamp` becomes a thin wrapper (`clamp_indexed(..).map(|(d,_)| d)`), so its behavior is unchanged.
- `SessionState::clamp` now uses `clamp_indexed` and looks up the original `lines[i]`/`files[i]` for pass-through entries, falling back to `<clamped>` only for synthesized ones.

## Why this first

This is the foundation the audit calls for ("first restore source provenance in `ops::clamp`"). With pass-through provenance preserved on the typed path, the **follow-ups** become safe: route the stateless `convert::clamp` through `clamp_indexed` (matching in-window outputs back to their original WIT directive) and then delete the JSON `clamp_entries` once it has no consumer.

## Tests

`clamp_indexed_tracks_source_provenance`: in-range entry maps back to its input index; synthesized summary is `None`; `clamp == clamp_indexed` sans indices. Existing clamp tests + `rustledger-ops`/`rustledger-ffi-component` suites pass; clippy + fmt clean.

### DIFF
--- a/crates/rustledger-ffi-component/src/convert.rs
+++ b/crates/rustledger-ffi-component/src/convert.rs
@@ -1403,9 +1403,19 @@ impl SessionState {
         ) else {
             return self.entries();
         };
-        rustledger_ops::clamp::clamp(&self.directives, begin_date, end_date)
-            .iter()
-            .map(|d| directive(ffi::convert::directive_to_json(d, 0, "<clamped>")))
+        // Pass-through entries (in-window directives, carried-forward prices)
+        // keep their original filename/lineno via the source index; only the
+        // synthesized opening-balance / earnings summaries fall back to the
+        // `<clamped>` sentinel. (#1425 — restores provenance that the old
+        // `clamp` -> `directive_to_json(d, 0, "<clamped>")` mapping dropped.)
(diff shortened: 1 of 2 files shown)


robcohen:
Agreed — added the source index (`a.1`) as the final sort tiebreaker so distinct directives with an identical (date, type, Display) key get a deterministic total order regardless of `sort_by`'s instability.

Issue #1419, opened 2026-06-17
Deprecate & remove the wasip1 JSON-RPC FFI surface (Phase 5 of #1384)


robcohen:
Split out from #1384 (Phase 5) so that issue can close once rustfava defaults to the component.

## Context

The wasip2 Component Model migration (#1384) is done through Phase 4: the WIT contract (#1400), the production `rustledger-ffi-component` guest crate (all 20 exports, parity-tested), and the release artifact (`rustledger-ffi-component-<ver>.wasm`, shipping since v0.16.2, #1410) all exist. rustfava can consume the component via `RUSTFAVA_RUSTLEDGER_BACKEND=component` (rustfava #168/#161).

The plan is **dual-ship**: the wasip1 JSON-RPC module stays untouched so the current integration never breaks, and we retire it only after the component is the proven default.

## Preconditions (must be true before starting)

- [ ] rustfava defaults to the component backend (the env flag flipped from opt-in to default) — the finishing step of #1384 Phase 4.
- [ ] A deprecation window has elapsed with the component as default in real use, no regressions.
- [ ] Audit confirms no remaining consumer of `rustledger-ffi-wasi-<ver>.wasm` (the consumer audit in #1384 found rustfava is the only one — re-verify at removal time).

## Work

- [ ] Announce deprecation: mark the wasip1 JSON-RPC surface deprecated in `docs/guides/integration.md` / `docs/reference/architecture.md`, note the target removal version.
- [ ] Stop publishing the wasip1 module from `release-build.yml` (keep the component artifact).
- [ ] Remove the `jsonrpc` router from `rustledger-ffi-wasi`.
- [ ] Remove the hand-maintained DTO mirror layer (`types/output.rs` ⇄ `types/input.rs` ⇄ `convert.rs`) that the JSON-RPC surface required.
- [ ] Remove the round-trip wire-format tests that policed the JSON shape (their purpose is subsumed by the WIT contract; see #1395).
- [ ] Decide the fate of the `rustledger-ffi-wasi` crate: fully removed, or retained only for the non-deprecated wasip1 use cases called out as out-of-scope in #1384 (the untrusted-plugin sandbox and the CPython/Python-WASI runtime, which stay wasip1). Document the decision.

## Out of scope

- The untrusted WASM plugin/importer sandbox and the Python-WASI (CPython) runtime — both deliberately stay wasip1 per #1384; this ticket only removes the **JSON-RPC embedding surface**.

## References

[...]

PR #1553, opened 2026-06-23
refactor(ffi): route stateless convert::clamp through clamp_indexed (BR6 step 2)

robcohen:
## What

Architecture-roadmap [L/BR6] — **step 2** of unifying on the typed clamp engine. Stacked on #1552 (now merged).

The stateless `entry.clamp` (the free `convert::clamp`) still went through the 846-line JSON `clamp_entries` — the #1425 provenance workaround — while `SessionState::clamp` used the typed `ops::clamp`. So **identical input could yield different clamped ledgers** depending on which embedding entry point you called.

## How

Rewrites `convert::clamp` to:
1. Convert the WIT inputs to core directives once, remembering each one's index in the input vector.
2. Call `rustledger_ops::clamp::clamp_indexed` (added in #1552).
3. For each output, hand back the **original** WIT directive for pass-throughs (by source index — full fidelity + provenance), and build only the synthesized opening-balance / earnings summaries fresh with the `<clamped>` sentinel.

This is exactly what `SessionState::clamp` does, so the stateless and session clamp paths now produce the same result. Deletes the now-dead `wit_directive_location` and `synthesized_transaction_to_wit` helpers.

## Scope

`clamp_entries` itself is **not** deleted here — it still backs the `ffi-wasi` JSON-RPC `clamp` command. That deletion is **step 3**, gated on routing (or removing) that surface (#1419). After this PR, `ffi-component` no longer calls `clamp_entries` at all.

## Tests

No clamp tests asserted the old JSON-synthetic WIT form, so none needed updating. `rustledger-ffi-component` suite passes; clippy `--all-features --all-targets -D warnings` + fmt clean.

### DIFF
--- a/crates/rustledger-ffi-component/src/convert.rs
+++ b/crates/rustledger-ffi-component/src/convert.rs
@@ -1094,7 +1094,14 @@ fn loaded_directive_to_input(d: &wit::Directive) -> ffi::InputEntry {
 }
 
 /// `entry.clamp` — clamp loaded directives to `[begin, end)` via the typed
-/// `rustledger_ops::clamp`. Round-trips WIT -> core -> ops -> WIT.
+/// [`rustledger_ops::clamp::clamp_indexed`].
+///
+/// In-window entries and carried-forward prices are handed back as their
+/// **original** WIT directive (full fidelity + provenance) via the source index;
+/// only the synthesized opening-balance / earnings summaries are built fresh
+/// (with the `<clamped>` sentinel). This matches `SessionState::clamp` so the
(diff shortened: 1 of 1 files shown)


robcohen:
Addressed: pre-allocated `core`/`orig_index` to `entries.len()`. On the semantics change — that is intentional and is what the PR is for: the stateless `entry.clamp` now matches `SessionState::clamp` (both via `clamp_indexed`), where they previously diverged because the stateless path used the buggy JSON `clamp_entries`. The fn doc already calls this out; the conversion-failure skip is documented inline as the only option on a surface with no error channel.

PR #1554, opened 2026-06-23
refactor(ffi-wasi): route entry.clamp through ops::clamp; delete the 846-line JSON clamp engine (BR6 step 3)

robcohen:
## What

Architecture-roadmap [L/BR6] — **step 3 (final)** of unifying on the typed clamp engine. Completes the ladder (#1552 step 1, #1553 step 2).

`ffi-wasi`'s JSON-RPC `entry.clamp` handler was the last consumer of the 846-line `clamp_entries` — the serde_json clamp engine that re-parses cost/inventory from stringified JSON and carried the documented "silently dropped every cost" data-loss bug. This routes it through the typed `rustledger_ops::clamp` (the engine `ffi-component` already uses) and **deletes the JSON engine**.

## How

- `handle_clamp_entries`: convert each input JSON entry to a core directive (`InputEntry` → `input_entry_to_directive`), remembering its index, then `rustledger_ops::clamp::clamp_indexed`. Pass-through outputs are returned as their **original JSON entry** (full fidelity + provenance via the source index); only synthesized summaries are built fresh (`<summarization>` source). Same pattern as the ffi-component surface, so all clamp paths now agree.
- Adds `rustledger-ops` as an `ffi-wasi` dependency.
- Deletes `clamp_entries` and its exclusive helpers (`create_earnings_transaction`, `accumulate_posting_balance`, `parse_cost_and_create_position`, `create_summary_transaction`, the account classifiers) and their tests — `commands/clamp.rs` goes **846 → 182 lines**.
- Keeps `filter_entries` + `ClampResult`/`FilterResult` (still used by `entry.filter`).

## Result

`clamp_entries` no longer exists; `rustledger_ops::clamp` is the single clamp engine across the whole workspace. The data-loss bug is gone from every surface.

## Tests

`filter_entries` tests retained and pass; the deleted-engine tests are removed with it. `rustledger-ffi-wasi` suite passes; clippy `--all-features --all-targets -D warnings` + fmt clean.


robcohen:
Fixed — now rejects rather than skips: convert every entry up front and return `invalid_params` on a parse/conversion failure (the handler has a `Result` channel). A dropped pre-`begin` entry would have silently under-counted the opening balance, so failing loud is the right call here. With nothing skipped, `core[i]` aligns with `params.entries[i]` and `orig_index` is gone.

PR #1555, opened 2026-06-23
refactor(query): route posting cost-resolve through Position::from_posting (BR8 step 1)

robcohen:
## What

Architecture-roadmap [XL/BR8] — **step 1** of unifying the four copies of "iterate postings, resolve cost→Position, accumulate Inventory, derive columns" onto one stream. This first slice single-sources the **cost-resolve leaf**, the most-duplicated piece.

The idiom `if cost_spec.resolve(units, date) { Position::with_cost } else { Position::simple }` (also guarding the no-cost case) was hand-written at 5 Position-building sites across the query executor. The canonical version already exists as `rustledger_core::Position::from_posting` (extracted in the L/7 work) — these sites just hadn't been routed through it.

## How

Routes all 6 query-executor Position-building cost-resolves through `Position::from_posting(units, posting.cost.as_ref(), date)`:
- `executor/mod.rs` — `build_balances_with_filter` accumulator + the `resolve_position` closure (running `balance`/`account_balance`)
- `executor/system_tables.rs` — `#postings` running-balance `pos` + the `position` column
- `executor/evaluation.rs` — the `position` posting-function

`Position::from_posting` is byte-for-byte equivalent to each replaced block (`cs.resolve(...)` → `with_cost`, else `simple`; outer `None` → `simple`), so query results are unchanged. The cost-*field*-extraction at system_tables (which reads the resolved cost's number/currency/date/label) and the units×cost weight in mod.rs are different idioms and are left alone.

## Why

Removes the "running balance / account_balance need to match so lot details aren't dropped" convention-only invariant (its comment is replaced by the shared call). A future `CostNumber` variant or cost-resolve fix now lands in one place.

## Tests

Full `rustledger-query` suite passes (these are result-producing paths); clippy `--all-features --all-targets -D warnings` + fmt clean. Follow-ups: the source-location-columns leaf (fixes bug #4's unchecked `as i64`), then the unified posting iterator.

### DIFF
--- a/crates/rustledger-query/src/executor/evaluation.rs
+++ b/crates/rustledger-query/src/executor/evaluation.rs
@@ -238,16 +238,11 @@ impl Executor<'_> {
                 // Position includes both units and cost.
                 // Uses resolve() to handle both per-unit and total cost syntax.
                 if let Some(units) = posting.amount() {
-                    if let Some(cost_spec) = &posting.cost
-                        && let Some(cost) = cost_spec.resolve(units.number, ctx.transaction.date)
-                    {
-                        Ok(Value::Position(Box::new(Position::with_cost(
-                            units.clone(),
-                            cost,
-                        ))))
-                    } else {
-                        Ok(Value::Position(Box::new(Position::simple(units.clone()))))
-                    }
+                    Ok(Value::Position(Box::new(Position::from_posting(
+                        units,
+                        posting.cost.as_ref(),
+                        ctx.transaction.date,
+                    ))))
                 } else {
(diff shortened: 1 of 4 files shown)


robcohen:
Good catch — fixed rather than reworded: routed the JOURNAL running-balance/position cost-resolve in `execution.rs` through `Position::from_posting` too, so the claim now holds for every Position-building site in the executor. (The `AtMode::Cost` resolve nearby computes `units × cost` — a weight, not a Position — so it is left alone, same as the units×cost site in mod.rs.)

PR #1556, opened 2026-06-23
refactor(query): single-source source-location columns; fix unchecked lineno cast (BR8 step 2)

robcohen:
## What

Architecture-roadmap [XL/BR8] — **step 2** of the PostingContext-stream unification. Single-sources the **source-location column** leaf (`filename`/`lineno`/`location`), and fixes **bug #4** along the way.

These three synthetic columns were built in ~3 places, two of which used an **unchecked `loc.lineno as i64`** cast (`lineno` is `usize`) — bug #4: on a hypothetical line number exceeding `i64`, the cast wraps to a negative `lineno`. The META path already guarded this (`source_location_meta_key` uses `try_from`); the *column* paths did not.

## How

Adds three shared helpers next to the existing `source_location_meta_key`:
- `source_filename_value`, `source_lineno_value`, `source_location_value` — each `Option<&SourceLocation> -> Value`.
- `source_lineno_value` is the single overflow-checked conversion (`i64::try_from(...).unwrap_or(i64::MAX)`), replacing every `as i64` cast.

Routes all column sites through them:
- `executor/evaluation.rs` — the `filename`/`lineno`/`location` posting-function arms
- `executor/system_tables.rs` — the standalone `filename`/`lineno` pair and the `(filename, lineno, location)` `#postings` tuple

Behavior is unchanged for every real ledger (linenos fit in `i64`); only the pathological overflow now saturates instead of wrapping.

## Tests

Added unit tests: `source_lineno_value` saturates `usize::MAX` to `i64::MAX` on 64-bit (bug #4 pin), basic lineno/filename/location values, and `None` handling. Full `rustledger-query` suite passes; clippy `--all-features --all-targets -D warnings` + fmt clean.

Follow-up (step 3): fold these leaves + the cost-resolve (step 1, #1555) into one posting-source iterator consumed by `build_postings_table`, `build_balances_with_filter`, and aggregation.

### DIFF
--- a/crates/rustledger-query/src/executor/evaluation.rs
+++ b/crates/rustledger-query/src/executor/evaluation.rs
@@ -381,17 +381,15 @@ impl Executor<'_> {
             )))),
             // Source location columns — resolved from the posting's own span
             // (falling back to the enclosing directive's location).
-            "filename" => Ok(self
-                .resolved_source_location(ctx)
-                .map_or(Value::Null, |loc| Value::String(loc.filename))),
-            "lineno" => Ok(self
-                .resolved_source_location(ctx)
-                .map_or(Value::Null, |loc| Value::Integer(loc.lineno as i64))),
-            "location" => Ok(self
-                .resolved_source_location(ctx)
-                .map_or(Value::Null, |loc| {
-                    Value::String(format!("{}:{}", loc.filename, loc.lineno))
-                })),
+            "filename" => Ok(Self::source_filename_value(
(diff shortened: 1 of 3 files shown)


robcohen:
Good catch — fixed: extracted `lineno_i64` (the saturating `try_from`) and both `source_lineno_value` and `source_location_value` now use it, so the `location` and `lineno` columns can never disagree on the same posting. Added an assertion to the overflow test pinning that consistency.

PR #1558, opened 2026-06-23
fix(query): reconcile #postings vs default-SELECT type + cost_date with the oracle (BR8 step 3)

robcohen:
## What

Architecture-roadmap [XL/BR8] — **step 3: find AND reconcile the posting-source divergences.** Using the parity harness, I mapped exactly where the default `SELECT` path (`collect_postings` + per-column evaluation) and the `#postings` table (`build_postings_table`) disagree, checked each against the **bean-query oracle**, and fixed the two that were wrong.

## Finding → fix

Per-column comparison across all 26 projectable columns: **24 agreed; 2 diverged.** Checked against bean-query (`bean-query <ledger> "SELECT type, cost_date"`):

| Column | Default (was) | `#postings` | bean-query oracle | Fix |
|--------|---------------|-------------|-------------------|-----|
| **`type`** | `"Transaction"` | `"transaction"` | **`transaction`** | default path → lowercase |
| **`cost_date`** | raw `CostSpec.date` (`None`) | resolved txn date | **`2024-01-14`** (resolved) | default path → resolve the cost |

In both cases the **default path was wrong** and `#postings` already matched the oracle. The fixes are in `executor/evaluation.rs`:
- `type` → `"transaction"`.
- `cost_date` → resolve the cost (`cs.resolve(units, txn.date)`) so an undated `{150 USD}` cost reports the booked txn date, exactly as `#postings` and beancount do.

The full `rustledger-query` suite passes unchanged — no test depended on the old (wrong) values.

## Net

The two posting-source paths now produce **identical values on every column** (sorted input). The parity test is upgraded to assert full per-column agreement, so it's the regression net proving the upcoming iterator unification behavior-preserving. The one remaining step-3 item is the **iteration order** on *unsorted* input (`build_postings_table` sorts by date; `collect_postings` doesn't) — moot in practice (loaded ledgers are pre-sorted) and resolved when the two iterations actually merge onto one stream.

## Tests

Parity harness (full-column agreement, per-column drift detector) + full `rustledger-query` suite pass; clippy + fmt clean.

### DIFF
--- a/crates/rustledger-query/src/executor/evaluation.rs
+++ b/crates/rustledger-query/src/executor/evaluation.rs
@@ -329,11 +329,19 @@ impl Executor<'_> {
                 .as_ref()
                 .and_then(|c| c.currency.as_ref())
                 .map_or(Value::Null, |c| Value::String(c.to_string()))),
-            // Cost date
+            // Cost date — the *resolved* cost date, not the raw spec date: an
+            // undated cost (`{150 USD}`) inherits the transaction date when
+            // booked, which is what bean-query and the `#postings` table report.
+            // Reading the raw `CostSpec.date` (often `None`) was the divergence.
             "cost_date" => Ok(posting
-                .cost
-                .as_ref()
-                .and_then(|c| c.date)
+                .amount()
+                .and_then(|units| {
+                    posting
+                        .cost
+                        .as_ref()
+                        .and_then(|cs| cs.resolve(units.number, ctx.transaction.date))
+                })
+                .and_then(|cost| cost.date)
                 .map_or(Value::Null, Value::Date)),
             // Cost label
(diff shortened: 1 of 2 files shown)


robcohen:
Both of these review the earlier recon-only commit and are superseded by the reconciliation commit (32264e2f): once `type` and `cost_date` were fixed to match the bean-query oracle, there is no divergence left, so I removed `AGREEING_COLUMNS`, `KNOWN_DIVERGENT_COLUMNS`, and the hardcoded `type, cost_date`. The current code has a single `ALL_COLUMNS` (which includes `type`/`cost_date`) and `columns_that_diverge` iterates exactly that — so there is no subset/hardcoded list to drift out of sync anymore.

Issue #1080, opened 2026-05-11
WASM hangs when reading accounts from Assets


alensiljak:
This is a heads-up after #1063 is resolved (I've noticed this earlier but thought it was just a glitch). The version that shows correct exchange rates completely hangs in some pages.
I will identify and post the queries and cases that seem to cause issues.
Some screens in Cashier just hang after the WASM has been rebuilt and the data reloaded.

- /accounts/tree
- /reports/assets

The main item is `/accounts/tree` so I'll focus on that. Will add more details as I identify them.

alensiljak:
So far seems to only happen on the desktop.

alensiljak:
The query that hangs is

```sql
SELECT account, value(sum(position), 'EUR') as value WHERE account ~ "^Assets" GROUP BY account ORDER BY account
```

also with one row

```sql
SELECT account, value(sum(position), 'EUR') as value WHERE account ~ "^Assets" GROUP BY account LIMIT 1
```

The same query takes several seconds in CLI but completes.

```
beanquery> SELECT account, value(sum(position), 'EUR') as value WHERE account ~ "^Assets" GROUP BY account LIMIT 1
account                              value
-----------------------------------  --------
Assets:Accounts-Receivable:BAM  0.00 EUR

1 row(s)
```

alensiljak:
Also

```sql
SELECT account, sum(position) as position WHERE account ~ "^Assets" GROUP BY account ORDER BY account
```

```sql
SELECT account, count(*) WHERE account ~ "^Assets" GROUP BY account
```

Even a simple

```sql
SELECT account WHERE account ~ "^Assets" LIMIT 10
```

alensiljak:
The query

```sql
SELECT account WHERE account ~ "^Assets"
```

takes multiple seconds in the CLI.
There are 22134 row(s) in the result but the time still somehow seems a bit more excessive than in the previous versions.

alensiljak:
Interesting and quite puzzling. Once this last query has completed, all queries no longer hung when executed and the performance is very similar the runs before the latest WASM build.
Very interesting thing to observe.
Now all the reports run within a couple of seconds again. (Vivaldi browser, Chromium-based)

alensiljak:
After reloading the page, the same issue repeats.

alensiljak:
This might be worth reopening and checking. The same behaviour has been reproduced.
The investments query

```sql
[...]

PR #1559, opened 2026-06-23
refactor(query): build_postings_table consumes the shared scan_postings stream (BR8 step 3 final)

robcohen:
## What

Architecture-roadmap [XL/BR8] — **step 3, final: the iterator merge.** The two posting-source iterations are now one. `build_postings_table` (the `#postings` table) consumes the same `scan_postings` stream the default `SELECT` path uses, instead of re-implementing the iterate-postings + accumulate-balances loop.

This is the structural dedup the audit's step 3 calls for — and it eliminates the duplicated balance-accumulation loop that was the source of bug #2 and the `type`/`cost_date` drift, replacing "two loops that must be kept in sync (and a test that catches it when they aren't)" with "one loop."

## How

1. **Extracted `scan_postings(from, where, needs_balance, needs_account_balance, where_reads_balance)`** from `collect_postings` — the single posting scan: iterate resolved directives, apply optional `FROM`/`WHERE`, accumulate cumulative `balance` + per-account `account_balance`, yield one `PostingContext` per posting. `collect_postings` is now a thin wrapper that computes the #1080 gates from the query and calls it.
2. **`build_postings_table` now calls `scan_postings(None, None, true, true, false)`** — every posting, no filter, both balances tracked — and reads `balance`/`account_balance` straight from each `PostingContext`. Its own `account_balances`/`cumulative_balance` accumulators are gone.
3. **Removed `build_postings_table`'s redundant `sort_by_key(date)`** — the loader already sorts directives, and the other two scanners (`collect_postings`, `build_balances_with_filter`) rely on that pre-sorted order without re-sorting. This also makes the three consumers consistent.

## Safety

- **No perf regression**: `build_postings_table` already paid the per-posting Inventory clone (the table materializes `balance`/`account_balance`); `collect_postings`'s #1080 gating is preserved (passed as explicit params).
- **Identical output**: the `#postings` parity test (every projectable column, including the accumulated balances) passes against the rewritten table, and the full `rustledger-query` suite is unchanged — `scan_postings` with no filter produces exactly the values the old inline loop did.

## Tests

`#postings` parity harness + full `rustledger-query` suite pass; clippy `--all-features --all-targets -D warnings` + fmt clean.

### DIFF
--- a/crates/rustledger-query/src/executor/mod.rs
+++ b/crates/rustledger-query/src/executor/mod.rs
@@ -474,9 +474,10 @@ impl<'a> Executor<'a> {
         where_reads_balance: bool,
     ) -> Result<Vec<PostingContext<'a>>, QueryError> {
         let mut postings = Vec::new();
-        // Per-account running balance — accumulates every posting regardless of
-        // FROM/WHERE filters, so `account_balance` always reflects the account's
-        // true ledger balance at the point of the posting.
+        // Per-account running balance — accumulates every posting the FROM clause
+        // keeps (plus the pre-`open_on` carry-in below), independent of the WHERE
+        // filter, so `account_balance` always reflects the account's true ledger
+        // balance at the point of the posting.
         let mut account_balances: FxHashMap<rustledger_core::Account, Inventory> =
             FxHashMap::default();
         // Single cumulative running balance across WHERE-filtered postings in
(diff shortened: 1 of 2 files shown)


robcohen:
All three addressed: (1) corrected the comment — `account_balance` is independent of the **WHERE** filter, not FROM (the FROM clause is honored; only the pre-`open_on` carry-in is added past it). (2) `build_postings_table`'s scan call is now `.expect(...)` — with no FROM/WHERE there are no predicates to evaluate, so failure is a real invariant violation, not something to swallow into an empty table. (3) `directive_index` is now `.expect(...)` since `scan_postings` always records it.

PR #1560, opened 2026-06-23
refactor(core): inventory unify (1+2) — single-source reduction gate + pad sub-account sum (BR6)

robcohen:
## What

Architecture-roadmap [L/BR6] — **inventory/lot-matching unify, parts 1 & 2.** Two duplications between the booking engine and the Late validator collapsed onto shared core sources.

### Part 1 — the reduction gate (`Inventory::is_booking_reduction`)

`BookingEngine::apply` and the validator's `update_inventories` made a **byte-identical** reduction decision in two crates:

```rust
method != BookingMethod::None && posting.cost.is_some()
    && inv.is_reduced_by(units, ReductionScope::CostBearingOnly)
```

This is the #1182 NONE-method gate — written twice, so the #1182 fix had to be applied twice and the two could drift. Now one core method (`Inventory::is_booking_reduction`) that both share. Only the identical 3-condition gate is unified; `book.rs` has two *other* `is_reduced_by` sites with deliberately different guards (one omits `cost.is_some()`), left untouched.

### Part 2 — the pad sub-account difference (a real bug)

A balance assertion sums an account **and its sub-accounts** (beancount semantic). The Late validator did this (`sum_account_and_subaccounts`); the booking **pad engine did not** — it measured `current` from the leaf account only (`inv.units()`). So a `pad` targeting a **non-leaf** account inserted the wrong synthetic amount, which then tripped the (sub-account-summing) validator.

**Oracle-verified** against `bean-check`/`bean-query`:
- `balance Assets:Bank 50 USD` with the 50 held in `Assets:Bank:Checking` → **passes** (sums sub-accounts); asserting `0 USD` fails with "accumulated 50 USD".
- `pad Assets:Bank` before `balance Assets:Bank 100 USD` (50 in the sub-account) → beancount pads **50**, and `Assets:Bank` totals **100**. Booking's leaf-only path padded 100 → the validator then saw 150 → spurious failure.

Fix: one `rustledger_core::sum_account_and_subaccounts` (built on the existing `is_subaccount_or_equal`), used by **both** the booking pad engine and the validator, so the pad amount and the validated amount are computed identically. New regression test (`test_process_pads_sums_subaccounts_for_nonleaf_target`) pins the oracle-confirmed value (50) — it fails on the old leaf-only code.

## Remaining

Part 3 of this audit item — have Late validation **consume** the booked inventory map instead of rebuilding it (the larger data-flow change) — follows as its own PR.

## Tests

New pad regression test + full `rustledger-core`/`-booking`/`-validate` suites pass; clippy + fmt clean.

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -467,18 +467,11 @@ impl BookingEngine {
                 let method = self.method_for(&posting.account);
                 let inv = self.inventories.entry(posting.account.clone()).or_default();
 
-                // Determine if this is a reduction: units reduce inventory when
-                // signs differ for the same currency. Only cost-bearing positions
-                // are considered, so simple (no-cost) positions don't trigger
-                // false reduction detection.
-                //
-                // NONE booking (issue #1182) treats every posting as an
-                // augmentation. Mixed-sign positions accumulate in the
-                // inventory — Python beancount's semantic for the NONE
-                // method, which the parallel guard in `book()` mirrors.
-                let is_reduction = method != BookingMethod::None
-                    && posting.cost.is_some()
-                    && inv.is_reduced_by(units, ReductionScope::CostBearingOnly);
+                // Reduction vs augmentation — the single source for this decision
+                // (`Inventory::is_booking_reduction`), shared with the Late
+                // validator so the two can't drift (including the #1182 NONE gate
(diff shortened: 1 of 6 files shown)


robcohen:
Clarified the doc: the gate keys on cost-spec *presence* (`cost.is_some()`, which includes an empty/unresolved spec like `{}`), now stated explicitly.

PR #1561, opened 2026-06-24
perf(validate): sub-account prefix index for balance assertions (M/2)

robcohen:
## What

Architecture-roadmap [M/2] — **precompute a sub-account prefix index for balance assertions, eliminating the O(assertions × accounts) scan.**

`sum_account_and_subaccounts` scanned **every** entry of the inventories map per balance assertion (`is_subaccount_or_equal` against each) — O(A·B) for A accounts and B assertions, on a file annotated as a hot path. A ledger with many accounts and many parent assertions paid a quadratic cost.

## Change

`LedgerState` gains `inventory_accounts: BTreeSet<Account>` — a lexically-sorted view of the inventory keys, kept in lockstep with `inventories` (both gain a key only in `register_open`/`register_open_late`; keys are never removed). `Account`'s `Ord` and `Borrow<str>` are lexical (via `Arc<str>`), so the subtree of `Assets:Bank` is exactly:

- `Assets:Bank` itself, **plus**
- the half-open range `["Assets:Bank:", "Assets:Bank;")` (`;` is the byte after `:`)

which captures every `Assets:Bank:*` and **nothing** else (e.g. excludes the `Assets:BankAlias` prefix sibling) — provably equivalent to `is_subaccount_or_equal`, answered by a `BTreeSet` range query in **O(log A + subtree)** instead of O(A) per assertion.

Implemented as a free fn `sum_account_subtree(inventories, index, account, currency)` taking the two fields directly (not `&self`) so call site 1 can keep its disjoint `&mut pending_pads` borrow. Both validator call sites use it. The unindexed `rustledger_core::sum_account_and_subaccounts` stays for the booking pad engine (its own map, far fewer calls).

## Correctness

- New parity test: the indexed `sum_account_subtree` equals the unindexed core scan for parent / leaf / prefix-sibling / missing targets, and `Assets:Bank` sums its subtree (10+40+5=55) while excluding `Assets:BankAlias` (99).
- The full `rustledger-validate` suite is unchanged — the index is behavior-equivalent, only faster.

## Benchmark

New `validate_subaccount_balance_scaling` scales the **accounts × assertions** product (32/64/128 accounts × 50 parent assertions) that the existing balance benchmark fixed at 2×1 — the dimension a quadratic regression would surface in, now covered so a future crossover is caught.

## Tests

Parity test + full validate suite pass; new benchmark compiles; clippy `--all-targets -D warnings` + fmt clean.

### DIFF
--- a/crates/rustledger-validate/benches/validate_bench.rs
+++ b/crates/rustledger-validate/benches/validate_bench.rs
@@ -208,10 +208,74 @@ fn bench_validate_balance_assertions(c: &mut Criterion) {
     group.finish();
 }
 
+/// Scale the (accounts × balance-assertions) product that
+/// `bench_validate_balance_assertions` fixes at 2×1. A parent balance assertion
+/// sums the whole subtree, so this is the dimension a quadratic
+/// `sum_account_and_subaccounts` regression would surface in — and the one the
+/// prefix index removes. Each `accounts` value runs `assertions` parent
+/// assertions, each summing every sub-account.
+fn bench_validate_subaccount_balance_scaling(c: &mut Criterion) {
+    let mut group = c.benchmark_group("validate_subaccount_balance_scaling");
+
+    let assertions = 50u32;
+    for accounts in [32u32, 64, 128] {
+        let mut directives = Vec::new();
+        directives.push(Directive::Open(Open::new(date(2024, 1, 1), "Assets:Bank")));
+        directives.push(Directive::Open(Open::new(
+            date(2024, 1, 1),
+            "Income:Salary",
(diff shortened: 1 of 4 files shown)


robcohen:
All three addressed: (1) the field doc now names `validate_open`/`register_open_late` and the `sum_account_subtree` free fn (the `&self` method was dropped to allow the disjoint `pending_pads` borrow); (2) same fix in the test comment; (3) the `[A:, A;)` bound strings are now built with `with_capacity` + `push_str` instead of `format!`, since this runs per balance assertion.

PR #1563, opened 2026-06-24
test(compat): runtime stale-check for the deliberate-divergence registries (S/2)

robcohen:
## What

Architecture-roadmap [S/2] — **mechanically stale-check the deliberate-divergence registries.**

`KNOWN_PYTHON_DIVERGENCES` / `KNOWN_RUST_DIVERGENCES` (in `scripts/compat-bql-test.py`) are hand-maintained `(filename, query_name)` allowlists with **no runtime check that the masked pair still actually diverges**. A registry entry can outlive the code change it documents (stale mask) and then **silently absorb a future genuine regression** that lands on the same pair — exactly the failure mode the registry's own rationale warns about. The codebase already has the right defense (`_is_beanquery_empty_aggregate_quirk` verifies a divergence fingerprint at runtime), but applied it only to the one empty-aggregate quirk, not to the static lists.

## Change

`stale_divergence_entries(results)` classifies every registry entry against the actual run results:
- **STALE** — the masked pair (or, for `(file, "*")`, *every* query on the file) now **matches** bean-query → the divergence is gone. **Hard-fails the run** (`::error::`, exit 1) so the entry is pruned.
- **DANGLING** — no run exercises the entry → the fixture/query was renamed or removed. **Warns** (`::warning::`), not fail, since a partially-fetched corpus can transiently drop a file.

This extends the runtime divergence-fingerprint discipline (`_is_beanquery_empty_aggregate_quirk`) to the static lists, so a mask can't outlive the divergence it documents.

The gate runs inside the existing BQL compat step (no extra heavy run). The detection logic is **pure** (`_stale_registry_entries` takes the registry + run indices) and unit-tested via a new `--self-test` (synthetic runs + registries, no corpus or tools needed), wired as a fast CI step before the corpus run so the logic is guarded even if fixtures fail to fetch.

## Tests

`python3 scripts/compat-bql-test.py --self-test` passes (asserts: a now-matching pair → stale; a still-diverging pair → not flagged; a never-run pair → dangling; a wildcard file is stale only when *all* its queries match). `py_compile` clean. The live stale-mask gate runs on this PR's `compatibility` job (pull_request trigger) — if any *current* registry entry is already stale, that job will surface it (the feature working).


robcohen:
Both addressed: (1) `_stale_registry_entries` now filters to *conclusive* runs (both tools succeeded) before deciding staleness — a tool timeout/non-zero exit (`match=False` without a real divergence) can no longer keep a stale mask alive, and a pair whose only runs failed is left in place as inconclusive (with a new self-test case). (2) `--self-test` uses explicit `check()` calls that raise instead of `assert`, so it still fires under `python -O`/PYTHONOPTIMIZE.

PR #1564, opened 2026-06-24
fix(loader): close fail-open hole in plugin path-security (BR4)

robcohen:
## What

Architecture-roadmap [M/4] (plugin-dispatch) — **fix the fail-open hole in the plugin path-security check** and single-source the traversal logic into one helper.

The `run_plugins` god-function had a third inline copy of plugin path-security (`process.rs:1062`) that was **bypassed on a `canonicalize` error**:

```rust
if options.path_security
    && let (Ok(canon_base), Ok(canon_plugin)) =
        (base_dir.canonicalize(), resolved.canonicalize())   // <- Err short-circuits
    && !canon_plugin.starts_with(&canon_base)
{ return Err(...); }
```

A plugin path that **can't be canonicalized** (doesn't exist yet, a typo, a permission error) makes the `let (Ok, Ok)` pattern fail, so the whole check is **skipped** and the path is allowed — a fail-OPEN path-traversal hole. Unlike the `include` traversal guard (which normalizes via the injected filesystem and never skips), this copy fails open.

## Fix

Extract two helpers:
- **`lexically_normalize`** — resolves `.`/`..` *without* touching the filesystem (preserving the root, clamping `..` at root), so a traversal is collapsed even for a non-existent path (`/ledger/../../etc` → `/etc`).
- **`plugin_path_within_base`** — `canonicalize`s both sides when the plugin file exists (symlink-safe), and otherwise falls back to `lexically_normalize` in the **same namespace**. It **fails CLOSED**: an unverifiable path is rejected, not allowed.

The inline check becomes `if options.path_security && !plugin_path_within_base(&resolved, base_dir)`.

## Scope note

The audit also flagged a "triplicated, non-equivalent file-vs-module classifier" — two of those three are **already unified** onto `rustledger_plugin::python::is_python_plugin_file_ref` by prior work (#1548/#1549 + follow-ups), so this PR targets the remaining concrete defect: the path-security copy. The larger `run_plugins` god-function extraction (moving native/wasm/python dispatch into `rustledger-plugin` behind one entry point) remains as a separate, bigger refactor.

## Tests

New `plugin_path_security_tests`: `..`-resolution for non-existent paths, traversal rejected even when `canonicalize` fails (the bug), in-bounds allowed, and a string-prefix sibling (`/ledger-evil`) correctly rejected. Full `rustledger-loader --features plugins` suite unchanged; clippy `--all-features --all-targets -D warnings` + fmt clean.

### DIFF
--- a/crates/rustledger-loader/src/process.rs
+++ b/crates/rustledger-loader/src/process.rs
@@ -869,6 +869,49 @@ struct PluginInvocation {
     force_python: bool,
 }
 
+/// Lexically resolve `.` / `..` in `p` WITHOUT touching the filesystem,
+/// preserving the root/prefix. Unlike `Path::canonicalize` this works on paths
+/// that don't exist yet, so a `..` traversal is still collapsed (e.g.
+/// `/ledger/../../etc` → `/etc`). A `..` at the root is clamped (it can't escape
+/// above the root).
+#[cfg(feature = "plugins")]
+fn lexically_normalize(p: &Path) -> std::path::PathBuf {
+    let mut out = std::path::PathBuf::new();
+    for component in p.components() {
+        match component {
+            std::path::Component::ParentDir => {
+                out.pop();
+            }
+            std::path::Component::CurDir => {}
+            other => out.push(other.as_os_str()),
+        }
+    }
+    out
+}
+
+/// Whether a `resolved` plugin path stays within `base_dir` under path-security.
+///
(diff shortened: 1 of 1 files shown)


robcohen:
Both addressed: the lexical fallback is now restricted to `ErrorKind::NotFound` (a not-yet-created path) — any other `canonicalize` error (permission denied, I/O) is genuinely unverifiable and is now REJECTED (fail closed), not lexically guessed. Doc updated to state the contract precisely.

PR #1565, opened 2026-06-24
refactor(dedup): single-source fuzzy dedup on core types + fix threshold drift (BR4)

robcohen:
## What

Architecture-roadmap [L/4] (ops-dedup) — **delete the duplicate CLI fuzzy-dedup; make `ops::dedup`'s fuzzy path core-typed; fix the threshold-boundary drift.**

Fuzzy duplicate detection was implemented **twice**, with a real bug:
- `rustledger_ops::dedup::find_fuzzy_duplicates` — maintained and tested, but on the wire type `DirectiveWrapper` and with **zero production callers**.
- `rledger extract`'s own `extract_cmd/duplicate.rs` — core-typed and **the live path**, but untested-by-ops and using a divergent threshold.

The two disagreed at exactly 50% word overlap: ops returns duplicate (`ratio >= 0.5`), the CLI copy did **not** (`matches*2 > shorter.len()`, i.e. `> 0.5`). The live path used the CLI copy, so the maintained/tested version's behavior never actually shipped.

## Change

- **Ported `find_fuzzy_duplicates` to `core::Directive`** (it had only test callers), and added a per-transaction `is_duplicate(new_txn, existing, config) -> bool` sharing one `is_fuzzy_duplicate` matcher. `find_structural_duplicates` stays on `DirectiveWrapper` — it has a real production caller (the `no_duplicates` native *plugin*), so it correctly lives at the plugin boundary (matching the audit's "confine `DirectiveWrapper` to the plugin boundary").
- **Deleted `extract_cmd/duplicate.rs`'s fuzzy logic** (`is_duplicate`, `first_posting_amount`, `txn_match_text`, `fuzzy_text_match`); `extract --existing` now calls `rustledger_ops::dedup::is_duplicate(...)`. Only the I/O helper `load_existing_transactions` remains.
- **Threshold single-sourced** on the ops config (`>= 0.5`, default). The CLI's divergent `> 0.5` is gone, pinned by a new boundary test: exactly 50% overlap is a duplicate, asserted through **both** `find_fuzzy_duplicates` and `is_duplicate`.

The CLI's now-redundant dedup unit tests were removed (the behavior is covered — and extended with the boundary case — by the ops tests).

## Tests

New `fuzzy_threshold_is_inclusive_at_exactly_half`; full `rustledger-ops` + `rledger extract` suites pass; clippy `--all-features --all-targets -D warnings` + fmt clean.

## Scope note

This is the fuzzy-dedup slice of the larger [L/4] item. Porting `transfer`/`ml`/`reconcile` off `DirectiveWrapper` (to drop their per-call core↔wrapper deep clones) remains a separate follow-up.

### DIFF
--- a/crates/rustledger-ops/src/dedup.rs
+++ b/crates/rustledger-ops/src/dedup.rs
@@ -15,7 +15,8 @@
 use std::collections::HashSet;
 
 use rust_decimal::Decimal;
-use rustledger_plugin_types::{DirectiveData, DirectiveWrapper, PluginError, TransactionData};
+use rustledger_core::{Directive, Transaction};
+use rustledger_plugin_types::{DirectiveData, DirectiveWrapper, PluginError};
 
 use crate::fingerprint::structural_hash;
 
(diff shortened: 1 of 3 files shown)


robcohen:
Both addressed: (1) introduced a `TxnKey` (date + amount + lowercased text) computed once per transaction — `find_fuzzy_duplicates` now pre-computes every existing key once (restoring the batch pre-compute the wrapper version had) and `is_duplicate` computes the new key once instead of per candidate. (2) `FuzzyDedupConfig::default()` is now built once before the `extract --existing` filter, not per directive.

PR #1567, opened 2026-06-24
fix(wasm): route parseMultiFile through process() for stream consistency (M/5)

robcohen:
## What

Architecture-roadmap [M/5] (WASM half — completes the item) — **route the WASM `parseMultiFile` surface through the canonical `process()` pipeline**, so its directive stream matches `validateMultiFile` / `queryMultiFile`.

`parse_multi_file` did `loader.load(...)` followed by a **manual per-transaction `interpolate` loop**, skipping sort, synth-plugins, and booking. A JS consumer therefore got a stream that was:
- **unsorted** (load order, not date order),
- **synth-free** (plugin-synthesized `Open`/`Document` directives never appeared), and
- merely **interpolated**, not booked,

— while `validateMultiFile` and `queryMultiFile` on the *same* files used `process()` and saw the sorted, synth-augmented, booked stream. Same inputs, inconsistent outputs across the multi-file surfaces.

## Change

On a clean parse, `parse_multi_file` now runs `process(load_result, &LoadOptions { validate: false, ..Default::default() })` — sort → synth-plugins → book → regular-plugins — and returns those directives. Validation stays **off**: this is the parse surface, so it surfaces parse + booking errors (collected into the existing `ParseResult.errors`) but not balance/assertion failures. On parse errors it keeps the raw parsed directives rather than booking a malformed ledger (mirroring the old "interpolate only when clean").

The `ParseResult` shape and the TypeScript `parseMultiFile(files, entryPoint): ParseResult` signature are unchanged — only the *contents* of the directive stream change (now consistent with the other surfaces).

## Contract note

This is a deliberate, JS-visible behavior change: the parsed directives are now sorted, include synthesized directives, and are booked. That is the audit's intended fix (the surfaces must agree). There are no Rust tests pinning the old behavior (these `JsValue` APIs have none — `validateMultiFile`/`queryMultiFile` likewise), and the change reuses the same heavily-tested `process()` pipeline. The doc comment now states the new behavior explicitly.

## Tests

`rustledger-wasm` builds + existing tests pass; clippy `--all-targets -D warnings` + fmt clean.

### DIFF
--- a/crates/rustledger-wasm/src/api.rs
+++ b/crates/rustledger-wasm/src/api.rs
@@ -440,11 +440,13 @@ pub fn bql_completions(partial_query: &str, cursor_pos: usize) -> Result<JsValue
 /// resolving `include` directives across the files. This enables multi-file ledgers
 /// in WASM environments where filesystem access is not available.
 ///
-/// The returned directives are run through the same processing pipeline as
-/// `validateMultiFile` / `queryMultiFile` (sort → synth-plugins → book →
-/// regular-plugins, validation excluded), so they are sorted, include
+/// On a clean parse, the returned directives are run through the same processing
+/// pipeline as `validateMultiFile` / `queryMultiFile` (sort → synth-plugins →
+/// book → regular-plugins, validation excluded), so they are sorted, include
 /// plugin-synthesized `Open`/`Document` directives, and have booked amounts —
-/// consistent with the other multi-file surfaces.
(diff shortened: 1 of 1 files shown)


robcohen:
Good catch — the doc now states the conditional explicitly: on a clean parse the directives are processed (sorted/synth/booked), but on parse errors the raw parsed directives are returned (the pipeline is not run on a malformed ledger) alongside the errors.

PR #1568, opened 2026-06-24
refactor(ops): port reconcile to core::Directive, drop the round-trip (L/4)

robcohen:
## What

Architecture-roadmap [L/4] (ops core-typed — reconcile slice) — **port `rustledger_ops::reconcile` to `core::Directive`, eliminating the audit's named `DirectiveWrapper` round-trip.**

`ops` promises "pure operations on core directives," but `reconcile` operated on the plugin **wire type** `DirectiveWrapper`. The audit's pointed example: `extract --balance` built a wrapper via `create_balance_directive` *only to convert it straight back* to a core `Directive` one line later via `wrapper_to_directive`.

## Change

- `create_balance_directive` now returns a `core::Directive::Balance` directly (built with `Balance::new(...).with_meta(...)`), so the CLI appends it with **no wrapper round-trip**. The wire-only synthetic `filename` is dropped (core directives carry source location via spans, not a field); the `import-reconcile` metadata is preserved.
- `reconcile` takes `&[Directive]` and reads typed core postings (`posting.amount()` → no `Decimal::from_str` on a wire string).
- `StatementBalance.date` is now a `NaiveDate` (parsed once at the CLI input boundary) instead of a `String`, keeping `create_balance_directive`/`reconcile` infallible.
- The `extract --balance` call site drops `wrapper_to_directive` and parses the balance date up front.

`reconcile()` itself has no production callers (tests only), so this is also a consistency/correctness alignment — but `create_balance_directive` is live, and its round-trip is now gone.

## Tests

The reconcile unit tests build core `Directive`s (instead of `DirectiveWrapper`s) and assert on the typed `Directive::Balance` (`Decimal` amount, `Account`, `meta`). Full `rustledger-ops` + `rledger extract` suites pass; clippy `--all-features --all-targets -D warnings` + fmt clean.

## Scope note

This is the **reconcile** slice of the L/4 ops core-typing (dedup landed in #1565). `transfer` and `ml` remain on `DirectiveWrapper` — `transfer` in particular carries a real per-call deep clone on the `rledger check` hot path (`find_transfers_in_ledger(&wrappers)`) and is the next, higher-impact slice.

### DIFF
--- a/crates/rustledger-ops/src/reconcile.rs
+++ b/crates/rustledger-ops/src/reconcile.rs
@@ -80,7 +80,7 @@ pub fn reconcile(
 
 /// Create a core balance-assertion [`Directive`] from a statement balance.
 ///
-/// Returns a `core::Directive` directly so the CLI can append it without a
+/// Returns a [`Directive`] directly so the CLI can append it without a
 /// `DirectiveWrapper` round-trip (it previously built a wrapper here only to call
 /// `wrapper_to_directive` on it one line later).
 #[must_use]
--- a/crates/rustledger-ops/src/reconcile.rs
+++ b/crates/rustledger-ops/src/reconcile.rs
@@ -5,16 +5,13 @@
 //! balance assertion directives for the ledger.
 
 use rust_decimal::Decimal;
-use rustledger_plugin_types::{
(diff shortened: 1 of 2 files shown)


robcohen:
Fixed — the doc now links the imported `[`Directive`]` type instead of the bare `core::Directive` path.

PR #1569, opened 2026-06-24
refactor(ops): port transfer detection to core::Directive (L/4)

robcohen:
## What

Architecture-roadmap [L/4] (ops core-typed — transfer slice) — **port `rustledger_ops::transfer` off the plugin wire type `DirectiveWrapper` onto `core::Directive`, eliminating the per-call deep clone in the `check` / `lint transfers` paths.**

Both consumers previously deep-cloned every directive into a `DirectiveWrapper` (`directive_to_wrapper_with_location`) just to run detection. That's the audit's "every CLI consumer O(n) deep-clones core→wrapper."

## Change

- New `LocatedDirective<'a> { directive: &'a Directive, filename, lineno }` carrier: detection now takes **borrowed** core directives plus their source location. No directive is cloned — only the small `(filename, lineno)` is owned. `find_transfers` / `find_transfers_in_ledger` / `find_matches_between` and the helpers (`first_posting_*`, `shares_link`, `classify_keywords`, `within_date_window`) all read core types.
- `TransferMatch` is **unchanged** — it still carries `filename`/`lineno`, now sourced from `LocatedDirective`. These feed the lint **idempotency hash** (`link_name_for`) and in-place **source rewriting** (`apply_links`), so the location is passed through **byte-identically** (the same source-map-derived values the wrapper carried) — re-applying transfer links stays a no-op.
- Behaviour preserved exactly, including edge cases: a `NumberOnly` first posting still matches with an empty currency and a `CurrencyOnly` one is still skipped (mirroring the old wire `number=""` parse failure). Date math moved from string-parsing to typed `jiff::civil::Date` arithmetic (`NaiveDate`).
- Consumers (`check.rs`, `lint/transfers.rs`) build `LocatedDirective`s borrowing `&spanned.value` instead of deep-cloning wrappers. `lint/transfers.rs` keeps the per-file `LoadResult`s alive so the flat borrowed view is valid.

## Tests

All 20 `transfer` unit tests ported to build core `Directive`s (a `find_in_groups`/`find_in_ledger` helper owns them and builds the borrowed view; the location test builds `LocatedDirective`s explicitly). `rustledger-ops` transfer (20) + `rledger` lint/transfers (11) suites pass; clippy `--all-features --all-targets -D warnings` + fmt clean.

## Scope note

This is the **transfer** slice of the L/4 ops core-typing (dedup #1565, reconcile #1568). Only **`ml`** now remains on `DirectiveWrapper`.

### DIFF
--- a/crates/rustledger-ops/src/transfer.rs
+++ b/crates/rustledger-ops/src/transfer.rs
@@ -17,9 +17,26 @@
 //! detector against an already-linked ledger is a no-op (idempotent).
 
 use rust_decimal::Decimal;
-use rustledger_plugin_types::{DirectiveData, DirectiveWrapper};
+use rustledger_core::{Directive, IncompleteAmount, Link, NaiveDate};
 use std::collections::{BTreeMap, HashSet};
-use std::str::FromStr;
+
+/// A core [`Directive`] paired with its source location — the input to transfer
+/// detection.
+///
+/// Carrying `(filename, lineno)` here keeps `ops` on `core::Directive` (no
+/// plugin `DirectiveWrapper` deep clone of the whole directive) while still
+/// letting [`TransferMatch`] report and rewrite `file:line`. The directive is
+/// borrowed; only the small location is owned, so cloning a `LocatedDirective`
+/// (e.g. during grouping) does not deep-copy the directive.
+#[derive(Debug, Clone)]
+pub struct LocatedDirective<'a> {
+    /// The directive (borrowed — not cloned).
+    pub directive: &'a Directive,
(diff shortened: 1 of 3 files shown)


robcohen:
Made the lifetime explicit (`Vec<LocatedDirective<'_>>`) for consistency with the other signatures. Note: it did compile as-is — lifetime elision infers the parameter in `&[(String, Vec<LocatedDirective>)]` (CI build + clippy were green) — but the explicit `<'_>` matches `find_transfers_in_ledger` and reads clearer.

Issue #1432, opened 2026-06-18
docs/UX: Python plugins referenced by module name fail confusingly; document file-path + sandbox-isolation requirements


robcohen:
Surfaced by discussion #1420 (zacchiro), where pip-installed beancount Python plugins failed under `rledger check` with a misleading error. The runtime behavior is intentional, but two gaps make it read as a bug.

## Gap 1: module-name plugin references are unsupported, but the error implies "not found"

`discover_module_source` (crates/rustledger-plugin/src/python/runtime.rs:457) only loads plugins given as a **file path** (`.py`, or a path containing a separator). A bare module name (beancount's standard `plugin "pkg.mod"` syntax) returns `ModuleNotFound` by design:

> "This intentionally does NOT auto-discover module sources via system Python. We want users to explicitly specify file paths so we can track which plugins need native Rust implementations."

The user-facing message is `Python plugin module not found: pkg.mod`, which implies rustledger searched for the module and could not find it. So users (reasonably) suspect a venv / PYTHONPATH problem rather than an unsupported declaration form.

## Gap 2: the sandbox does not mount the host venv

Python plugins run in a pinned CPython-WASI sandbox (wasmtime) with only `/work` (the plugin source plus the `beancount.core.data` compat shim) and `/lib` (bundled stdlib) on `sys.path`. The host venv's `site-packages` is never mounted (no `site-packages` / `VIRTUAL_ENV` handling exists). So even with the file-path workaround, a plugin loads only if it is **self-contained**: standard library plus whatever the compat shim provides.

## Proposed changes

1. **Better error.** When a plugin reference is a module name (no `.py`, no path separator), say it is unsupported and that an explicit file path is required, rather than "module not found". `suggest_module_path()` already resolves the module to a file via system Python for exactly this purpose; print it as the hint, e.g. `plugin "pkg.mod" is not supported by name; reference the file directly: plugin "/abs/path/to/pkg/mod.py"`.
2. **Document the model in `docs/reference/plugins.md`.** Python plugins must be referenced by file path; the sandbox cannot see the host venv / `site-packages`, so plugins must be self-contained (stdlib plus the beancount compat shim). List what the compat shim covers (`beancount.core.data` namedtuples, etc.) and point users to native Rust plugins for anything heavier.
3. **Optional:** a short note in `docs/migration/from-beancount.md`, since beancount's module-name plugin syntax does not carry over.

## Acceptance

[...]

PR #1571, opened 2026-06-24
refactor: extract plugin dispatch into rustledger-plugin (resolve/run split, M/4 both stages)

robcohen:
## What

Architecture-roadmap [M/4] — extract plugin dispatch out of the `run_plugins` god-function. **Two stages, both in this PR:**

**Stage 1 (loader-internal):** split the inline native/WASM/Python dispatch into a `resolve` / `run` shape so the loop builds wrappers + applies ops once per plugin instead of once per runtime (~200 lines of branching → ~30).

**Stage 2 (cross-crate):** move dispatch into `rustledger-plugin` behind one entry point, leaving the loader to orchestrate passes and apply ops.

## Stage 2 detail

- New `rustledger_plugin::dispatch`: `resolve_plugin()` (classify → path-security → feature-gate → #1432 module-name reject) returns a `ResolvedPlugin` enum; `ResolvedPlugin::run()` executes the chosen runtime. Errors are **typed** (`PluginResolveError` / `PluginRunError`) — not host `LedgerError`s — so `rustledger-plugin` stays runtime-knowledge-pure and the loader keeps ownership of its `E8001/E8002/E8004/E8005` codes via `resolve_error_to_ledger` / `run_error_to_ledger`.
- Path-security (`path_within_base` + `lexically_normalize`), the file-vs-module classifier, and the WASM/Python runners all moved into the dispatcher.
- Feature gates align across the boundary: the loader pulls `rustledger-plugin` with `default-features = false`, so loader-`wasm-plugins` ⟺ plugin-`wasm-runtime` and loader-`python-plugins` ⟺ plugin-`python-plugins`. A native-only build still yields the exact "requires the wasm-plugins feature" error.

## Intentional behavior change (a bug fix)

Previously the loader's WASM/Python runners converted plugin errors to `LedgerError` **without** propagating `source_file`/`line_number`, while native used `record_plugin_errors` (which does). Now all runtimes return raw `Vec<PluginError>` and the loader records them through **one uniform path** — so a **WASM/Python plugin error that carries a source location now shows it** (Python plugins set it; the loader was silently dropping it). All error **codes, messages, severity, and `.with_phase("plugin")` are byte-for-byte preserved** — only the optional location is now retained.

## Verification

3-angle deep review (behavior-preservation, cross-crate cfg-correctness, line-by-line). It caught and I fixed: a **path-security regression** (a collapsed `match` reopened the #1564 fail-open hole on non-`NotFound` canonicalize errors — restored to fail-closed, now covered by tests), a **test-compile break** (a moved classifier's tests relocated to the dispatcher), and a clippy-pedantic doc lint. Clippy `-D warnings` clean on both the default (native-only) and `--all-features` configs; `rustledger-plugin` + `rustledger-loader` + `rustledger` CLI suites pass under all-features; fmt clean.

### DIFF
--- a/crates/rustledger-loader/src/process.rs
+++ b/crates/rustledger-loader/src/process.rs
@@ -869,56 +869,6 @@ struct PluginInvocation {
     force_python: bool,
 }
 
-/// Lexically resolve `.` / `..` in `p` WITHOUT touching the filesystem,
-/// preserving the root/prefix. Unlike `Path::canonicalize` this works on paths
-/// that don't exist yet, so a `..` traversal is still collapsed (e.g.
-/// `/ledger/../../etc` → `/etc`). A `..` at the root is clamped (it can't escape
-/// above the root).
-#[cfg(any(feature = "wasm-plugins", feature = "python-plugins"))]
-fn lexically_normalize(p: &Path) -> std::path::PathBuf {
-    let mut out = std::path::PathBuf::new();
-    for component in p.components() {
-        match component {
-            std::path::Component::ParentDir => {
-                out.pop();
-            }
-            std::path::Component::CurDir => {}
-            other => out.push(other.as_os_str()),
-        }
-    }
-    out
-}
-
-/// Whether a `resolved` plugin path stays within `base_dir` under path-security.
-///
-/// When the plugin file exists, both sides are `canonicalize`d so symlinks are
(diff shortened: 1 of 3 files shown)


robcohen:
Fixed — the doc now says the result is absolute only when `base_dir` is (a relative `name` is joined onto `base_dir` as-is, which can be relative under the test VFS).

PR #1572, opened 2026-06-24
refactor: deny unwrap/expect on wasm + loader never-panic surfaces (L/4)

robcohen:
## What

Architecture-roadmap [L/4] (deny-panic slice) — **enable `clippy::unwrap_used` + `clippy::expect_used` as `deny` on the two never-panic surfaces with the fewest production sites: `rustledger-wasm` and `rustledger-loader`.** A stray `unwrap`/`expect` on these paths is a real crash vector (WASM ships `panic = "abort"`; the loader handles attacker-controlled `include` paths and parser output), and the lints were previously not enabled, so a new one would compile and ship.

## Change

- Both crate roots gain `#![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used))]`. The `not(test)` scopes the deny to **production** code and exempts all test code — including `#[cfg(all(test, feature = ...))]` modules, which clippy's `allow-unwrap-in-tests` config does *not* recognize as test context (that was my first attempt; `cfg_attr(not(test))` is the robust, cfg-shape-agnostic fix).
- Audited the production sites the lint flagged:
  - `wasm/cache.rs`: `header[8..12].try_into().unwrap()` → rewritten to `from_le_bytes([header[8], header[9], header[10], header[11]])`. `header` is exactly `HEADER_SIZE` (12) bytes (length-checked above), so the indices are always present — **no unwrap, no panic.**
  - `wasm/helpers.rs`: `naive_date(2999, 12, 31).unwrap()` — a known-valid calendar date, so an **audited `#[allow(clippy::unwrap_used)]`** with a justification comment.
  - `loader`: 0 production sites — the deny just locks the invariant going forward.

## Verification

Clippy `-D warnings` clean on **both** the default and `--all-features --all-targets` configs (the all-features build is what surfaced the feature-gated test-helper sites that drove the `cfg_attr(not(test))` decision); `rustledger-wasm` + `rustledger-loader` suites pass; fmt clean.

## Scope

This is the wasm+loader slice of the deny-panic item. The **parser** (~19 production `unwrap`/`expect` sites — the CLAUDE.md "must handle malformed input gracefully" surface) and **`clippy::indexing_slicing`** (the noisy one, needing many audited `#[allow]`s) are the larger follow-ups. The other three parts of the original roadmap item (CI bench job, `fuzz_query_execute` target, parallel-threshold brackets) are independent and out of scope here.

### DIFF
--- a/crates/rustledger-loader/src/lib.rs
+++ b/crates/rustledger-loader/src/lib.rs
@@ -27,6 +27,12 @@
 
 #![forbid(unsafe_code)]
 #![warn(missing_docs)]
+// Never-panic surface: the loader resolves attacker-controlled `include` paths
+// and parser output, so production code must not `unwrap`/`expect`. `not(test)`
+// scopes the deny so test code (incl. `#[cfg(all(test, feature = ...))]` modules)
+// may unwrap/expect freely. Proven-safe production sites carry an audited
+// `#[allow]` with a justification.
+#![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used))]
 
 #[cfg(feature = "cache")]
 pub mod cache;
--- a/crates/rustledger-loader/src/lib.rs
+++ b/crates/rustledger-loader/src/lib.rs
@@ -29,9 +29,11 @@
 #![warn(missing_docs)]
 // Never-panic surface: the loader resolves attacker-controlled `include` paths
 // and parser output, so production code must not `unwrap`/`expect`. `not(test)`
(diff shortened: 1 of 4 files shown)


robcohen:
Tightened both comments. You're right — `cfg_attr(not(test))` exempts only this crate's own `cfg(test)` code (unit tests / `#[cfg(test)]` modules); integration tests under `tests/` are separate crates and aren't governed by this attribute either way (and the lib they link is correctly denied as production). The wording now says exactly that.

PR #1574, opened 2026-06-24
refactor(parser): deny unwrap/expect on the parser never-panic surface (L/4)

robcohen:
## What

Architecture-roadmap [L/4] (deny-panic — parser slice) — **enable `clippy::unwrap_used` + `clippy::expect_used` as `deny` on `rustledger-parser`**, the most security-critical never-panic surface (CLAUDE.md: *"Parser: must handle malformed input gracefully (no panics)"*). Continues #1572 (wasm + loader) with the same `cfg_attr(not(test))` mechanism.

## Change

- `#![cfg_attr(not(test), deny(clippy::unwrap_used, clippy::expect_used))]` on the parser crate root.
- Audited every production site clippy flagged:
  - **Fixed (input-reachable)** — the transaction header-string match (`convert.rs`) no longer counts-then-`unwrap()`s; it consumes via `it.next()` (slice of `Option`s), preserving the exact 0/1/2/3+ behavior with no panic. The range-snapshot `unwrap`s (`format.rs`) became a `let-else` that returns `None`. `format_node_range` (which already returns `Option`) now `?`s the cast instead of `expect`ing.
  - **Audited `#[allow]` (proven-safe invariants)** — `SourceFile::parse` (the `parse_structured` root is *always* `SOURCE_FILE`) and `format_node`'s `SOURCE_FILE` precondition (a wrong node is a caller bug, not input-driven). Each carries a justification comment.

## Verification

Clippy `-D warnings` clean on **both** the default and `--all-features --all-targets` configs (all-features surfaces feature-gated/test sites — none remained after the `cfg_attr(not(test))` scoping). `rustledger-parser` test suite passes (the header-string rewrite preserves behavior); fmt clean.

## Scope

This extends deny-panic to the parser (after wasm + loader in #1572). `clippy::indexing_slicing` (the noisy one, many audited `#[allow]`s) and the CI `cargo bench` job remain as the documented L/4 leftovers.

### DIFF
--- a/crates/rustledger-parser/src/cst/ast.rs
+++ b/crates/rustledger-parser/src/cst/ast.rs
@@ -485,7 +485,11 @@ impl SourceFile {
     #[must_use]
     pub fn parse(source: &str) -> Self {
         let node = crate::cst::parser::parse_structured(source);
-        Self::cast(node).expect("parse_structured always returns a SOURCE_FILE")
+        // `parse_structured` always produces a SOURCE_FILE root, so this cast is
+        // infallible — a parser invariant, not an input-driven path.
+        #[allow(clippy::expect_used)]
+        let source_file = Self::cast(node).expect("parse_structured always returns a SOURCE_FILE");
+        source_file
     }
 
     /// All recognized directives, in source order.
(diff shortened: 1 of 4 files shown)


robcohen:
Addressed all three: (1) dropped the header-string pre-clone — the 3+ arm now uses `it.last().unwrap_or(c)`, so no clone in the common 0/1/2 cases; (2) fixed `format_node_range`'s rustdoc (returns `None` on a non-SOURCE_FILE node, no longer panics); (3) added `transaction_three_plus_header_strings_surface_last_as_narration` locking the 3+ behavior (the lossless CST does keep 3+ header strings, so this path is reachable).

PR #1575, opened 2026-06-24
refactor(query): route BALANCES through scan_postings; fix ignored FROM window (XL/8 tail)

robcohen:
## What

Architecture-roadmap [XL/8 tail] — **route `build_balances_with_filter` (the BALANCES command) through the shared `scan_postings` stream**, finishing the posting-iteration unification (the other consumers — default SELECT, `#postings`, aggregation — already share `scan_postings`). **This also fixes a latent bug.**

## The bug it fixes

`BalancesQuery.from` is a full `FromClause` (the parser accepts `BALANCES FROM ... OPEN ON d CLOSE ON d` — there's a parse test), but `build_balances_with_filter` applied **only** `from.filter`, silently **ignoring `open_on`/`close_on`**. So `BALANCES FROM CLOSE ON 2024-01-22` summed *every* posting regardless of the close date — diverging from the SELECT path (and bean-query), which window correctly.

(`OPEN ON` is a no-op for BALANCES *totals* — its pre-date carry-in keeps those postings in the sum — so the visible bug is the `CLOSE ON` / window-upper-bound side.)

## Change

- `scan_postings` now returns a small `PostingScan { postings, account_balances }` (it already maintained `account_balances` internally for the `account_balance` column; this just exposes it). The two existing callers take `.postings`.
- `build_balances_with_filter` becomes a one-liner: `scan_postings(from, None, /*needs_balance*/ false, /*needs_account_balance*/ true, false)?.account_balances`. That map is the **windowed, FROM-filtered, WHERE-independent** per-account total — exactly what BALANCES wants — and it shares the single `Position::from_posting` cost resolution.

## Behavior

For the common case (no `OPEN/CLOSE ON`), the result is **identical** (both sum FROM-filtered postings per account — the full existing BALANCES suite passes unchanged). The only change is the **fix**: `BALANCES FROM CLOSE ON d` now honors the window, consistent with `SELECT ... FROM CLOSE ON d`. New regression test `test_balances_honors_close_on_window` pins it.

## Verification

New regression test passes (proving the window is applied); full `rustledger-query` suite green (no SELECT/`#postings`/BALANCES regression); clippy `-D warnings` + fmt clean.

### DIFF
--- a/crates/rustledger-query/src/executor/mod.rs
+++ b/crates/rustledger-query/src/executor/mod.rs
@@ -405,7 +405,7 @@ impl<'a> Executor<'a> {
         // — `BALANCES` applies its own `WHERE` to the result afterward, and
         // `account_balances` is WHERE-independent by construction anyway.
         Ok(self
-            .scan_postings(from, None, false, true, false)?
+            .scan_postings(from, None, false, true, false, false)?
             .account_balances)
     }
 
@@ -450,6 +450,7 @@ impl<'a> Executor<'a> {
                 needs_balance,
                 needs_account_balance,
                 where_reads_balance,
+                true,
             )?
             .postings)
     }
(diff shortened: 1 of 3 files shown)


robcohen:
Good catch — fixed. Added a `collect_contexts` flag to `scan_postings`: BALANCES (`build_balances_with_filter`) now passes `false`, so it accumulates `account_balances` without building/pushing a `PostingContext` (or its per-posting `Inventory` clone) for every posting — the stream it would immediately discard. SELECT / `#postings` pass `true` (unchanged). BALANCES results are identical (full suite + the windowing tests still green); only the wasted materialization is gone.

PR #1576, opened 2026-06-24
feat(parser): emit MetaValue::Int for integer-literal metadata (L/7 codec phase 2)

robcohen:
## What

Architecture-roadmap **[L/7 MetaValue::Int codec — Phase 2]**: make the **parser emit `MetaValue::Int`** for integer-literal metadata, completing the integer-metadata type fidelity feature.

The codec's groundwork (the `MetaValue::Int(i64)` variant, `CACHE_VERSION` 11, every exhaustive match/wire arm, and `lineno → Int` in BQL) already landed earlier. This PR adds the **behavioral** piece — `key: 42` now parses as `Int(42)` rather than `Number(42)` — plus two round-trip gaps the groundwork left.

## Change

- **Parser** (`cst/convert.rs`): a shared `number_meta_value(token_text, value)` helper applied at all three metadata-number sites (`meta_value_from_entry`, custom-directive values, `pushmeta`). An **integer literal** — no `.`/`e`/`E` in the source text and within `i64` range — becomes `Int`; **decimals, exponents, and `i64`-overflow stay `Number`** (lossless). A `NUMBER CURRENCY` pair is still `Amount`; sign handling is unchanged.
- **Plugin `from_wrapper`**: the wire has no integer case (`to_wrapper` already sends `Int` as a numeric string), so the return path now tries `i64` first — integer metadata round-trips as `Int` instead of degrading to `Number`.
- **LSP `meta_to_f64`**: added the `Int` arm (was falling into `_ => None`).

## Why it's safe

I audited every `MetaValue::Number` consumer: there are **no partial `if let MetaValue::Number` matches** that could silently stop matching, and every exhaustive match already carries an `Int` arm (added with the groundwork). `parse_precision_meta` accepts `Int`, so `precision: N` is unaffected (verified by the round-trip test below).

## Tests

- New `metadata_integer_vs_decimal_vs_overflow` (parser): `42 → Int`, `-7 → Int`, `42.0`/`4.2e1` → `Number`, an `i64`-overflowing literal stays `Number`.
- Updated the existing assertions that pinned the old `Number` behavior for integer literals to expect `Int` (`open_directive_with_metadata`, `custom_directive_heterogeneous_values`, and `test_commodity_precision_metadata_round_trips` — the last confirms `precision: 2` round-trips `Int(2)` through format→reparse).

## Verification

Workspace builds clean (`--all-features --all-targets`); **all 10 affected crates' test suites pass** (parser, core, plugin, query, validate, wasm, lsp, wire-format-tests, ffi-wasi, ffi-component); clippy `-D warnings` clean. The BQL compat suite runs in CI.

### DIFF
--- a/crates/rustledger-core/src/directive.rs
+++ b/crates/rustledger-core/src/directive.rs
@@ -53,6 +53,12 @@ pub enum MetaValue {
     Amount(Amount),
     /// Null/None value
     None,
+    /// Integer value (e.g. beancount `key: 42`, distinct from `42.0`).
+    ///
+    /// MUST remain the LAST variant: rkyv encodes enum discriminants by
+    /// declaration order, so appending keeps the existing variants' discriminants
+    /// stable. `CACHE_VERSION` (rustledger-loader) was bumped when this was added.
+    Int(i64),
 }
 
 impl fmt::Display for MetaValue {
@@ -68,6 +74,7 @@ impl fmt::Display for MetaValue {
             Self::Bool(b) => write!(f, "{b}"),
             Self::Amount(a) => write!(f, "{a}"),
             Self::None => write!(f, "None"),
+            Self::Int(i) => write!(f, "{i}"),
         }
     }
 }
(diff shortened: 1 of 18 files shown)


robcohen:
Closing as redundant: the MetaValue::Int codec — including the parser emitting Int for integer literals (the phase-2 deliverable), cache v12, all wire/match arms, and lineno->Int — is already on main, merged via a separate PR. This branch was stale (predated that merge and duplicated it with an obsolete cache-v11 commit). Verified main has the full codec; nothing to add here.

PR #1577, opened 2026-06-24
ci: compile criterion benches so they do not bitrot (L/4)

robcohen:
## What

Architecture-roadmap **[L/4 leftover — `cargo bench` CI]**: the workspace has 6 criterion bench targets (`parser_bench`, `query_bench`, `validate_bench`, `inventory_bench`, `pipeline_bench`, `v016_baseline`) that **compile nowhere in CI** — so a source change can silently break a bench build and no one notices until someone runs `cargo bench` locally.

A stale comment in `bench-pr.yml` even claimed *"Criterion benchmarks removed - they're tracked in ci.yml on main branch pushes"* — but `ci.yml` has **zero** bench references. They were bitrot-exposed.

## Change

- **New `Benchmarks Compile` job in `ci.yml`** — runs `cargo bench --workspace --no-run`, gated on `should_build` like the other code jobs. **Compile-only**: actually *running* criterion benches on shared CI runners is timing-noisy and adds no gating value; the real risk this guards is a bench that stops *building*.
- Wired into the **`build` gate** (`needs:` + the results check) so a broken bench build blocks merge, consistent with `doctest`/`regressions`.
- **Corrected the stale `bench-pr.yml` comment** to describe the new reality (criterion benches are compiled — not run — by the new job).

## Verification

`cargo bench --workspace --no-run` builds all six bench executables clean (0 errors) locally; both workflow YAMLs parse valid. No source changes — CI-config only.

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -143,8 +143,10 @@ jobs:
 
           echo "rustledger: ${rustledger_mb}MB"
           echo "beancount:  ${beancount_mb}MB"
-      # Note: Criterion benchmarks removed - they're tracked in ci.yml on main branch pushes
-      # and weren't being reported in the PR comment anyway
+      # Note: Criterion benchmarks aren't run here (timings are noisy on shared
+      # runners and weren't reported in the PR comment anyway). They are compiled
+      # — but not run — by the `Benchmarks Compile` job in ci.yml, which guards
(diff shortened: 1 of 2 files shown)


robcohen:
Reopening under a compliant branch name (ci/ is not an allowed prefix; the branch-name check requires feat|fix|docs|chore|refactor|... ). Same change, new branch chore/cargo-bench.

PR #1579, opened 2026-06-24
refactor(booking): route apply onto the shared Position::from_posting (L/7)

robcohen:
## What

Architecture-roadmap **[L/7 inventory-accumulation tail]**: route `BookingEngine::apply`'s inventory addition onto the canonical **`Position::from_posting`** — the same cost-resolve helper the Late validator (`process_inventory_addition`), `build_balances`, and the query engine already share. `apply` was the lone remaining hand-roller.

## The investigation

The roadmap framed this as "Late validation reuses booked inventories" — but that's **infeasible**: the validator's inventories are *incremental* (point-in-time balance assertions), *pad-adjusted*, and *error-emitting*, none of which booking's final-state inventories provide. The reduction-detection (`is_booking_reduction`) was the genuinely-shared piece and is already unified.

The real dedup is the inverse: `apply` hand-rolled ~40 lines (per-unit match + cost-currency inference + `Position::with_cost`) to build a position that `Position::from_posting` already builds. So this replaces that block with a one-liner.

## Why it's equivalent

- **per-unit / date / label**: identical — `CostSpec::resolve`'s own doc says *"Matches the zero-units guard in BookingEngine::apply"*.
- **cost currency**: the old inline price/cross-posting inference was **dead code** for `apply`'s actual inputs. `apply` only runs on *booked* transactions (every caller `book_and_interpolate`s first — see its `debug_assert`), and `book()` fills the inferred currency *into* `cost_spec.currency` for augmentations (book.rs ~368-405). So by apply-time the currency is already resolved, and `from_posting` (which reads `cost_spec.currency`) produces the same cost.

## Test

`test_cost_spec_currency_inference` previously called `apply` **directly** on an unbooked txn to exercise the standalone inference. Updated it to use the real `book_and_interpolate → apply` pipeline — it still asserts the price-inferred USD cost, now **proving** booking fills the currency and the dedup is behavior-preserving (rather than pinning a path production never takes).

## Verification

Full `rustledger-booking` suite green (104+ tests incl. all lot-matching + the inference test); clippy `-D warnings` clean; the booking fuzz target compiles. Public API unchanged.

### DIFF
--- a/crates/rustledger-booking/fuzz/Cargo.lock
+++ b/crates/rustledger-booking/fuzz/Cargo.lock
@@ -23,6 +23,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "archery"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70e0a5f99dfebb87bb342d0f53bb92c81842e100bbb915223e38349580e5441d"
+
+[[package]]
 name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


robcohen:
Good precision catch — reworded both comments. They now describe the **supported pipeline** (production always `book_and_interpolate`s before `apply`, per the `debug_assert`) rather than claiming `apply` is *only ever* called on booked txns, and note that direct-`apply` unit tests use explicit-currency fixtures (which need no inference).

PR #1580, opened 2026-06-24
fix(check): treat E7003 duplicate option as a warning, not an error (#1546)

### LINKED ISSUE #1546: [Bug]: rledger check reports E7003 (duplicate option) as an error for multi-entity include trees, diverging from beancount
## Summary

`rledger check` treats **E7003 (duplicate non-repeatable option)** as a hard error, but Python beancount does **not** error when a scalar option is redefined — it silently takes the last value. This makes `rledger check` reject a common, legitimate layout: a master ledger that `include`s several self-contained sub-ledgers, each of which sets its own `option "title"` / `booking_method` / `account_rounding` / etc. for standalone use.

Notably, the **loader already records E7003 as a warning**, so the loader/`validate` path reports the ledger as valid. Only `cmd/check` promotes it to an error, so `validate` and `check` disagree on the same file.

## Environment
- rustledger v0.16.5

## Repro

`a.beancount`:
```beancount
option "title" "Entity A"
2020-01-01 open Assets:A
```
`b.beancount`:
```beancount
option "title" "Entity B"
2020-01-01 open Assets:B
```
`master.beancount`:
```beancount
option "title" "Combined"
include "a.beancount"
include "b.beancount"
```
```console
$ rledger check master.beancount
error[E7003]: Option "title" can only be specified once
error[E7003]: Option "title" can only be specified once

$ bean-check master.beancount      # Python beancount: no error
$ rledger validate ... # loader path: "Ledger is valid"
```

## Root cause

`crates/rustledger/src/cmd/check.rs` (~L437):
```rust
// Report option errors (E7001, E7002, E7003)
// In Python beancount, invalid options are errors, not warnings
for warning in &load_result.options.warnings { /* emitted as errors */ }
```

The justification — *"In Python beancount, invalid options are errors"* — is correct for **E7001** (unknown option) and **E7002** (invalid value), which beancount rejects. It is **not** correct for **E7003** (duplicate non-repeatable option): beancount permits redefining a scalar option with last-wins semantics; only list options accumulate. rustledger already models the list case correctly via `REPEATABLE_OPTIONS` in `crates/rustledger-loader/src/options.rs`. Lumping E7003 in with E7001/E7002 over-promotes a beancount-legal construct to an error.

## Impact

Breaks the standard multi-entity / "master ledger" pattern, where each entity ledger validates standalone *and* is combined via `include`. A combined ledger that is valid in beancount — and that passes rustledger's own loader/`validate` — fails `rledger check` purely because each sub-ledger declares `title`/`booking_method`.

## Suggested fix (any of)

1. **Don't promote E7003 to an error in `cmd/check`** — keep it a warning (matching the loader and `validate`), while still erroring on E7001/E7002.
2. **Implement beancount last-wins** for duplicate non-repeatable scalar options (take the last value, warn at most). For an include tree this gives natural precedence — the most-recently-included / nearest declaration wins.
3. At minimum, gate E7003-as-error behind an opt-in flag.

This is essentially the loader-vs-validator unification discussed in #858; the divergence still lives in `cmd/check`.

## Workaround

Move options into standalone-only files that the master `include` doesn't pull in. Undesirable for genuinely self-contained entity ledgers.

robcohen:
## What

Fixes **#1546**: `rledger check` treated **E7003 (duplicate non-repeatable option)** as a hard error, diverging from `bean-check` (which silently takes the last value, exit 0) **and** from rledger's own loader / `validate` (which record it as a *warning*). This rejected a legitimate, common layout — a master ledger that `include`s several self-contained sub-ledgers, each declaring its own `option "title"` / `booking_method` / … for standalone use.

## Root cause

`cmd/check.rs` looped over *all* `load_result.options.warnings` and emitted each as `severity: "error"`, lumping E7003 in with E7001 (unknown option) and E7002 (invalid value). The error-promotion was a *deliberate* "stricter than beancount" deviation (the rationale: a same-file duplicate is usually a typo) — but it can't distinguish that from the legitimate cross-`include` case, so it over-rejected.

## Change

- **`cmd/check.rs`**: E7001/E7002 stay errors (beancount rejects those); **E7003 is now a warning** — emitted with `warning` severity, counted in `warning_count` (not `error_count`), and no longer fails the check. The option **value was already last-wins** (the loader applies the latest), so only the error-promotion needed removing.
- **`loader/options.rs`**: updated the stale rationale comment (no logic change — it already emits E7003 as a warning).

## Tests

- `test_check_duplicate_option_is_error` → **`test_check_duplicate_option_warns`**: now asserts `check` **exits 0** and surfaces E7003 as a *warning*.
- New **`test_check_multi_entity_include_duplicate_option_ok`**: the exact issue repro — `master.beancount` including `a`/`b` sub-ledgers that each set `option "title"` — passes `check`.
- The loader's `test_duplicate_option_warning` is unchanged (still emits the warning).

## Verification

`rustledger` + binary build clean; the E7003 CLI tests + the loader test pass; clippy `-D warnings` clean. `validate` and `check` now agree, both matching beancount.

Closes #1546

### DIFF
--- a/crates/rustledger-loader/src/options.rs
+++ b/crates/rustledger-loader/src/options.rs
@@ -246,10 +246,12 @@ impl Options {
 
         // Check for duplicate non-repeatable options (E7003).
         //
-        // Stricter than beancount: `bean-check` silently lets the last value
-        // win (exit 0). We flag the duplicate so the CLI can surface it as an
-        // error (a repeated non-repeatable option is almost always a mistake) —
-        // a DELIBERATE deviation. See `cmd::check` E7003 handling.
+        // Emitted as a WARNING (not an error), matching `bean-check`, which
+        // silently lets the last value win (exit 0). A master ledger that
+        // `include`s self-contained sub-ledgers — each setting its own
+        // `option "title"` / `booking_method` / ... for standalone use — is a
+        // legitimate layout (issue #1546). The value below is applied last-wins
+        // to match. `cmd::check` and `validate` both surface this as a warning.
         let is_repeatable = REPEATABLE_OPTIONS.contains(&key);
         if is_known && !is_repeatable && self.set_options.contains(key) {
             self.warnings.push(OptionWarning {
--- a/crates/rustledger/src/cmd/check.rs
+++ b/crates/rustledger/src/cmd/check.rs
@@ -463,7 +463,8 @@ pub fn run_with_writer<W: Write>(args: &Args, stdout: &mut W) -> Result<ExitCode
 
     // All option warnings collected by `Options::set` (E7001 unknown option,
     // E7002 invalid value, E7003 duplicate non-repeatable, E7004/E7005/E7006
-    // read-only and related) are surfaced here as hard errors.
+    // read-only and related) are surfaced here. Everything except E7003 is a
+    // hard error; E7003 is a warning (see below).
     //
     // E7001/E7002 match beancount: `bean-check` exits non-zero on an unknown
     // option or an invalid option value.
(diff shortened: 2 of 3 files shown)


robcohen:
Fixed — the header paragraph now says everything except E7003 is a hard error and E7003 is a warning, matching the loop below.

PR #1581, opened 2026-06-24
feat(ci): nightly dhat heap profiling of the load+process pipeline (slice 1)

robcohen:
## What

**Nightly heap profiling — slice 1.** Adds a `dhat` heap-profiling harness for the `load → process` pipeline plus a nightly workflow that runs it over a generated ledger and uploads the profile as an artifact. This complements `bench.yml` (wall-clock timing / regressions) by answering *"where do the allocations come from"* — the optimization-opportunity radar you asked for. **Advisory only — uploads an artifact, never gates merges.**

(Verified first that `[profile.bench] debug = true` and PGO already exist; this reuses the PGO-style workload generator and the bench artifact pattern rather than rebuilding them.)

## Changes (4, all additive)

- **`Cargo.toml`** — new `[profile.profiling]` (inherits `release` but keeps `debug` symbols, since `release` `strip`s them — required for symbolized dhat backtraces) + `dhat` as an optional workspace dep.
- **`crates/rustledger/Cargo.toml`** — `dhat-heap` feature (`dep:dhat`) + the optional dep. Off by default; **never in the shipped binary** (confirmed: dhat absent from the default dependency tree).
- **`crates/rustledger/examples/profile_pipeline.rs`** — runs the one-shot `rustledger_loader::load(path, &opts)` (parse + includes + options → book + plugins + validate, same as `rledger check`) under dhat's global allocator; writes `dhat-heap.json`. Builds as a no-op without the feature, so it stays green in `--all-targets`.
- **`.github/workflows/profile.yml`** — nightly (3 AM UTC, offset from bench's 2 AM) + `workflow_dispatch`; generates a 10K-txn workload, runs the harness, writes a step-summary pointer, uploads `dhat-heap.json` + the workload as an artifact (30-day retention).

## How to use

Download the **dhat-heap-profile** artifact from a nightly run and open `dhat-heap.json` at <https://nnethercote.github.io/dh_view/> — it shows which call paths allocate the most bytes.

Locally: `cargo run --profile profiling --example profile_pipeline --features dhat-heap -- <ledger.beancount>`.

## Deliberately deferred (future slices)

flamegraphs (`cargo flamegraph`), `iai-callgrind` deterministic instruction counts, and a `profiling` data branch for night-over-night trends.

## Verification

Example compiles **with and without** the feature; `dhat` absent from the default tree; the harness runs the full pipeline and emits a valid `dhat-heap.json` (v2); `profile.yml` YAML valid; clippy `-D warnings` clean. No production code touched.

### DIFF
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -1,0 +1,93 @@
+# Nightly heap profiling — surfaces allocation hotspots in the load → process
+# pipeline so we can spot optimization opportunities. Complements `bench.yml`
+# (wall-clock timing / regressions); this answers "where do the allocations
+# come from". Advisory only — uploads an artifact, never gates merges.
+#
+# Slice 1 (this file): a `dhat` heap profile of `examples/profile_pipeline`
+# over a generated ledger, uploaded as an artifact. Future slices may add
+# flamegraphs, iai-callgrind instruction counts, and a `profiling` data branch
+# for night-over-night trends.
+name: Profile
+
+on:
+  schedule:
+    - cron: "0 3 * * *" # Nightly at 3 AM UTC (offset from bench.yml's 2 AM)
+  workflow_dispatch:
+    inputs:
+      transactions:
+        description: "Number of transactions in the generated workload"
+        default: "10000"
+
+# Profiling runs are independent; never cancel an in-flight one.
+concurrency:
+  group: profiling
+  cancel-in-progress: false
+
+permissions:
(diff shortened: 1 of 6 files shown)


robcohen:
Both fixed: the doc-comment run command now includes `-p rustledger` (needed from the workspace root), and the harness now prints the `ledger.errors.len()` pipeline-error count alongside the directive count so a profile over an erroring ledger is not silently reported as clean.

PR #1583, opened 2026-06-24
feat(ci): nightly CPU flamegraph of the load+process pipeline (profiling slice 2)

robcohen:
## What

**Nightly profiling — slice 2: CPU flamegraphs.** Adds a `pprof`-based CPU-profiling harness for the `load → process` pipeline and wires it into the existing nightly `profile.yml`, so each run now uploads **both** a `dhat-heap.json` (where the bytes go, slice 1) **and** a `flamegraph.svg` (where the time goes). Advisory only — uploads artifacts, never gates merges.

## Why pprof (not `cargo flamegraph`)

`pprof` samples **in-process** (signal-based), so it needs **no `perf` privileges** — there's no `perf_event_paranoid` precedent in this repo's runners, and the criterion `pprof`-integration has version-coupling risk with criterion 0.8. This keeps the job robust and mirrors slice 1's structure exactly (feature-gated optional dep + example).

## Changes (3, all additive)

- **`Cargo.toml`** — `pprof` (with `flamegraph` feature) as an optional workspace dep.
- **`crates/rustledger/Cargo.toml`** — `flamegraph` feature (`dep:pprof`) + the optional dep. Off by default; **never in the shipped binary** (confirmed: pprof absent from the default tree).
- **`crates/rustledger/examples/flamegraph_pipeline.rs`** — loops the one-shot `load` under a `pprof` guard and writes `flamegraph.svg`. No-op without the feature, so it stays green in `--all-targets`.
- **`.github/workflows/profile.yml`** — adds a build + run step and includes `flamegraph.svg` in the (renamed) **nightly-profile** artifact; summary now points to both.

## Verification (local, before CI)

Example compiles **with and without** the feature; `pprof` absent from the default tree; the harness ran 2000 iterations and produced a **valid 67 KB `flamegraph.svg`**; clippy `-D warnings` clean; YAML valid. No production code touched.

## Remaining deferred slices

`iai-callgrind` deterministic instruction counts, and a `profiling` data branch for night-over-night trends.

### DIFF
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -70,24 +70,32 @@ jobs:
       - name: Run heap profile
         run: ./target/profiling/examples/profile_pipeline workload.beancount
 
+      - name: Build CPU-flamegraph harness
+        run: cargo build -p rustledger --profile profiling --example flamegraph_pipeline --features flamegraph
+
+      - name: Run CPU profile (flamegraph)
+        # In-process sampling (pprof) — no `perf` privileges needed. Loop the
+        # pipeline so the 1 kHz sampler collects enough samples.
+        run: ./target/profiling/examples/flamegraph_pipeline workload.beancount 50
+
       - name: Summarize
         run: |
           {
-            echo "## Nightly heap profile — load + process"
+            echo "## Nightly profile — load + process"
             echo ""
             echo "- Workload: \`$(wc -l < workload.beancount)\` lines"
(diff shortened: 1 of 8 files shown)


robcohen:
All three fixed: (1) iterations now read via `args_os` so a non-UTF8 ledger path can't panic the arg iterator; (2) the loop records `ledger.errors.len()` and prints a note so a flamegraph over an erroring workload isn't silently "clean"; (3) renamed the job `heap-profile` → `profile` / "Profile — heap + CPU" to match the now-dual output.

PR #1587, opened 2026-06-25
perf(parser): pre-size lexer token vectors to kill reallocation churn (-10.8% heap)

robcohen:
## What

**Profiling-driven optimization.** Pre-sizes the two lexer token vectors that the new nightly profiling (the `profiling` data branch) flagged as the largest allocations — eliminating reallocation churn.

`tokenize_inner` (`logos_lexer.rs`) and `lossless_kind_tokens` (`cst/lossless_tokens.rs`) both built their output `Vec` from `Vec::new()` and grew via `push`, reallocating ~17× as the geometric doubling kicked in. A beancount token averages ~4 bytes of source, so `Vec::with_capacity(source.len() / 4)` reserves close to the final count up front.

| `tokenize_inner` allocations | 20.97 MB in **17 blocks** | 10.25 MB in **1 block** | churn eliminated |
| **Total heap allocated** | 158.7 MB | **141.6 MB** | **−17.1 MB (−10.8%)** |
| Instructions | 1,121,094,109 | 1,118,784,650 | −2.3M |

(`lossless_kind_tokens` dropped out of the top-4 allocators entirely.)

## Why it's safe

Allocation-only change — no behavioral difference. `cargo test -p rustledger-parser` passes; clippy `-D warnings` clean. The win will show up as a step down in `heap_total_bytes` on the next nightly `profiling` trend point.

## How it was found

This is the payoff of the profiling work (slices 1–4): the dhat heap profile ranked these two `Vec::new()` sites #1 and #2 by bytes, and cachegrind confirmed ~12% of CPU was in `malloc`/`free`. More targets remain (the `Transaction::clone` ×2 per txn, the 190k-allocation `walk_descendants_once`, happy-path error-scan allocations) — follow-ups.

### DIFF
--- a/crates/rustledger-parser/src/cst/lossless_tokens.rs
+++ b/crates/rustledger-parser/src/cst/lossless_tokens.rs
@@ -26,8 +26,11 @@ const UTF8_BOM_LEN: usize = 3;
 #[must_use]
 pub fn lossless_kind_tokens(source: &str) -> Vec<(SyntaxKind, Range<usize>)> {
     // Pre-size (~4 bytes per token) to avoid reallocation churn — profiling
-    // flagged this as the 2nd-largest lexer allocation. See `tokenize_inner`.
-    let mut out: Vec<(SyntaxKind, Range<usize>)> = Vec::with_capacity(source.len() / 4);
+    // flagged this as the 2nd-largest lexer allocation. Capped (like
+    // `tokenize_inner`) so a pathological input can't over-reserve. See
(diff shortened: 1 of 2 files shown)


robcohen:
Good catch — capped both reservations at `TOKEN_CAPACITY_CAP` (4M entries, ~100MB; far above any real ledger). A pathological input (e.g. a giant whitespace run that lexes to few tokens) now grows normally past the cap instead of amplifying `len/4` into a multi-GB up-front reservation. Keeps the win for real files; respects the parser's graceful-on-malformed-input contract.

PR #1589, opened 2026-06-25
perf(booking): skip book() clone for cost-spec-free transactions (-1% instructions, -6MB)

robcohen:
## What

**Profiling-driven optimization ③.** Skips the booking step's verbatim clone for transactions with no cost specs (the common case).

`book()` has a documented fast path: a transaction with no cost specs needs no booking, so it returns `BookedTransaction { transaction: txn.clone(), gains: [], booked_indices: [] }`. But `book_and_interpolate` — the pipeline's entry point — **only consumes `booked.transaction`** (the `gains` / `booked_indices` are discarded), and in the fast path that transaction *equals* `txn`. So `interpolate(&book(txn).transaction)` is **provably identical** to `interpolate(txn)`. This PR interpolates the original directly when there are no cost specs, eliminating the clone:

```rust
if !txn.postings.iter().any(|p| p.cost.is_some()) {
    return Ok(interpolate(txn)?);   // skip book()'s verbatim clone
}
```

| **Allocations** | 1,223,484 | 1,213,484 | **−10,000** (one Transaction clone per txn) |
| Heap | 158.7 MB | 152.6 MB | **−6.08 MB** |
| **Instructions** | 1,121,094,109 | 1,109,315,922 | **−11.8M (−1.05%)** |

The `Transaction::clone` footprint halves (12 MB → 6 MB); the remaining clone is `interpolate` building its result, which the pipeline needs.

## Why it's safe

- **Provably equivalent**: the guard only diverts the no-cost-spec case, where `book()` is an identity that just clones `txn` — so interpolating `txn` vs its clone yields the same `InterpolationResult`. The cost-spec path is untouched.
- No signature / ownership / error-path changes (the `failed`-partition data is re-merged into output, so a `mem::take`-style ownership refactor would have been risky — this avoids all of that).

## Verification

- `cargo test -p rustledger-booking` (105) + `-p rustledger-loader` (89) pass; clippy `-D warnings` clean.
- Functional check on a cost-free ledger with elided postings: interpolation fills amounts correctly (`Assets:Checking 2874.50`, `Income:Salary -3000.00`), `rledger check` clean.
- **BQL compat suite**: runs as the CI gate on this PR (couldn't run locally — the oracle needs rootless podman + the external corpus). Per CLAUDE.md, booking changes get the compat gate + review.

## Found via

The nightly `profiling` data branch — third optimization it surfaced (after the lexer pre-size #1587 and the cost-brace guard #1588).


robcohen:
Good catch — removed. That was a generated artifact from running the compat harness locally; `git add -A` swept it in. The compat results live on the `compatibility` data branch via CI, never in main.

PR #1590, opened 2026-06-25
perf(parser): skip LSP-only occurrence collection on the load path (-23MB heap, -2.6% instructions)

robcohen:
## What

**Profiling-driven optimization ④ — the biggest yet.** Skips collecting the `currency_occurrences` / `account_occurrences` indices on the loader / CLI processing path, where they're never read.

`walk_descendants_once` was the **#1 allocation-count site** in the profile. For every `ACCOUNT` / `CURRENCY` token it builds a `Spanned<Account>` / `Spanned<Currency>` (an `Account::new` / `Currency::new` allocation) **and** does an O(depth) in-`ERROR_NODE` ancestor walk. But these two indices are consumed **solely by the LSP** (rename / references / highlight) — confirmed by grep: zero consumers outside `rustledger-parser` + `rustledger-lsp`. The loader / `rledger check|query` pipeline builds them and throws them away.

## Change

- New `parse_without_occurrences(source)` threads `collect_occurrences = false` into `walk_descendants_once`, which then skips the `ACCOUNT`/`CURRENCY` construction *and* their ancestor walks. Inline errors + top-level comments are still collected (the pipeline needs them).
- `parse()` is **unchanged** — the LSP and all 293 other call sites keep full behavior.
- The loader's 2 parse sites switch to the new variant.

| **Heap** | 158.7 MB | 135.1 MB | **−23.6 MB (−14.9%)** |
| **Allocations** | 1,223,484 | 1,183,412 | **−40,072** |
| **Instructions** | 1,121,094,109 | 1,092,010,959 | **−29.1M (−2.6%)** |

Largest heap *and* CPU win of the profiling series.

## Safety

- `parse()` semantics unchanged → LSP unaffected; the returned `ParseResult` differs from `parse()`'s only in the two now-empty occurrence vectors.
- Regression test (`parse_without_occurrences_skips_indices_but_matches_directives`) pins: occurrences empty, but `directives` / `errors` / `options` identical to `parse()`.
- `cargo test -p rustledger-parser` (280) + `-p rustledger-loader` (89) pass; parser+loader+lsp `cargo check` + clippy clean.
- Parser change → the BQL **compat suite runs as the CI gate** on this PR.

## Found via

The nightly `profiling` data branch — fourth optimization it surfaced.

### DIFF
--- a/crates/rustledger-loader/src/lib.rs
+++ b/crates/rustledger-loader/src/lib.rs
@@ -470,7 +470,9 @@ impl Loader {
             } else {
                 self.fs.read(path)?
             };
-            let parsed = rustledger_parser::parse(&src);
+            // The processing pipeline never reads currency/account occurrences
+            // (LSP-only); skip collecting them — see `parse_without_occurrences`.
+            let parsed = rustledger_parser::parse_without_occurrences(&src);
             (src, parsed)
         };
 
@@ -645,7 +647,8 @@ impl Loader {
                             // Read through the FileSystem trait so all I/O goes
                             // through one code path (UTF-8 handling, error types, etc.)
                             let source = fs.read(p).ok()?;
-                            let parsed = rustledger_parser::parse(&source);
+                            // Occurrences are LSP-only; skip them on the load path.
+                            let parsed = rustledger_parser::parse_without_occurrences(&source);
                             Some((source, parsed))
                         })
                         .collect();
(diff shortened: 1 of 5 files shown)


robcohen:
Both addressed: (1) the assertions now compare full `directives`/`errors`/`options` contents via `Debug`, not just lengths — a content regression that preserves counts now fails. (2) On naming: this file is ~50/50 between `test_*` and bare descriptive names (e.g. `parse_ok`, `string_escapes_are_decoded`), so I kept the descriptive form to match the latter group.

PR #1599, opened 2026-06-25
perf(cli): mimalloc global allocator for the rledger binary (-7% load)

robcohen:
## What

Set **`mimalloc`** as the global allocator for the `rledger` binary.

## Why

After the green-conversion campaign (#1594/#1595/#1596/#1598) cut the parser ~30%, a fresh cachegrind breakdown showed the **allocator machinery is now the #1 cost (~14%)** of the load — `_int_malloc` / `free` / `malloc_consolidate` / … The load → process pipeline makes thousands of small, short-lived allocations (green-tree build + AST conversion + booking); glibc malloc is a poor fit for that pattern.

## Result

Measured on the 10k-txn workload (cachegrind, same harness, only the allocator differs):

```
glibc malloc = 740,408,953
mimalloc     = 688,729,631   (-7.0%, -51.7M instructions)
```

cachegrind counts *instructions*; mimalloc's wall-clock win is typically larger (thread-local heaps, cache locality, less fragmentation). And it's **process-wide** — every command (check / query / format) benefits, not just parse.

## Design

- **Binary-only.** `#[global_allocator]` lives in `src/bin/rledger.rs`, never the library — embedders pick their own allocator.
- **Not on wasm.** Gated `cfg(not(target_arch = "wasm32"))` (mimalloc is a C lib).
- **Default-on, opt-out.** New default `mimalloc` feature; `--no-default-features` drops it (e.g. a build host without a C toolchain).

## Tradeoff

Adds `libmimalloc-sys` (a vendored C library) — build time, binary size, and a non-Rust dep for `cargo deny`. Pure-Rust allocators were evaluated and rejected: `ferroc` doesn't compile on rustc 1.96 (let-chains/edition) across all published versions, and `talc` v5 is an embedded/wasm-oriented rewrite with no std-heap drop-in. The root-cause pure-Rust fix (arena-allocating the conversion) is a larger follow-up.

## Verification

`rledger check` on the workload: `✓ No errors found` — output identical (allocator swap, no logic change).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2116,6 +2116,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.49"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a45a52f43e1c16f667ccfe4dd8c85b7f7c204fd5e3bf46c5b0db9a5c3c0b8e9"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


robcohen:
Good catch — fixed in 7083694. Moved `mimalloc` out of `default` features to an opt-in `mimalloc` feature, so it no longer pulls the vendored C dep into library consumers. Confirmed via `cargo tree`: 0 in the default tree, present only under `--features mimalloc`. Release binaries will enable it explicitly (a follow-up wires `--features mimalloc` into release-build.yml, alongside producing the ag-rledger artifact).

Issue #1291, opened 2026-06-09
[Feature]: Make output json and easier for agents


matthiasdebernardini:
### Problem Statement

I'm not a huge fan of MCP, would this project be interested in adding agent friendly output? 

I have a crate (agcli) that does just this, creating a new binary for myself in my fork calling it `agledger` and its a collection of best practices that I have found make tools much more efficient and effective. 

Let me know if you are interested or if you have any ideas for what the best version of this looks like.

### Proposed Solution

Use agcli in a new binary 

### Alternatives Considered

I'm not a huge fan of MCP, but if you know the reasons for why its better than an agent friendly CLI then I would switch to it. 

### Area

Other

### Beancount Compatibility

- [ ] This feature exists in Python Beancount
- [ ] This is a new feature not in Python Beancount

robcohen:
Yes, we'd like to bring this in. Your `agledger` binary on the fork looks like exactly the right shape: a separate agent-native CLI alongside `rledger`, leaving the main CLI untouched.

Licensing works out cleanly. `agledger` is GPL-3.0 like rustledger, and `agcli` (MIT) is fine as a dependency of a GPL project.

Would you be up for opening a PR to upstream it? You're the author and you made the `cmd::*` `run_with_writer` refactors, so it's cleanest coming from you (and you get the credit). A few things to align on:

- **Pin `agcli`**: the fork is on `0.10.2`. Want to target the latest `0.13.0` (the newer `--select` / `--compact` / `--quiet` plus the `doctor` self-audit), or keep `0.10.2` for now?
- **`tokio`** comes in via `agcli`, which is fine. We'll scope it to the `agledger` binary so the main `rledger` CLI stays synchronous.
- **Maintenance**: happy to have you as the contributor/owner of the `agledger` binary. We'll wire it into CI (build/test, plus cargo-deny/vet for `agcli`).

If you'd rather we bring it in ourselves, we can do that with attribution, but I think it's better as your PR. Let me know which you prefer.

robcohen:
Quick follow-up so this isn't blocked on a round-trip: we went ahead and drafted it on a branch, `feat/1291-agledger-spike`.

https://github.com/rustledger/rustledger/tree/feat/1291-agledger-spike

It's the same shape as your fork, adapted to current `main`:
- A separate `ag-rledger` binary (named to mirror `rledger`); the synchronous `rledger` CLI is untouched, `tokio` is scoped to the new binary.
[...]

PR #1600, opened 2026-06-25
ci(release): produce ag-rledger as a release artifact

robcohen:
## What

Builds the agent-native CLI (`ag-rledger`) in CI releases and includes it in the release archives, alongside `rledger` + `rledger-lsp`.

## Why

`ag-rledger` (the #1291 agent-native CLI, gated behind the `ag-rledger` feature → pulls `agcli`/`tokio`) was never built or shipped — release archives only contained `rledger` + `rledger-lsp`, so there was no way to obtain `ag-rledger` from a release.

## How

In `release-build.yml`, build `ag-rledger` in all three build paths (PGO Unix, PGO Windows, cross/native) exactly like `rledger-lsp` — separately, feature-gated, no PGO:

```
cargo build --release --target <T> -p rustledger --bin ag-rledger --no-default-features --features ag-rledger
```

and add it to the Unix `tar.gz` + Windows `zip` archives.

## Verification

- Build invocation verified locally: `cargo check -p rustledger --bin ag-rledger --no-default-features --features ag-rledger` is clean.
- `release-build.yml` is valid YAML.
- Non-breaking: the extra binary doesn't affect the Docker/AUR/binstall consumers, which extract `rledger` by name.
- The release pipeline runs only on tag push, so the full cross-compiled `ag-rledger` build is first exercised at the next release; this mirrors the proven `rledger-lsp` build/package pattern.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -213,6 +213,10 @@ jobs:
 
           # Build LSP separately (no PGO needed)
           cargo build --release --target ${{ matrix.target }} -p rustledger-lsp
+
+          # Build the agent-native CLI (ag-rledger) separately — feature-gated
+          # (pulls agcli/tokio), no PGO.
+          cargo build --release --target ${{ matrix.target }} -p rustledger --bin ag-rledger --no-default-features --features ag-rledger
       - name: Build with PGO (Windows)
         if: matrix.pgo && runner.os == 'Windows'
         shell: pwsh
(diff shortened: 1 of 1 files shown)


robcohen:
Thanks — but the agent-native binary *is* exercised in normal CI: `ci.yml` runs `cargo check/clippy --all-features --all-targets` and `cargo nextest run --all-features`, and `--all-features` enables the `ag-rledger` feature, which satisfies the bin's `required-features` — so the binary is compile-checked + clippy-linted and its `tests/ag_rledger_test.rs` integration test runs on every PR (the search only missed it because it matched `--all-features`, not the literal `--features ag-rledger`).

What is release-only is **cross-target** builds (musl / Windows-ARM / etc.) — but that is true of `rledger` and `rledger-lsp` too; none of the binaries are cross-compiled in PR CI. So this change keeps `ag-rledger` at exactly the same pre-release coverage as the existing binaries, rather than introducing a new gap. Adding cross-builds to PR CI would be a separate (and broad) change applying to all three binaries.

PR #1601, opened 2026-06-25
ci(release): enable mimalloc in release rledger binaries (-7% load)

robcohen:
## What

Enables the `mimalloc` global allocator in the release `rledger` binaries (the `-7%` load win from #1599), for native targets.

## Why

#1599 added `mimalloc` as an **opt-in** feature (correctly non-default, so it doesn't force a C dep on library consumers). But the release pipeline builds with `--no-default-features`, so the win wasn't reaching shipped binaries. This wires `--features mimalloc` into those builds.

## How

Append `--features mimalloc` to the `rledger` build commands in `release-build.yml`:
- PGO Unix + Windows (`cargo pgo build` / `cargo pgo optimize build`)
- native non-cross (`cargo build --profile …`)

## Cross targets intentionally excluded

mimalloc is a vendored **C** library; its cross build (musl / Windows-ARM / …) under `cross` isn't validated here, and a release shouldn't hinge on it. Cross-compiled targets ship **without** the allocator (no win, no risk); native targets get it. `rledger-lsp` / `ag-rledger` are separate binaries and unaffected.

## Verification

- Feature resolves at the virtual-workspace root without `-p`: `cargo build --no-default-features --features mimalloc` builds clean (tested locally).
- `release-build.yml` is valid YAML; mimalloc added to 4 native build lines, cross excluded.
- Output is unchanged — mimalloc is an allocator swap, no logic change (smoke-tested in #1599: `rledger check` identical).
- Release-only: first exercised at the next tag; the rledger+mimalloc native build itself was verified locally in #1599.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -256,16 +256,17 @@ jobs:
           fi
 
           if [ "${{ matrix.cross }}" = "true" ]; then
-            # NOTE: mimalloc (a vendored C lib) is intentionally NOT enabled for
-            # cross-compiled targets — its C build isn't validated under `cross`
-            # here, and a release shouldn't hinge on it. Cross targets ship
-            # without the allocator; native (PGO + non-cross) targets get it.
+            # Cross-compiled targets: NO mimalloc. mimalloc is a vendored C lib
+            # whose cross build isn't validated under `cross`, and a release
+            # shouldn't hinge on it — these targets ship without the allocator.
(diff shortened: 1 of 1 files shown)


robcohen:
Good point — tightened in the latest push. Reworded the cross-branch note to say plainly "Cross-compiled targets: NO mimalloc" and labeled the non-cross else branch "Native (non-cross) target: mimalloc enabled", dropping the ambiguous "PGO + non-cross" phrasing from inside the non-PGO step.

Issue #1240, opened 2026-05-28
Replace rust_decimal with arbitrary-precision decimal to close last Python compat divergence


robcohen:
## Background

Existing TODO in `CLAUDE.md` under "Known Limitations & TODOs":

> `rust_decimal` has a maximum precision of 28 digits, while Python's `decimal.Decimal` has arbitrary precision. This causes 1 compatibility test failure out of 694 (99.86% pass rate).
>
> **Affected file**: `beancount-lazy-plugins/tests_data_output_some_fund_output.beancount`
>
> Contains amounts with 28 decimal places (e.g., `0.7142857142857142857142857143`). Python detects a `2×10⁻²⁵ USD` residual imbalance. Rust considers it balanced due to precision limits.

This issue exists to give that TODO a tracked home for visibility and discussion.

## What "done" looks like

`rustledger-core` migrates from `rust_decimal::Decimal` to an arbitrary-precision decimal type. Likely options:

- **`bigdecimal`** — pure Rust, arbitrary precision, slower than rust_decimal in the common case.
- **`fastnum`** — newer, claims to be both arbitrary-precision and fast. Less proven.
- **Wrap `rust_decimal` + fall back to `bigdecimal` only when precision is exhausted** — complex; preserves the fast path but doubles the type surface.

After migration, the 1 standing Python divergence closes; the BQL compat suite hits 100% on `Check Compatibility` / `BQL Compatibility` / `Full AST Match` for the affected fixture.

## Why this is "known limitation, not active work"

CLAUDE.md says:

> **Practical impact**: None for real-world usage. No legitimate ledger has 28-decimal-place amounts.

The 1 affected fixture is a synthetic test in upstream lazy-plugins, not a real-world ledger. The refactor is significant — `Decimal` flows through every numeric path in core, parser, booking, query, and the wire format. The work is justified by the 100%-compat goal, not by any user-facing pain.

### 1. Performance regression budget

`bigdecimal` is materially slower than `rust_decimal` for the common case (heap-allocated arbitrary-precision arithmetic vs. stack-allocated fixed 128-bit). Need a benchmark target: e.g. "JOURNAL on the bench fixture (#1123) regresses by no more than 20%."

### 2. Wire-format implications

[...]

PR #1613, opened 2026-06-26
feat(core): exceed Beancount decimal precision via a zero-cost side channel

robcohen:
## Summary

Match-or-exceed Beancount's decimal precision **at zero hot-path cost**, without migrating off `rust_decimal`.

`Decimal` stays `rust_decimal::Decimal` (16 bytes, `Copy`, untouched). rust_decimal holds ~29 significant digits; Beancount (Python's arbitrary-precision `Decimal`) keeps more, so a literal beyond ~29 digits is rounded on parse and a balance residual that depends on those digits can read **zero** — the transaction looks balanced — where Beancount flags an imbalance.

The exact value of the (vanishingly rare) over-precise literal is stashed in a side table (`rustledger_core::decimal_exact`), consulted **only in cold paths**:
- **parse** records the exact `BigDecimal`, gated by a digit count so ≤29-digit literals (every real amount) cost nothing — no lock, no allocation;
- **balance residual** (`booking::to_big`) uses the exact value when present;
- **display** (`Amount`) shows the exact value;
- the **validator** escalates to the precise `BigDecimal` residual when the fast rust_decimal residual is non-zero **or** any over-precise literal exists (`any_overprecise()` is one relaxed atomic load).

## Why not change the number type

Three full migrations were prototyped and benchmarked (fastnum `D512`/`D256`/`D128`, and a `bigdecimal` hybrid). All regressed the hot paths **+22–35%** (e.g. validate +35%) because a hybrid enum taxes *every* operation of *every* ledger — `match` dispatch plus a larger `Decimal` inflating `Amount`/`Posting` — to serve a case that occurs in essentially none. The side channel inverts that: the 99.999% pay nothing.

## Result

| | this PR (side channel) | best hybrid |
|---|---|---|
| `validate` vs main | **~0%** (31.2µs vs 31.1µs) | +35% |
| precision ceiling | **unbounded** (BigDecimal) | fixed/unbounded |
| `Decimal` | rust_decimal, 16 B, unchanged | enum, 32–48 B |
| over-precise residual | ✅ flags `1e-30`, **matches Beancount** | ✅ |

Verified end-to-end: a transaction with `1.000000000000000000000000000001 USD` and `-1 USD` (a `1e-30` residual rust_decimal rounds to zero) is now flagged `Transaction does not balance: residual 1E-30 USD` — **byte-identical to Beancount**. The existing 28-digit `some_fund` compat fixture is unaffected (already handled by the two-tier validator).

## Soundness

- **Thread-safe** — the loader parses includes in parallel (rayon); the table is an `RwLock`, lookups take a shared read lock.
- **Collision-safe** — two distinct literals rounding to the same `Decimal` poison the key and both fall back to plain rust_decimal behaviour: a collision can only *lose* the precision bonus, never produce a wrong value.
- **Append-only** — over-precise literals are so rare that not auto-clearing is cheaper than risking a `clear` racing a concurrent load; `clear()` is exposed for explicit reset.
- **Cache-aware** — a parse-cache hit would bypass the side channel, so the loader doesn't write the on-disk cache for an over-precise load (those files re-parse every time).

## Tests

- `decimal_exact::side_channel_behaviour` — record/skip/preserve/collision-fallback.
- `loader_test::over_precise_literal_residual_detected_beyond_rust_decimal` — full parse→validate detecting the `1e-30` imbalance.

## Follow-ups (documented inline)

- Wire exact values through the aligned `display_context` formatter (this PR wires `Amount::Display`).
- Optionally cache the exact values (rkyv) instead of skipping the cache, if over-precise files ever become common.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -246,9 +246,9 @@ rledger aims for full Python beancount compatibility on **correct behavior**, bu
 
 Each Python bug we encounter falls into one of three buckets:
 
-1. **Fixable** — we deviate, stay stricter or more correct than Python. Examples: cost-spec precision (beanquery#275, fixed in #1106 / #1113), `FIRST` aggregator short-circuit (beanquery#279, we evaluate eagerly), empty-aggregate quirk (beanquery#1055, we return the SQL identity), elided-zero-to-unopened-account check (Python #877-equivalent, we catch via two-phase validation).
+1. **Fixable** — we deviate, stay stricter or more correct than Python. Examples: cost-spec precision (beanquery#275, fixed in #1106 / #1113), `FIRST` aggregator short-circuit (beanquery#279, we evaluate eagerly), empty-aggregate quirk (beanquery#1055, we return the SQL identity), elided-zero-to-unopened-account check (Python #877-equivalent, we catch via two-phase validation), and **decimal-precision residuals** (the `decimal_exact` side channel + two-tier balance validator — we meet *or exceed* Python without leaving `rust_decimal`; see Known Limitations).
 
-1. **Not fixable locally** — we match Python and document the limitation. Example: the `rust_decimal` 28-digit ceiling (would require migrating to an arbitrary-precision library like `bigdecimal` — significant refactor for negligible practical benefit).
+1. **Not fixable locally** — we match Python and document the limitation. (The long-standing example here, the `rust_decimal` 28-digit ceiling, graduated to "Fixable" above once the `decimal_exact` side channel landed in #1613 — no standing example at present.)
 
 1. **Out of scope** — we mask the Python-side divergence in the compat suite so it doesn't pollute the metric. Example: `KNOWN_PYTHON_DIVERGENCES` entries in `scripts/compat-bql-test.py`.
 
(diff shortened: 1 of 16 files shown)


robcohen:
## Deep review found the side channel unsound — fixed with identity-keying

A high-effort architecture/correctness review (manual + the multi-agent workflow, which agreed) found the original **value-keyed global** side table was unsound. Confirmed, with a reproducer:

```
over-precise  Assets:A 1.000…001 USD   (rounds to 1)  → SHOULD be flagged
plain         Assets:C 1 USD / Assets:D -1 USD         → genuinely balanced
```
rledger flagged **both** (Beancount flags **one**): the plain `1` absorbed the exact value `1.000…001` keyed under `1`. Using the feature corrupted the rest of the ledger. Related defects from the same two root causes (value-keying + global state): sign bug (negatives keyed positive), sub-1 literals keyed under `Decimal 0` (corrupting zero amounts), recording on *every* token (metadata/cost/price), never-reset global flag (cross-load contamination + permanent slow-path), and a stale-cache bypass.

### Fix: identity-keying
- `decimal_exact` is now **stateless** — it only detects over-precision and reconstitutes a value.
- The exact literal is carried on the **specific posting** that owns it, in posting metadata (`__exact_number__`, an internal key filtered from output) — so it can't leak, it rides through booking *and the rkyv cache* for free, and the validator escalates **per-transaction** (no global flag).
- Wired in **both** parser converters — `convert.rs` (red) and `green.rs` (green). The green path was why a txn placed directly after an `open` (no blank line) initially went undetected.
- `CACHE_VERSION` 12→13 so pre-fix caches re-parse.

### Verification
- Reproducer now flags **exactly one** imbalance; pinned by a regression test (`over_precise_literal_residual_detected_without_value_leak`) in a fixture mixing an over-precise txn with a plain balanced one.
- Positive / negative / green-path / red-path all detect; **3789 tests pass**.
- Hot path unchanged (`Decimal` is still `rust_decimal`); the per-txn escalation probe short-circuits on `all_zero` and skips empty-metadata postings.

### Remaining scope (documented, not a correctness gap)
Only over-precise **units** literals are carried; cost/price/metadata over-precision rounds like plain `rust_decimal` — sound (never wrong), just not extended to exceed Beancount there.

robcohen:
## Closing: the side channel isn't the right vehicle

This PR set out to recover over-precise (>28-digit) units literals that `rust_decimal` rounds away, so a balance residual that depends on those digits matches Beancount. Two deep architecture/correctness reviews later, the evidence is consistent: **even done correctly, the side channel doesn't deliver enough to justify its surface area.**

### What we learned
- **Review #1** found the original value-keyed global table was unsound — an over-precise literal's exact value leaked onto any unrelated amount equal to its rounded value, falsely flagging genuinely-balanced transactions. Fixed with **identity-keying** (per-posting metadata, stateless helper). The leak is gone and stays gone.
- **Review #2 (post-fix)** found the design still falls short, decisively:
  - **🔴 Tolerance gap (defeats the feature).** The escalated *exact* residual is still compared against a tolerance derived from the *rounded* scale. For any over-precise literal that rounds to a value with a multi-digit integer part, the tolerance is too coarse and masks the residual. `100.000…001 − 100` → exact residual `1e-30` → **rledger: balanced; Beancount: imbalanced.** The shipped fixture only passes because its scale-28 tolerance *underflows to 0* in rust_decimal. The side channel does nothing in most of the regime it exists for.
  - **🟠 Egress leak.** `__exact_number__` leaks into BQL `meta()` results and wasm JSON — putting internal data in the shared user-facing `meta` map means every egress needs a filter, and only the text formatter had one.
  - **🟠 Cost/price uncovered.** Residual is exact-aware only in the simple-units branch; cost/price postings and cost/price literals drop precision.
  - **🟡 Sign bug** in `overprecise_literal` (negative >28-char literals that fit are falsely escalated), plus doc/alloc/dup cleanups.

### Why shelve rather than keep fixing
Fully delivering this needs exact-aware **tolerance**, exact-aware **cost/price weights**, and `__`-filtering at **every** egress — substantial, fragile plumbing — for literals beyond ~28 digits that **do not occur in real ledgers**. Meanwhile `main`'s two-tier validator (`calculate_residual` → `calculate_residual_precise`) already catches every realistic (≤28–29 digit) over-precision case, including the original #1240 `some_fund` residual, at full `rust_decimal` hot-path speed.

The pragmatic call: **keep `Decimal = rust_decimal` + main's two-tier validator; do not merge the side channel.** Closing. The branch stays for reference; the identity-keying approach is the correct shape *if* unbounded-precision balance ever becomes a real requirement (it would still need the tolerance + cost/price + egress work above).

Issue #1239, opened 2026-05-28
miri CI job for transitive-dep UB detection on a curated test subset


robcohen:
## Background

`miri` is an interpreter for Rust's MIR that catches undefined behavior at test runtime — uninitialized memory reads, dangling pointers, data races (with `-Zmiri-tree-borrows`), invalid integer-to-pointer casts, etc.

The workspace has `#![forbid(unsafe_code)]` on every first-party crate, plus a grep-level invariant check (`scripts/check-unsafe-invariant.sh`). So *first-party* code can't introduce UB via `unsafe` blocks.

But the dependency graph contains plenty of `unsafe` — `rayon`, `wasmtime`, `parking_lot`, `tokio`, `serde_wasm_bindgen`, `rkyv`, etc. If a transitive dep has a soundness bug that our specific call pattern hits, we'd see crashes in production with no first-party `unsafe` to investigate.

`miri` running a curated subset of tests would catch UB hit through transitive `unsafe`.

## What "done" looks like

A CI job that runs `cargo +nightly miri test --workspace --exclude rustledger-wasm --exclude rustledger-ffi-wasi` on a small curated subset of fast tests. Excludes:

- WASM/FFI crates (miri doesn't support the targets they need)
- Slow tests (`#[ignore]` or `slow` feature flag)
- Tests using `wasmtime` directly (miri-incompatible)

Failures block merge for the curated subset; everything else stays the responsibility of the regular test suite.

## Why this is "honorable mention"

Most of `rustledger` doesn't exercise the kind of code paths where `unsafe` UB lurks — we're a parser, a query engine, and an LSP, not a kernel or a lock-free data structure. The probable yield from `miri` is one finding every several months, on a dep we'd update anyway.

But the *cost* is also low — one new CI job, minimal maintenance. The asymmetry (low yield, low cost) is worth it for a project that ships compiled artifacts and links to a lot of native code.

### 1. Which tests to run under miri?

Naïve `cargo miri test --workspace` is slow and many tests won't pass under miri's stricter checks (e.g. anything that reads the filesystem or calls into wasmtime). Need a curated `cfg(miri)` cfg-gating pass to mark miri-incompatible tests, similar to how some crates already gate on `cfg(target_arch = "wasm32")`.

### 2. Tree Borrows vs Stacked Borrows

[...]

PR #1618, opened 2026-06-28
ci(miri): run under Tree Borrows so rowan ThinArc is accepted

robcohen:
## What

The nightly **Miri** job has been red on **every run for months** with a Stacked Borrows violation:

```
error: Undefined Behavior: trying to retag from <…> for SharedReadWrite … tag
       does not exist in the borrow stack
   --> crates/rustledger-parser/src/cst/convert.rs:348   // SyntaxNode::green().into_owned()
```

`rustledger-parser` is `#![forbid(unsafe_code)]`, so the UB is **entirely inside `rowan`** — its green tree lives in a `ThinArc` whose internal `unsafe` pointer arithmetic is **valid under Tree Borrows but trips a Stacked Borrows false positive**. This is a well-known rowan pattern; rowan and rust-analyzer run Miri under Tree Borrows for exactly this reason.

The job has always run under default **Stacked Borrows** (no `-Zmiri-tree-borrows` in `MIRIFLAGS`). The intended Tree-Borrows migration (`chore/1239-miri-tree-borrows`) was opened but **never committed** (the branch has 0 commits ahead of main), so the job has failed since #1239 created it.

## Fix

Add `-Zmiri-tree-borrows` to `MIRIFLAGS`. Tree Borrows is the more permissive successor aliasing model and still catches real UB — it just accepts rowan's `ThinArc`.

## Verification

Verified via a `workflow_dispatch` run of **parser-only Miri under Tree Borrows** (the crate that hit the UB) — link in a comment below. (A full 4-crate Miri run is ~2.5h, dominated by `rustledger-core`'s proptests; the next scheduled nightly exercises the full set.)

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -60,13 +60,22 @@ jobs:
           cargo +nightly miri test --package rustledger-query --all-features
           echo "::endgroup::"
         env:
+          # -Zmiri-tree-borrows: use the Tree Borrows aliasing model instead of
+          # the default Stacked Borrows. `rowan` (our lossless-CST library) holds
+          # the green tree in a `ThinArc` whose internal `unsafe` pointer
+          # arithmetic is valid under Tree Borrows but trips a Stacked Borrows
+          # false positive — a "retag ... tag does not exist in the borrow stack"
+          # error at `SyntaxNode::green().into_owned()` (rustledger-parser is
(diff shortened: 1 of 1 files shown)


robcohen:
Superseded by the scope-to-rowan-free-crates approach: Tree Borrows does NOT fix this (verified — rowan ThinArc fails under both SB and TB, rowan#108 is an open upstream UCG question). Closing in favor of the new PR.

Issue #1164, opened 2026-05-20
Type-driven: make CostSpec invalid states unrepresentable


robcohen:
## Problem

`CostSpec` has two mutually-exclusive numeric fields modeled as independent `Option<Decimal>`:

```rust
pub struct CostSpec {
    pub number_per: Option<Decimal>,
    pub number_total: Option<Decimal>,
    // ...
}
```

Per the beancount spec, at most one should be set. The type allows the invalid state `Some(_) + Some(_)`. Today the invariant is enforced indirectly by validation and by the parser only producing one. Booking and downstream consumers carry "what if both are set?" branches that should not have to exist.

## Proposed change

Replace the two fields with an enum that makes the invalid state unrepresentable:

```rust
pub enum CostNumber {
    /// Cost per single unit: `{150.00 USD}`.
    PerUnit(Decimal),
    /// Total cost for all units: `{{ 1500.00 USD }}`.
    Total(Decimal),
    /// Cost spec with no numeric component (e.g., `{}` — booking-deferred lot match).
    Unspecified,
}

pub struct CostSpec {
    pub number: CostNumber,
    pub currency: Option<InternedStr>,
    pub date: Option<NaiveDate>,
    pub label: Option<String>,
    pub merge: bool,
}
```

## Wins

- Invalid state (both `_per` and `_total` set) becomes unrepresentable.
- Booking and validation code drops the "what if both?" defensive branches.
- Pattern matching on `CostNumber` is more explicit than nested `Option`-checking and forces the reader to consider all three cases.

## Costs

- Bounded refactor. `CostSpec` has ~17 construction sites across the workspace (counted during PR #1160 investigation):
  - parser/parser.rs (the main parser site)
  - core/format/mod.rs (8 sites — formatter)
  - booking/book.rs (4 sites)
  - booking/interpolate.rs (2 sites)
  - plugin/convert/from_wrapper.rs (1 site)
  - ffi-wasi/types/input.rs (1 site)
- All construction sites get a mechanical rewrite: pick the right `CostNumber` variant based on which field was previously set.
- All read sites (`if let Some(n) = cost.number_per`, etc.) get a `match cost.number { CostNumber::PerUnit(n) => ..., ... }` rewrite.

## Estimated scope

- Type definition: ~15 lines.
- Construction sites: ~17 mechanical updates.
- Read sites: probably 30–50 across booking, query, format, validate. All mechanical.

## Open questions

[...]

PR #1704, opened 2026-07-03
fix(parser,booking)!: compound cost {a # b} carries both components (N*a + b)

### LINKED ISSUE #1700: booking: per-unit-from-total cost ({a # b}) weight is wrong in all three forms
Found while writing marshalling tests in rustfava. Beancount semantics: `{a # b}` on N units weighs `N*a + b`. rustledger v0.19.0 (component build) gets all three forms wrong. With `Assets:B  2 VTI {COST}` against `Assets:Cash  -10.00 USD`:

| cost | expected weight | should balance −10? | rustledger says |
|---|---|---|---|
| `{5.00 USD}` | 10.00 | yes | ✅ balances |
| `{5.00 # 0.00 USD}` | 10.00 | yes | ❌ `residual -10.00` |
| `{# 10.00 USD}` | 10.00 | yes | ❌ `residual 10.00` |
| `{5.00 # 10.00 USD}` | 20.00 | **no** | ❌ balances silently |

So valid ledgers using `#` cost syntax error out, and *invalid* ones pass validation. The loaded posting does carry the correct `per_unit_from_total {per_unit: 5.00, total: 10.00}` cost shape — only the balance-weight computation is wrong.

Suggested differential-corpus addition: a `#`-cost fixture cross-checked against beancount (the existing corpus covers `@@` total *prices* but not `#` total *costs*, which is how this survived).

robcohen:
Fixes the last and most severe route-smoke finding, per the design discussed on #1700.

## The defect (all four table rows repro'd)
The parser folded `{a # b}` into `{{b}}` — dropping `a` on an explicitly wrong premise — and mis-read `{# b}` as `PerUnit{b}` (`{#` lexes as a fused `L_BRACE_HASH` the old detector missed). Valid compound ledgers errored, **invalid ones balanced silently**, and lot cost bases were wrong whenever `N·a + b ≠ b` (the issue's example coincidentally masked this).

## The fix (design option A)
New `CostNumber::Compound{per_unit, total}` carried **as written** (units may interpolate later). Booking folds it into the existing `Total→PerUnitFromTotal` conversion with `effective_total = N·a + b` — every post-booking consumer is form-agnostic. Both residual tiers weigh `N·a + b` through the shared generic arm.

## Reach (the #1164 exhaustive-enum design at work)
The compiler walked the change to **12 consumer sites across 8 crates** — parser red+green (the green/red parity corpus caught the green copy drifting on a fuzz input), booking, validate, implicit prices, dedup fingerprint (distinct hash tag), and five wire mirrors. **WIT 3.2.0 → 3.3.0** (additive; the case only crosses on pre-booking egress — rustfava regen is a follow-up).

## Tests
6-case integration suite (full issue table + `{6 # 10}` booking at 11.00/unit — the case the issue's example masked) · committed differential fixture (`tests/compatibility/files/costs/`, the corpus covered `@@` prices but never `#` costs) · two legacy tests that pinned the old contract flipped with history documented.

## ⚠️ Breaking
Ledgers that balanced under the `{a # b}` ≡ `{{b}}` misreading were genuinely unbalanced per beancount and now error — compat-policy bucket 1; release-notes callout needed.

Verified: 104 workspace suites green · clippy `-D warnings` · rustdoc `-D warnings` · fmt · fixture VALID via `rledger check`.

Closes #1700


robcohen:
Note on the advisory **SemVer (plugin-types)** red: it is correctly reporting the intentional API addition (`CostNumberData::Compound`). The enum is deliberately exhaustive — plugin authors get a compile error pointing at compound costs rather than silently mishandling them (the same mechanism that found all 12 in-repo consumer sites). Crate version bump rides the next release cut, as with prior wire-shape additions. The check is advisory by design (`ci.yml`: "deliberately NOT part of the build merge gate").

Issue #1026, opened 2026-05-07
validate: rledger accepts sale of non-existent inventory + unsolvable interpolation


robcohen:
## Problem

rledger's `check` accepts a file that bean-check correctly rejects with **"Too many missing numbers for currency group"** errors. The file sells positions that were never purchased, and includes missing-amount postings that can't be interpolated. rledger silently books negative inventories with no cost basis.

## Real-world impact

Niche but real. Triggering this requires the combination of `{}` empty cost spec + sale of a never-bought commodity + a missing-amount posting in the same currency group. End users rarely hand-write this. Importer-generated journals (like the htsec fixture this came from) do, and the silent acceptance can mask real ledger bugs.

## Minimal reproduction

```beancount
option "operating_currency" "CNY"

1970-01-01 open Assets:Htsec:Cash
1970-01-01 open Assets:Htsec:Positions
1970-01-01 open Expenses:Htsec:Commission
1970-01-01 open Income:Htsec:PnL

2022-01-12 * "sell what was never bought"
  Assets:Htsec:Positions  -13000.00 SH513050 {} @ 1.300 CNY
  Assets:Htsec:Cash        16900.00 CNY
  Assets:Htsec:Cash           -0.85 CNY
  Expenses:Htsec:Commission    0.85 CNY
  Income:Htsec:PnL
```

```bash
$ bean-check repro.beancount
repro.beancount:9: Too many missing numbers for currency group 'CNY'

$ rledger check repro.beancount
✓ No errors found

$ rledger query repro.beancount "BALANCES"
account                    balance
-------------------------  ------------------
Assets:Htsec:Cash          16899.15 CNY
Assets:Htsec:Positions     -13000.00 SH513050     ← negative position, no cost basis
Expenses:Htsec:Commission  0.85 CNY
Income:Htsec:PnL           (empty)                ← unsolvable interpolation silently zeroed
```

Full original fixture: `example/htsec/example-htsec-output.beancount` from `deb-sig/double-entry-generator` (downloaded by the compat harness). The fixture has 4 such transactions; bean-check rejects all 4 with the same error class.

## Root cause

**One bug. Three observable manifestations.**

The CNY currency group has TWO unknowns in this transaction:

1. The cost basis of `-13000.00 SH513050 {}` (the empty cost spec says "match any prior lot" but no prior buy exists, so the COST is unknown — and a position's contribution to the balance equation is `units × cost`, not `units × price`).
[...]

PR #1706, opened 2026-07-03
fix(booking): infer augmenting empty-{} cost from the residual

### LINKED ISSUE #1705: [Compat]: augmenting posting with empty cost spec {} is not interpolated (rejected as "2 unknowns")
## Description

An augmenting posting with an empty cost spec `{}` is accepted by `bean-check` but rejected by `rledger check`. Python Beancount infers the lot's per-unit cost from the transaction balance; rustledger reports it as `interpolation failed: ... (2 unknowns)`.

**Steps to reproduce**
1. Save the file below.
2. `bean-check file.bean` → passes (exit 0).
3. `rledger check file.bean` → fails with a `BOOK` error on the second transaction.

The file has two transactions: one augments a lot via empty `{}` (cost inferable from the cash leg), the other reduces it via `{}` alongside one explicit leg and a residual PnL posting.

_Possible cause (speculative):_ rustledger seems to count `{}` as an unknown *before* booking/lot-matching (cf. the "at most one unknown per currency" rule in `interpolate.rs`), whereas beancount books first and only then interpolates the single remaining unknown. Consistent with this, in isolation the augmenting lot ends up with no cost basis (see the `cost_number` query below).

## Beancount File

```beancount
option "operating_currency" "EUR"

2024-01-01 open Assets:Broker
2024-01-01 open Assets:Cash
2024-01-01 open Expenses:Misc
2024-01-01 open Income:Trading:PnL

; augmenting lot via empty {} -- per-unit cost should be inferred from the cash leg (0.90 EUR)
2024-01-02 * "buy USD"
    Assets:Broker    1000 USD {}
    Assets:Cash      -900 EUR

; later reduce that lot with {} + one explicit leg + PnL residual
2024-01-05 * "sell part"
    Assets:Broker    -100.00 USD {}
    Expenses:Misc     95 EUR
    Income:Trading:PnL
```

## Python Beancount Output

`bean-check` accepts the file (exit 0, no errors). The augmenting lot is booked at the inferred cost 0.90 EUR:

```
$ bean-check file.bean        # exit 0, no output

$ bean-query file.bean "SELECT account, number, cost_number, cost_currency WHERE account = 'Assets:Broker' AND date = 2024-01-02"
account        number  cost_number  cost_currency
-------------  ------  -----------  -------------
Assets:Broker    1000          0.9            EUR
```

## Rustledger Output

`rledger check` rejects the file:

```
$ rledger check file.bean
<unknown>: error[BOOK]: interpolation failed: multiple postings missing amounts or with unresolved cost specs for currency EUR (2 unknowns) (2024-01-05, "sell part")
✗ 1 error
```

The augmenting lot carries no cost basis (contrast with beancount's `cost_number = 0.9`):

```
$ rledger query file.bean "SELECT account, number, cost_number, cost_currency WHERE account = 'Assets:Broker' AND date = 2024-01-02"
BOOK: interpolation failed: multiple postings missing amounts or with unresolved cost specs for currency EUR (2 unknowns) (2024-01-05, "sell part")

account        number  cost_number  cost_currency
-------------  ------  -----------  -------------
Assets:Broker    1000
```

## Python Beancount Version

```
3.1.0
```

## Rustledger Version

```
0.19.1
```

yurenju:
An augmenting posting with an empty cost spec `{}` was accepted by `bean-check` but rejected by `rledger check` — rledger now infers the lot's per-unit cost from the transaction balance, matching beancount.

## Root cause
The pipeline **books first, then interpolates**, but `interpolate` only ever filled missing *units*, never a cost *number*. An augmenting `{}` was counted as one cost-unknown and then **never solved** — `1000 USD {}` + `-900 EUR` was accepted with **no cost basis** (the issue's empty `cost_number`), so a later `-100 USD {}` reducing that lot could no longer match a cost-bearing lot and became a spurious second unknown beside the elided PnL leg (the **`2 unknowns`** error). Fixing the augmentation fixes the reduction by cascade.

## The fix
`interpolate.rs` only: for an empty, **price-less** `{}` that is the **sole unknown in its cost currency group**, infer the per-unit cost from the residual (`total = -residual · signum(units)`, stored as `PerUnitFromTotal` to preserve precision). The structural rule (at most one unknown per currency) is **unchanged** — the unknown is now solved, not left empty. A priced `{}` still defers to booking-time lot matching (keeps #1026). Compat-policy **bucket 1** — **not** a breaking change.

Before: `error[BOOK]: interpolation failed: ... (2 unknowns)`. After: `✓ No errors found`, the augmenting lot books at **0.90 EUR** (matches `bean-query`) and PnL solves to **-5 EUR**. Regression tests: `test_interpolate_augmenting_empty_cost_inferred_from_residual` · `test_augmenting_empty_cost_then_reduce`. Green: booking / loader / validate / query / core · clippy `-D warnings` (booking) · fmt.

## Note
Full disclosure: I don't write Rust — this fix was authored entirely by Claude (Claude Code), and I'm relaying it. The one judgement call worth a close look is the **"no price" heuristic** used to tell an augmentation from a reduction at interpolation time. Happy to iterate on anything.

Closes #1705

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -1077,6 +1077,59 @@ mod tests {
         assert!(results[1].is_ok());
     }
 
+    /// Issue #1705: an augmenting `{}` lot gets its cost inferred from the
+    /// residual, and a later `{}` reduction of that lot then books cleanly
+    /// (previously the reduction hit a spurious "2 unknowns" error because
+    /// the augmenting lot carried no cost basis to match).
+    #[test]
+    fn test_augmenting_empty_cost_then_reduce() {
+        // buy: 1000 USD {} against -900 EUR  → lot booked at 0.90 EUR/unit
+        let buy = Transaction::new(date(2024, 1, 2), "buy USD")
+            .with_synthesized_posting(
+                Posting::new("Assets:Broker", Amount::new(dec!(1000), "USD"))
+                    .with_cost(CostSpec::empty()),
+            )
+            .with_synthesized_posting(Posting::new("Assets:Cash", Amount::new(dec!(-900), "EUR")));
+
+        // sell: -100 USD {} + 95 EUR + PnL residual  → reduce at 0.90, PnL = -5 EUR
+        let sell = Transaction::new(date(2024, 1, 5), "sell part")
+            .with_synthesized_posting(
+                Posting::new("Assets:Broker", Amount::new(dec!(-100.00), "USD"))
+                    .with_cost(CostSpec::empty()),
+            )
+            .with_synthesized_posting(Posting::new("Expenses:Misc", Amount::new(dec!(95), "EUR")))
+            .with_synthesized_posting(Posting::auto("Income:Trading:PnL"));
+
+        let results = book_transactions(&[buy, sell], BookingMethod::Fifo);
+        assert_eq!(results.len(), 2);
+
+        let buy_txn = &results[0].as_ref().expect("buy should book").transaction;
+        let buy_cost = buy_txn.postings[0].cost.as_ref().expect("cost present");
+        assert_eq!(
+            buy_cost
+                .number
+                .as_ref()
+                .and_then(rustledger_core::CostNumber::per_unit),
+            Some(dec!(0.90))
+        );
+        assert_eq!(buy_cost.currency.as_deref(), Some("EUR"));
+
+        // The `{}` reduction booked cleanly and the PnL residual solved to -5 EUR.
+        let sell_txn = &results[1].as_ref().expect("sell should book").transaction;
+        let pnl = sell_txn
+            .postings
+            .iter()
+            .find(|p| p.account.as_str() == "Income:Trading:PnL")
+            .expect("PnL posting present");
+        let pnl_amount = pnl
+            .units
+            .as_ref()
+            .and_then(|u| u.as_amount())
+            .expect("PnL filled");
+        assert_eq!(pnl_amount.number, dec!(-5));
+        assert_eq!(pnl_amount.currency.as_str(), "EUR");
+    }
+
     #[test]
     fn test_book_augmentation_not_reduction() {
         let mut engine = BookingEngine::new();
--- a/crates/rustledger-booking/src/interpolate.rs
+++ b/crates/rustledger-booking/src/interpolate.rs
@@ -4,7 +4,9 @@
 
 use rust_decimal::Decimal;
 use rust_decimal::prelude::Signed;
-use rustledger_core::{Amount, Currency, IncompleteAmount, Transaction};
+use rustledger_core::{
+    Amount, BookedCost, CostNumber, CostSpec, Currency, IncompleteAmount, Transaction,
+};
 use std::collections::HashMap;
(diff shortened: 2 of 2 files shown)


robcohen (changes requested):
Thanks — the root cause and the core mechanism here are exactly right, and validated: I ran this branch against a 17-fixture matrix cross-checked with real bean-check 3.2.3 (matrix + methodology in the #1705 plan comment). The residual solve (e01) and the NONE-method case both book `{0.90 EUR}` identical to beancount, and constructing `PerUnitFromTotal` via `BookedCost::new` directly is the right shape (exact total preserved for the BigDecimal residual tier).

Four things before this can land — one merge blocker, three semantic divergences the matrix caught:

## Blocker: #1708 (merged after this branch) makes the solver dead code
Main now has an interim guard in `book()` that errors on number-less augmenting cost specs — and the pipeline books before interpolating, so on a rebase **every path this PR fixes errors in the guard before your solver runs** (verified: all 11 augmentation fixtures die with `EmptyCostAugmentation` on the merge). The rebase should **remove the guard** (this PR supersedes it) — but see the next two points first, because two shapes the guard currently catches must keep erroring.

## 1. `posting.price.is_none()` skip diverges from beancount
Beancount solves from the residual **even when a price annotation is present** — including when they disagree:
- `1000 USD {} @ 0.90 EUR` vs `-900 EUR` → bean books `{0.90}`; this PR books **uncosted** (the exact silent corruption #1705 is about).
- `1000 USD {} @ 0.90 EUR` vs `-950 EUR` → bean books **`{0.95}` — residual wins, silently**; this PR books uncosted.

Suggest dropping the `price.is_none()` condition. (The price annotation still feeds implicit-price directives; it just never participates in cost solving — also verified: `{} @ price` + auto-posting is a 2-unknowns REJECT in bean, price never substitutes as a weight.)

## 2. Negative solved cost must reject
`1000 USD {}` against `+900 EUR` → bean-check: `Cost is negative` **error**; this PR books a `{-0.90 EUR}` lot. Needs an explicit rejection.

## 3. Ambiguous residual currency must reject
`1000 USD {}` + `-500 EUR` + `-100 GBP` → bean-check: `Failed to categorize posting` **error**; this PR picks a currency and books a wrong-valued lot. The solve should require the cost currency to be unambiguous (spec-given, or a unique candidate).

## Also
- Please add the matrix fixtures as committed differential corpus (`tests/compatibility/files/interpolation/`, same pattern as `costs/`) — happy to hand you the 17 files; they encode all of the above as regression tests including the issue's original file end-to-end.
- The two rejects (negative, ambiguous) can reuse or adapt the #1708 error path rather than inventing new wording.

With those four addressed this is exactly the fix from the validated plan — looking forward to landing it.

robcohen:
Fixtures ready to cherry-pick: branch [`chore/1705-edge-fixtures`](https://github.com/rustledger/rustledger/tree/chore/1705-edge-fixtures) — one commit adding all 17 as `tests/compatibility/files/interpolation/` (committed differential corpus, gitignore exception included, bean-check 3.2.3 cross-checked). `git cherry-pick 14d12283e534909e89cab8669fd019a6ea95bedd` onto your branch and the compat suite covers every case from the review.

PR #1711, opened 2026-07-03
fix(query): typed scalar functions propagate NULL (beanquery parity)

### LINKED ISSUE #1699: query: only() type-errors on NULL first argument (beanquery propagates NULL) — Holdings by_currency 500s
Found by rustfava's new route-smoke suite: the frontend's Holdings **by_currency** query fails compilation on any ledger where a group has no cost currency:

```
Query error: type error: ONLY: first argument must be a currency string
```

Repro (any ledger with a costless commodity group, e.g. rustfava's `example.beancount`):

```sql
SELECT
  safediv(number(only(first(cost_currency), cost(sum(position)))),
          number(only(first(currency), units(sum(position))))) as average_cost
WHERE account_sortkey(account) ~ "^[01]"
GROUP BY currency, cost_currency
```

`first(cost_currency)` is NULL for groups without costs; beanquery's `only()` propagates NULL in that case, rustledger raises a type error. **User impact: the Holdings → by currency page in rustfava 500s** for such ledgers (the standard example ledger included) — same user-facing class as rustfava#190.

Expected: `only(NULL, inventory)` → NULL (beanquery parity). Likely applies to other typed scalar functions receiving NULL from `first()`/aggregates — worth a NULL-propagation sweep in the function type-checker, and a differential-corpus case.

robcohen:
Closes #1699 (the remaining half — `only()` was fixed in #1702).

Scalar functions applied to NULL raised a type error instead of propagating (SQL/beanquery semantics), so expressions like `abs(sum(number(value(position))))` — which rustfava's Holdings report runs on every render — failed compilation whenever the inner aggregate yielded NULL for a group (rustfava keeps this excused via a self-policing gap list whose guard test will demand re-arming once this ships).

**NULL now propagates through:** `YEAR MONTH DAY QUARTER UPPER LOWER TRIM ABS LENGTH NEG ACCOUNT_SORTKEY PARENT LEAF ROOT ROUND`, and the **second argument** of `FILTER_CURRENCY` / `POSSIGN`. `EMPTY` keeps its deliberate `NULL → true` semantic. The already-correct ones (`STR/INT/DECIMAL/BOOL` casts, `NUMBER CURRENCY UNITS COST CONVERT SAFEDIV COALESCE`) are untouched.

New `scalar_null_propagation_test.rs` locks the entire scalar surface, including the motivating Holdings shape and second-arg propagation. Full `rustledger-query` suite green (401 unit + all integration tests), clippy clean, fmt clean.

### DIFF
--- a/crates/rustledger-booking/src/pad.rs
+++ b/crates/rustledger-booking/src/pad.rs
@@ -319,13 +319,16 @@ fn create_padding_transaction(
 /// re-applies them against an inventory that already includes the prior
 /// synth. A `debug_assert!` guards against this in dev builds.
 pub fn merge_with_padding(directives: &[Directive]) -> Vec<Directive> {
-    debug_assert!(
-        !directives
-            .iter()
-            .any(|d| matches!(d, Directive::Transaction(t) if is_synthesized_pad(t))),
-        "merge_with_padding called on input that already contains synth pad transactions; \
-         re-running would double-count pad effects",
-    );
+    // Idempotence: input that already contains synth pad transactions has
+    // been merged before (e.g. an embedder queries entries it loaded via
+    // load-full, which merges pads — rustledger#1712). Re-running would
+    // double-count pad effects, so return the input unchanged instead.
+    if directives
+        .iter()
+        .any(|d| matches!(d, Directive::Transaction(t) if is_synthesized_pad(t)))
+    {
+        return directives.to_vec();
+    }
 
     let result = process_pads(directives);
 
@@ -704,13 +707,11 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "merge_with_padding called on input that already contains synth")]
-    fn test_merge_with_padding_double_apply_debug_asserts() {
-        // Calling merge_with_padding twice would double-count pad
-        // effects (original Pads survive in the output and would be
-        // re-applied against an inventory that already includes the
-        // prior synth). A debug_assert in dev builds guards against
-        // this caller mistake.
+    fn test_merge_with_padding_is_idempotent() {
+        // Re-merging already-merged input must be a no-op, not a
+        // double-count (and not an abort): embedders legitimately feed
+        // load-full output — which is already merged — back into
+        // query/window ops (rustledger#1712).
         let directives = vec![
             Directive::Open(Open::new(date(2024, 1, 1), "Assets:Bank")),
             Directive::Open(Open::new(date(2024, 1, 1), "Equity:Opening")),
@@ -722,7 +723,8 @@ mod tests {
             )),
         ];
         let merged_once = merge_with_padding(&directives);
-        let _merged_twice = merge_with_padding(&merged_once); // should panic
+        let merged_twice = merge_with_padding(&merged_once);
+        assert_eq!(merged_once.len(), merged_twice.len());
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


robcohen:
Downstream job status — the peeling is complete, three rounds:

1. **`abs(NULL)`** (this PR's original scope) — fixed, which let the Holdings query *execute* for the first time…
2. …reaching a **pre-existing pad re-merge abort** in `query_entries` (#1712): debug components trapped (poisoning the instance → the 40-failure cascade), release components silently risked double-counting. `merge_with_padding` is now idempotent.
3. …which let the query reach its expression tail, `safediv(…) * 100` → **`NULL * 100` arithmetic type error**. Arithmetic operators now propagate NULL.

Final downstream verdict: **658 passed, 1 failed — and the 1 is rustfava's gap-list guard demanding its `abs` marker removal** because the gap is verifiably fixed. That is the by-design advisory red for a contract-improving PR (per `downstream.yml`'s own docs); the marker deletion happens in rustfava when this ships in a release.

robcohen (commented):
Reviewed with bean-check 3.2.3 ground truth + full local verification. **The sweep is correct and complete** — this is the follow-up the #1699 analysis asked for, and the operator-level propagation closes the arithmetic gap too.

## Verified
- **Parity direction confirmed empirically**: `bean-query` returns blank/NULL for `upper(payee)` on a payee-less transaction; the 16 new `Value::Null` arms + the binary-operator arm match that semantic.
- **Both args covered** on the two-arg functions: first-arg NULL on `FILTER_CURRENCY` already propagates (mod.rs ~1163) and the PR adds the second arg — checked both, no gap.
- **pad.rs idempotence guard**: verified reachable via #1712's `query_entries`-on-`load-full`-output path (nice find chain — this PR's fix un-blocked the Holdings query that exposed it). Strictly better than the old behavior (debug abort / release double-count).
- Suites: `scalar_null_propagation_test` 4/4 · query+booking 18 suites green.

## One CI-blocking nit
`scalar_null_propagation_test.rs:82` — clippy `doc-markdown` (missing backticks) fails `-D warnings`; the quality gate will red until fixed.

## Two non-blocking notes
1. **pad.rs partial-merge edge**: the guard skips ALL processing when any synth pad exists — if an embedder ever appends a NEW raw `pad` directive onto already-merged entries, that new pad is silently unprocessed. Not reachable through today's `load-full` path (which merges everything), and the old code crashed on the same input, so no blocker — but a `debug_assert` for the mixed state (synth pads present AND a pad directive with no synth twin) would keep it honest, or a doc line on `merge_with_padding`.
2. **`EMPTY`'s NULL→true exception is bean-UNVERIFIABLE, not bean-aligned**: beanquery *type-rejects* `empty(upper(payee))` outright (`no function matches ... name and argument types`) — rustledger's looser typing accepts the expression, so there is no Python behavior to match. `true` ("a missing inventory IS empty") is defensible, but per the compat-policy checklist this is a deliberate divergence worth an inline comment naming it as such, rather than "pre-existing semantic".

robcohen:
Two non-blocking reds explained: the **downstream-rustfava** failure is the by-design gap-list guard (see thread above), and the **fuzz_green_eq_red** failure is a pre-existing parser green/red divergence that reproduces on untouched main — filed with repro as #1713 (this PR touches no parser code; the fuzzer's stochastic exploration just found it here first). Mutation testing is the only job still running.

robcohen:
Addressed all four review threads in 0e852e75: date/numeric blocks now assert via the `NULL` literal directly (with `round(NULL)` added — it was in the propagation list but untested), `empty(NULL)` uses the literal, and the doc now notes `payee` is a transaction-level field the posting rows inherit. The string block deliberately keeps `payee` so one path still exercises the real column pipeline.

PR #1714, opened 2026-07-04
fix(parser): green compound-cost conversion retries unparsable numbers like red

### LINKED ISSUE #1713: parser: green/red divergence found by fuzz_green_eq_red (pre-existing; reproduces on main)
`fuzz_green_eq_red` found an input where the green-wired `parse` and `parse_red_only` produce different directives. **Reproduces on untouched main** (verified) — it surfaced on #1711's fuzz run only because libFuzzer exploration is stochastic; it is unrelated to that PR's query/booking changes.

Repro (418 bytes, base64):

```
yJ1EAEJCSgVCQkJsOkREOgoxNjI0LzYvMiMPCiBOOk7grr8JAAFBAHt0R3Q6XUd0e3p7e2JhbDpsAHtBNTpERDoKMTYyNC82LzIjDwogTjpO4K6/CQABQQB7dEd0Ol1HdHt6e3tiYWw6bAB7QTU2NjZRLkvbtnR7LzIjDwogTjpO4K6/CQABQQB7dEd0Ol1HdHt6e3tiOmxhbABmLzExMiMjIyMjIx0LIzEyMSMjMTEnACMjMR0LIzExMiMjIyMdCyMxMTIfIzExMSMdCyMxJwAjIzEde0E1NjY2US50ezY2NlEudHtINjY2US50e0hKSkritbURSVEqLTJBNQsjMTEyIyMjHQsjMTEyHyMxMTEjIzExMh8jMTY2NjFRMS4xdHt6e3tiYWw6bAB7QTU2NjZRLnR7LzIjDwogTjpO4K6/CQABQQB7dEd0Ol1HdHt6e3tiOmxhbABmLzExMiMjIyMjIx0LIzEyMSMjMTEnACMjMR0LIzExMiMjIyMdCyMxMTIfIzExMSMdCyMxJwAjIzEde0E1NjY2US50ezY2NlEudA==
```

```rust
let data = base64_decode(...);
let src = std::str::from_utf8(&data).unwrap();
assert_eq!(
    format!("{:?}", rustledger_parser::parse(src).directives),
    format!("{:?}", rustledger_parser::cst::parse_red_only(src).directives),
); // fails
```

The input is mostly binary garbage around beancount-ish fragments (`bal:`, dates like `1624/6/2`, `{...}` cost braces, NUL bytes, control chars) — likely an error-recovery path where the green converter returns `Some` without exactly replicating red, violating the target's contract ("returns Some only when it exactly replicates red, and bails to red otherwise").

Worth minimizing with `cargo fuzz tmin` and adding to the corpus once fixed.

robcohen:
Closes #1713 (the `fuzz_green_eq_red` divergence found on #1711's fuzz run).

**Root cause — an asymmetry introduced in #1704**: red got two number semantics (`cost_compound_numbers` retries past unparsable NUMBER tokens on each side of the `#`; the plain path latches the first token via `cs.number()`), but green kept a single latched tracker satisfying only the plain path. On garbage like `{<unparsable> 2 # ...}`, red retried to `per_unit: 2` while green latched `None → 0`.

**Fix**: green carries both — a retried pre-hash tracker for the compound path alongside the latched first-token tracker for the plain path; the post-hash side switches from latch to retry, matching red.

**Verification**
- The 418-byte fuzzer input is pinned as a regression test (green==red asserted).
- `CACHE_VERSION` 13→14 + changelog (parse output changes for garbage-adjacent compound costs; stale caches would resurrect the divergence). Tripwire fixture versions advance; rkyv layouts unchanged.
- **Fuzz-validated**: 5.34M executions / 9 min on the fixed converter — zero artifacts.
- Workspace: clippy `-D warnings`, fmt, doc, parser + loader suites all clean.

Drive-bys: pre-existing typos in cache.rs comments (hook-enforced); sample_stub fixture lockfiles regenerated for 0.20.1 (missed by the release bump; their staleness broke prek pre-push stash handling).


robcohen:
Good catch — `cost_total_after_hash` died in #1704. Fixed both references in 872d4ebc; the parity-corpus comment also narrated the old latch semantics, so it now describes the retry-on-both-sides behavior it actually pins (`Compound{7, 0}` agreement).

PR #1726, opened 2026-07-08
Correct balances and balsheet reports for reductions in holdings

weslem:
## Summary

The `balances` and `balsheet` output for `rledger report` would include both positive and negative positions for held commodities. Instead, they should produce the total sum of the currency amounts in a single non-zero value per account, per commodity.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Feature (non-breaking change that adds functionality)

## Changes

- Add only the amount from each included position, rather than the cost. This way `Inventory.add()` will keep a correct sum of the commodity balance.

## Testing

- [x] `cargo test` passes
- [x] `cargo clippy` passes
- [x] Tested manually with a beancount file
- [ ] Added new tests for the changes

## Checklist

- [ ] I have added tests that prove my fix/feature works
- [ ] I have updated documentation if needed

## Review Focus

- [ ] Please review test coverage
- [x] General review is fine

## Related Issues

### For Reviewers

### DIFF
--- a/crates/rustledger/src/cmd/report_cmd/balances.rs
+++ b/crates/rustledger/src/cmd/report_cmd/balances.rs
@@ -25,15 +25,7 @@ pub(super) fn report_balances<W: Write>(
                 for posting in &txn.postings {
                     if let Some(amount) = posting.amount() {
                         let inv = balances.entry(posting.account.clone()).or_default();
-                        let position = if let Some(cost_spec) = &posting.cost {
-                            if let Some(cost) = cost_spec.resolve(amount.number, txn.date) {
-                                rustledger_core::Position::with_cost(amount.clone(), cost)
-                            } else {
-                                rustledger_core::Position::simple(amount.clone())
-                            }
-                        } else {
-                            rustledger_core::Position::simple(amount.clone())
-                        };
+                        let position = rustledger_core::Position::simple(amount.clone());
                         inv.add(position);
                     }
                 }
--- a/crates/rustledger/src/cmd/report_cmd/balsheet.rs
+++ b/crates/rustledger/src/cmd/report_cmd/balsheet.rs
@@ -33,15 +33,7 @@ pub(super) fn report_balsheet<W: Write>(
                     };
 
                     let inv = balances.entry(posting.account.clone()).or_default();
-                    let position = if let Some(cost_spec) = &posting.cost {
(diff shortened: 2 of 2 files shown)


robcohen (commented):
Verified against bean-check 3.2.3 ground truth — **the fix is correct, and I'm going to merge it.** Thanks for the clean catch, @weslem.

## Confirmed
- **Bug reproduces on main**: `report balances` on a buy(10 AAPL {150})/sell(-5 AAPL {150} @ 180) fixture prints two rows, `10 AAPL` and `-5 AAPL` — the reduction never nets. With your patch it prints `5 AAPL`, matching beancount's `sum(position)` = `5 AAPL {150.00 USD}`.
- **Root cause** is exactly what your patch implies: the reduction posting resolved to a different lot-key than the augmentation (the `@ price`/`resolve` path yields a position that doesn't share the buy's cost key), so `Inventory.add` couldn't cancel them. Dropping cost lets them net by commodity.
- **This reduces drift rather than adding it**: `income.rs` — the third report consumer with the same pattern — *already* uses `Position::simple`. Your change makes `balances.rs`/`balsheet.rs` consistent with it.
- **No valuation regression**: balsheet's `sum_by_currency`/net-worth only read `units.number`+`currency`; the cost was only ever a lot-key and was never rendered, so nothing downstream loses information.

## I'll take it from here
Your fix is good and I don't want to make you chase the extra bits — **I'll merge this and add the two follow-ups myself** in a small commit on top:
1. **Regression tests** — one per report asserting the buy/sell fixture yields a single `5 AAPL` row, so this can't silently rot in a future refactor.
2. **A one-line comment** noting these are deliberately simplified, cost-less views: two lots at *different* costs now merge (`buy 10 AAPL {150}` + `buy 10 AAPL {200}` → `20 AAPL`), where beancount keeps them as separate lots. That's arguably better than the old output (two indistinguishable `10 AAPL` rows), but the next reader should know these reports sum by commodity and don't track lots.

For the record, the long-term direction (not this PR): these three report consumers each re-derive an inventory from raw postings; the cost-accurate fix is to have them read the already-booked inventory instead. Larger change, tracked separately.

Nice, tight fix. Merging shortly with your authorship intact. 👍

Thank you for taking the time to submit this fix.

PR #1728, opened 2026-07-09
feat(report): balances/balsheet show lots at cost, like beancount

robcohen:
#1727 made `account_balances` the single source of truth but realized **cost-less** (a held commodity showed `5 AAPL`). This realizes through the **booking engine** so reports match beancount: `5 AAPL {150.00 USD}`.

## What changed
- `account_balances` drives a `BookingEngine` over the already-booked directive stream and returns its realized per-account inventories — the *same* lot-matching the loader's book phase uses, not a re-implementation. Reductions net against lots by the account's booking method (FIFO/LIFO/…).
- **Supersedes #1727's trade-off**: different-cost lots no longer merge — `10 AAPL {150}` + `10 AAPL {200}` show separately, like beancount.
- `rustledger-booking`: new `BookingEngine::into_inventories()` exposes the realized map.
- balances/balsheet render lot cost in text/CSV/JSON (new `cost` column/field; empty for plain currency → cash/income rows unchanged).

## Verified vs beancount 3.2.3
- reduction (buy 10 / sell 5) → `5 AAPL { 150.00 USD, 2024-01-05}`, matching `sum(position)` (rledger also shows the lot date like beancount's `Position` str)
- two lots → shown separately
- report + CLI suites pass, clippy 0, fmt clean
- the one failing integration test (`test_account_lifecycle_consistency`) is **pre-existing on main**, unrelated (bean-check validation)

`holdings` (own cost-basis path) and `income` (no cost) are unaffected.

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -156,6 +156,19 @@ impl BookingEngine {
         self.inventories.get(account)
     }
 
+    /// Consume the engine and return its realized per-account inventories.
+    ///
+    /// After [`apply`](Self::apply)ing a booked directive stream, this is the
+    /// canonical realization — each account's `Inventory` carries its lots at
+    /// cost, with reductions already matched (FIFO/LIFO/etc. per the account's
+    /// booking method). Callers that render balances (reports, holdings) read
+    /// this rather than re-summing postings, so cost is retained and lot
+    /// matching is the engine's, not a re-implementation.
+    #[must_use]
+    pub fn into_inventories(self) -> FxHashMap<rustledger_core::Account, Inventory> {
+        self.inventories
+    }
+
     /// Book a transaction: fill in empty cost specs and calculate gains.
     ///
     /// This does NOT modify the internal inventories - call `apply` for that.
--- a/crates/rustledger/src/cmd/report_cmd/balances.rs
+++ b/crates/rustledger/src/cmd/report_cmd/balances.rs
@@ -17,8 +17,9 @@ pub(super) fn report_balances<W: Write>(
     // `super::account_balances`); no report re-derives them itself.
     let balances = super::account_balances(directives);
 
-    // Collect data for output
-    let mut rows: Vec<(&str, Decimal, &str)> = Vec::new();
+    // Collect data for output. `cost` is the beancount-style lot annotation
+    // (e.g. ` {150.00 USD}`) for held commodities, empty for plain currency.
+    let mut rows: Vec<(&str, Decimal, &str, String)> = Vec::new();
     for (account, inventory) in &balances {
         if let Some(filter) = account_filter
             && !account.starts_with(filter)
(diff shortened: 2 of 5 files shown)


robcohen:
## Deep self-review

Ran an adversarial pass on this change — interrogating the booking-engine realization on edge cases and the risky contract assumptions, not just the happy path. **No blockers found**; a handful of notes worth recording.

### Verified correct against beancount 3.2.3
| Case | rledger | beancount | ✓ |
|---|---|---|---|
| Reduction (buy 10 / sell 5) | `5 AAPL {150.00 USD}` | `5 AAPL {150.00 USD}` | ✓ |
| Two different-cost lots | shown separately | separately | ✓ |
| **Same-cost lots** (buy 10 + buy 5 @ same price) | **two lots by date** | **two lots** | ✓ |
| Price-only (`@`, no cost) | `100 EUR` cost-less | `100 EUR` | ✓ |
| Full liquidation | account omitted | shown empty | see note 2 |

The same-cost case is the notable one: beancount keeps lots distinct by *date* even at identical cost, and the engine realization matches — so this is more faithful than the PR title's "different-cost" framing implied.

### Contract / safety
- **`apply` requires booked input** — verified `balance_input` derives from `ledger.directives`, which is booked (costs resolved). `booking` is a **default feature** (`full`), and the CLI takes the loader with defaults, so directives are always booked in any normal build. The unbooked-over-sell path needs `--no-default-features`, which the CLI never does. Not reachable (note 4).
- **Determinism**: `FxHashMap → BTreeMap` collect sorts by account; output is stable.
- **CSV injection**: the cost string contains a comma (`{ 150.00 USD, 2024-01-05}`); `csv_escape` correctly quotes it. JSON escapes cleanly. Verified both.
- Tests: report_pad_expansion (8), reduction_netting (3), cli_commands (57) pass; clippy 0; fmt clean. The one failing integration test is pre-existing on main (bean-check validation).

### Notes (non-blocking)
1. **Lot date is shown** (`{ 150.00 USD, 2024-01-05}`) where bean-query's `sum(position)` omits it. Deliberate — matches beancount's canonical `Position` str and fava. More informative; flagging as a conscious choice.
2. **Full liquidation**: rledger omits zero-balance accounts (pre-existing balances behavior of skipping empties), beancount shows them empty. Not a regression from this PR.
3. **Perf**: reports now run a second inventory-accumulation pass (`engine.apply` over already-booked directives) on top of the loader's book phase. Fine for one-shot report commands; the way to avoid it later is the deferred consolidation (loader exposes its realized inventories, reports/BQL/FFI read them).
4. **Theoretical unbooked build**: `account_balances` unconditionally realizes via `BookingEngine`; a `--no-default-features` loader build (no `booking`) would feed it unbooked directives. Unreachable via the CLI, but a defensive doc note on the fn wouldn't hurt.
5. **Schema change**: CSV/JSON gain a trailing `cost` column/field. Additive, but positional CSV consumers should know.

Verdict: correct, beancount-faithful, contained. Notes 1 and 5 are the only user-facing decisions (lot-date display, added column) — both intentional.

PR #1745, opened 2026-07-10
chore: fix Rust 1.97 toolchain breakage (clippy lints + BQL parse stack)

robcohen:
## What

Fixes the ten `useless-borrows-in-formatting` violations that stable Rust **1.97.0** (released 2026-07-07) introduced as a new clippy lint — redundant `&` on `format!`/`panic!` arguments in `rustledger-core/src/format/transaction.rs` (5) and `rustledger-query/tests/bql_integration_test.rs` (5). Mechanical; no behavior change.

## Why

CI's `dtolnay/rust-toolchain@stable` picked up 1.97.0 overnight and **every open PR went red** (Clippy/build/Test/quality) on code none of them touched. This PR clears main so the merge queue (#1731–#1744) can proceed; each PR branch then gets a base update to pick this up.

Verified with a local 1.97.0 sweep: `cargo clippy --all-features --all-targets -- -D warnings` is clean workspace-wide after this change, `cargo fmt --check` unaffected, touched crates' tests green.

**Merge this first**, before anything else in the queue.

### DIFF
--- a/crates/rustledger-core/src/format/transaction.rs
+++ b/crates/rustledger-core/src/format/transaction.rs
@@ -34,18 +34,18 @@ pub fn format_transaction_lines(txn: &Transaction, config: &FormatConfig) -> Vec
     metadata_lines(&txn.meta, &config.indent, &mut lines);
 
     // Posting-level metadata is indented one level deeper than the posting.
-    let meta_indent = format!("{}{}", &config.indent, &config.indent);
+    let meta_indent = format!("{}{}", config.indent, config.indent);
 
     for posting in &txn.postings {
         // Comments that appear before this posting.
         for comment in &posting.comments {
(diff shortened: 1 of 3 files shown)


robcohen:
Scope grew while landing: 1.97 also broke the **Test** and **Code Coverage** jobs — the chumsky grammar's debug frames grew enough that `moderate_nesting_still_parses` SIGABRT'd a 2 MiB cargo-test thread (1.96 handled 100+ levels on the same stack). Second commit runs the recursive BQL parse on a dedicated 16 MiB scoped thread on native targets, so the depth budget is owned by the parser instead of inherited from whichever thread calls it; nesting cap unchanged at 128, wasm32 path unchanged. Verified green on both 1.96 (nix) and 1.97 (rustup) locally.

PR #1795, opened 2026-07-14
feat(config): add RLEDGER_CONFIG_DIR and deprecate RLEDGER_CONFIG.

idheepan:
## Summary

Adds RLEDGER_CONFIG_DIR, an env var that relocates the entire user config directory (both config.toml and importers.toml) to an arbitrary path, replacing the narrower RLEDGER_CONFIG (which only relocated config.toml and never affected importers.toml discovery). This closes the gap where a dotfile manager (chezmoi/home-manager) could only relocate half of rledger's config. In the future when we have more configs like `report.config` etc setting the directory once is cleaner. Setting RLEDGER_CONFIG now prints a deprecation warning instead of silently doing nothing. We can remove that after a few weeks to make sure no one else has adopted that feature. 

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation (changes to docs and some alignment in markdown to make it easy to read)

I have selected breaking, but it should be okay because the feature that it breaks was only added a few days back and this PR extends that and makes it future proof.

## Changes

-

## Testing

- [x] `cargo test` 
- [x] `cargo clippy` passes

## Checklist

- [ ] New and existing tests pass locally

## Review Focus

- [x] Please review the algorithm/logic
- [ ] General review is fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ date_format = "%m/%d/%Y"
 rledger extract --importer chase chase-statement.csv
 ```
 
-Config is searched in: current directory, `~/.config/rledger/importers.toml`, or via `--config`.
+Config is searched in: current directory, the user config directory (`$RLEDGER_CONFIG_DIR` if set, else `~/.config/rledger/importers.toml`), or via `--config`.
 
 ### Transaction Categorization
 
--- a/crates/rustledger/src/bin/rledger.rs
+++ b/crates/rustledger/src/bin/rledger.rs
@@ -18,7 +18,7 @@
 //!
 //! If no file is specified, rledger will use the default file from config:
 //! ```bash
-//! rledger check  # Uses file from ~/.config/rledger/config.toml
+//! rledger check  # Uses file from the user config
 //! ```
 //!
 //! # Aliases
--- a/crates/rustledger/src/cmd/config_cmd.rs
+++ b/crates/rustledger/src/cmd/config_cmd.rs
@@ -112,18 +112,13 @@ fn run_show<W: Write>(raw: bool, format: &str, out: &mut W) -> Result<()> {
                 }
                 config::ConfigSource::Environment => {
                     writeln!(out, "# === Environment Variables ===")?;
-                    if let Ok(file) = std::env::var("RLEDGER_FILE") {
-                        writeln!(out, "RLEDGER_FILE={file}")?;
-                    }
-                    if let Ok(format) = std::env::var("RLEDGER_FORMAT") {
-                        writeln!(out, "RLEDGER_FORMAT={format}")?;
-                    }
+                    write_non_empty_env(out, "RLEDGER_CONFIG_DIR")?;
+                    write_non_empty_env(out, "RLEDGER_FILE")?;
+                    write_non_empty_env(out, "RLEDGER_FORMAT")?;
                     if std::env::var("NO_COLOR").is_ok() {
                         writeln!(out, "NO_COLOR=1")?;
                     }
-                    if let Ok(profile) = std::env::var("RLEDGER_PROFILE") {
-                        writeln!(out, "RLEDGER_PROFILE={profile}")?;
-                    }
+                    write_non_empty_env(out, "RLEDGER_PROFILE")?;
                     writeln!(out)?;
                 }
                 _ => {}
@@ -134,6 +129,13 @@ fn run_show<W: Write>(raw: bool, format: &str, out: &mut W) -> Result<()> {
         print_config(&loaded, format, out)?;
     }
 
+    Ok(())
+}
+
+fn write_non_empty_env<W: Write>(out: &mut W, name: &str) -> Result<()> {
+    if let Some(value) = std::env::var_os(name).filter(|v| !v.is_empty()) {
+        writeln!(out, "{name}={}", value.to_string_lossy())?;
+    }
     Ok(())
 }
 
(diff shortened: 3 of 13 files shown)


robcohen (dismissed):
Thanks for this — the direction is right (#1616's actual ask was always "relocate my config *directory*", and the file-level `RLEDGER_CONFIG` only got half of it), and the test discipline here is exemplary: pure `*_with_override` helpers for unit coverage, subprocess tests that set env on `Command` instead of `set_var` (exactly this repo's hermeticity rule), a wiring test that would catch a typo'd env-var name, the `PROGRAMDATA` shielding, and the macOS XDG interplay handled *and* documented. Genuinely nice work.

One premise needs correcting before this can merge, plus red CI:

### 1. `RLEDGER_CONFIG` is not "a few days old" — it has shipped in five releases

`RLEDGER_CONFIG` landed in #1650 on **2026-06-30** and is in **v0.17.0 through v0.21.0**, including Homebrew builds. And its target audience (chezmoi/home-manager users, per #1616) is exactly the crowd that bakes env vars into dotfile templates and updates them rarely. Making it **inert** with only a stderr warning silently breaks their config resolution — the warning helps interactive users, but dotfile-managed shells often never surface stderr from a status bar/cron invocation.

Suggested shape (small change to `user_config_file_with_override`): keep honoring it during the deprecation window —

1. `RLEDGER_CONFIG_DIR` set → wins (dir override, both files).
2. else `RLEDGER_CONFIG` set → still honored **for `config.toml` only** (its historical scope), with the deprecation warning you already wrote.
3. else native + macOS XDG fallback, as in this PR.

If both are set, warn and state that `RLEDGER_CONFIG_DIR` won. Then the removal "after a few weeks" becomes a release-notes item instead of a silent break. (Deprecation policy nit: weeks-not-releases is hard for downstream packagers — I'd say "removed in the release after next".)

### 2. CI is red — three checks

- **Clippy** (×2, both in the new code): `map(<f>).unwrap_or_else(<g>)` → `map_or_else` (in `importers_config_not_found_message`), and a `this could be a const fn`. `cargo clippy --all-features --all-targets -- -D warnings` locally will show both.
- **validate-docs**: the vitepress build fails on this branch (`build error` in the docs job) — likely a dead link/anchor in the docs edits; `cd website && npm run build` reproduces.

### Smaller notes (non-blocking)

- `warn_if_rledger_config_set` fires on every `Config::load()` — right call for a deprecation, but with the honor-during-window approach above, only warn when the var is actually *used* (or superseded), so `RLEDGER_CONFIG_DIR` adopters don't get a stale warning about a var they've already migrated off.
- Heads-up on a benign conflict: open PR #1786 touches `extract_cmd` around `find_importers_config` — whichever lands second needs a trivial rebase.
- `expand_os_path`'s lossy fallback for non-UTF-8 paths is a reasonable trade and the comment says why — good.

Happy to re-review quickly once the deprecation window + CI are sorted.

robcohen:
Update after maintainer review: **the hard break of `RLEDGER_CONFIG` is accepted** — one override mechanism is cleaner than two interacting ones, and the inert-plus-warning approach stands. Deprecation-window request withdrawn (review dismissed).

Design-wise this is the right end state: verified that the override reaches every rledger config-dir consumer on main (`config.toml` and `importers.toml` — both routed). The only `dirs::config_dir()` sites left are the query REPL's history/init files, which live under `…/beanquery/` — *beanquery's* app directory, shared with the Python tool deliberately. Leaving them out of `RLEDGER_CONFIG_DIR` is correct; please add a one-line doc comment on `get_history_path`/`get_init_path` (crates/rustledger/src/cmd/query/interactive.rs) saying the exclusion is deliberate, so the next config-relocation pass doesn't "fix" it.

What remains before merge is mechanical:

1. **Clippy** (×2 in the new code): `map(<f>).unwrap_or_else(<g>)` → `map_or_else` in `importers_config_not_found_message`, plus a `could be a const fn`. `cargo clippy --all-features --all-targets -- -D warnings` shows both.
2. **validate-docs**: the vitepress build fails on this branch — `cd website && npm run build` reproduces.

Happy to merge promptly once CI is green.

robcohen:
Dug into the CI failures — three findings, one of them cryptic enough to save you the decoding:

1. **Clippy** (2 errors in the new code):
   - `extract_cmd/mod.rs:334`: `map(...).unwrap_or_else(...)` → clippy's suggested `map_or_else` form.
   - `config.rs:779` `resolve_user_config_path`: `missing_const_for_fn` — on non-macOS the body reduces to returning `native`, so clippy wants `const fn`, but it can't be const on macOS (the XDG fallback). Suggest `#[allow(clippy::missing_const_for_fn)]` with a one-line comment saying why, rather than a cfg-split.

2. **`build` failing is downstream only** — it's the aggregator gate and Clippy is in its `needs`; it goes green when Clippy does.

3. **validate-docs — the markdown "alignment" is the culprit.** The reformat changed the *first line* of `docs/commands/extract.md` from a `______________` thematic-break rule to `---`. A `---` as the first line of a file opens **YAML frontmatter** in vitepress, which then chokes trying to parse `## title: …` as YAML — that's the "end of the stream or a document separator is expected at line 11" error. Restore the underscore-style rule (or drop the leading break entirely) in `extract.md` — and check the same first-line pattern in the other three touched docs files (`config.md`, `configuration.md`, `importing.md`), which have the same shape on main.

The branch base is fine (July 13, post-lockfile-fix main) — no rebase needed.

robcohen:
Pushed `ba110763` to green up CI and cover the deprecation (maintainer edit access — hope that's welcome; revert freely if you'd rather own these):

- **clippy** ×2: `map_or_else` in `importers_config_not_found_message`; `#[allow(clippy::missing_const_for_fn)]` on `resolve_user_config_path` (it can't be const on macOS — the XDG fallback hits the filesystem), with a why-comment.
- **validate-docs**: restored the `____`-style rule on the first line of `extract.md` and `configuration.md` (the reformat had turned it into `---`, which VitePress reads as a YAML frontmatter opener and then chokes on the `## title:` line). I reproduced the real `npx vitepress build docs` against the website repo's config — it **passes** now.
- **New test** `deprecated_rledger_config_is_inert_and_warns`: the breaking-change behavior (`RLEDGER_CONFIG` inert + warns) had no coverage, so a future refactor could have silently re-enabled it. Subprocess test asserts its config does not load and the warning prints.
- Documented that the beanquery REPL history/init paths under `.../beanquery/` are deliberately outside `RLEDGER_CONFIG_DIR` (per the earlier review note).

Design LGTM — unifying the override under one dir is the right end state and I confirmed both `config.toml` and `importers.toml` route through it.

Two minor, non-blocking notes for a follow-up (left as-is):
1. The deprecation warning prints **twice** for `config show` (`Config::load()` runs twice); a `OnceLock` guard would dedupe it.
2. `EnvironmentConfig.profile` is loaded but never applied in `apply_env_config` (profile is consumed in `rledger.rs`); its only effect here is flipping `has_source()`, so setting `RLEDGER_PROFILE` alone now records a `ConfigSource::Environment` whose values weren't merged. Cosmetic.

CI should re-run on the new commit.

Issue #1802, opened 2026-07-17
Per-source historical price support (follow-up to #1794)


robcohen:
#1794's fix gave Yahoo real historical support and made every latest-only source **refuse** `--date` instead of mislabeling the latest quote. This tracks upgrading refusal → real support where the provider APIs allow it:

| Source | Provider historical API | Notes |
|--------|------------------------|-------|
| coinbase | `GET /v2/prices/{pair}/spot?date=YYYY-MM-DD` | Straightforward |
| ratesapi | `exchangerate.host/YYYY-MM-DD?base=X&symbols=Y` | Straightforward |
| quandl | `start_date`/`end_date` dataset params | Already returns dated rows |
| alphavantage | `TIME_SERIES_DAILY` | Free tier is rate-limited |
| coincap / coinmarketcap | history endpoints (interval params) | CoinMarketCap historical requires a paid tier — may stay refusal |
| oanda | historical rates need a different product | May stay refusal |
| ecb | 90-day XML feed (`eurofxref-hist-90d.xml`) | Covers recent history only |
| eastmoneyfund | NAV history endpoint | |
| tsp | share-price CSV history | |

Semantics for each: fetch a window ending at the requested date, emit the last quote **on or before** it under the **quote's own date**, hard error when none (the beanprice `get_historical_price` contract, as now implemented for Yahoo).

Also from #1794's closing observation: the bean-price compat harness validates commodity discovery and source resolution but not quote correctness. A mock-HTTP fixture harness for the source implementations (record real provider responses, replay in CI) would let these behaviors regression-test without network — candidate for the testing roadmap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

robcohen:
Researched how beanprice, pricehist, Finance::Quote, and ccxt architect multi-provider price fetching (primary sources, verified against code). Proposal for the architecture before porting sources down the table above:

[...]

PR #1804, opened 2026-07-18
feat: historical support for the five remaining sources (#1802)

robcohen:
## What

Ports the five remaining #1802 table rows onto the window primitive from #1803. After this, only coincap, coinmarketcap, and oanda remain latest-only (paid-tier / product-gap questions per the issue).

| Source | Historical mechanism | Coverage |
|--------|---------------------|----------|
| ecb | SDMX data API `startPeriod`/`endPeriod` — the full EU-reference series, not just the 90-day XML feed the issue anticipated | `Since(1999-01-04)` |
| alphavantage | `TIME_SERIES_DAILY` (stock tickers; compact/full outputsize by window age). Forex/crypto ticker forms refuse **hermetically** before any network I/O, pointing at ratesapi/coinbase | `Full` |
| quandl | `start_date`/`end_date` dataset slice; the row parser generalizes to all rows | `Full` |
| tsp | the single-payload history endpoint sliced to the window | `Full` |
| eastmoneyfund | `f10/lsjz` NAV history with `startDate`/`endDate` (and the Referer header the API requires) | `Full` |

Design notes, applying the #1801/#1803 review lessons up front:

- **ECB cross-rates over a window are a per-date join** of the two legs — the latest path's leg-date-mismatch bail has no historical counterpart, because unjoined days (a frozen HRK/RUB series) simply produce no point and the dispatch's on-or-before selection works with what joined.
- **Skip, never today-label**: every parser skips observations with missing dates or unparsable values rather than falling back to a clock date — a today-fallback on a historical row would be discarded by the on-or-before selection and turn a served value into a spurious no-quote error.
- **Every parser is fixture-tested without HTTP** (`parse_series`, `parse_daily_series`, `parse_history`, plus URL-form pins for all five) — the parse/fetch split from #1803 makes this nearly free.

## How to test

`cargo test -p rustledger price` (626 tests, all hermetic). Live smoke: `rledger price EUR -c USD -s ecb --date 2024-06-14`, `rledger price AAPL -s alphavantage --date 2024-06-14` (needs API key), `rledger price 000001 -s eastmoneyfund --date 2024-06-14`.

Part of #1802. Follow-ups still tracked there: typed `SourceError`, window write-through into the price cache, mock-HTTP fixture harness.

https://claude.ai/code/session_013gJH7uBBF9Xw9et9QFRcsU

### DIFF
--- a/crates/rustledger/src/cmd/price/sources/alphavantage.rs
+++ b/crates/rustledger/src/cmd/price/sources/alphavantage.rs
@@ -2,7 +2,7 @@
 //!
 //! Fetches stock, forex, and crypto prices from Alpha Vantage.
 
-use super::{PricePair, PriceSource, user_agent};
+use super::{DateWindow, HistoricalCoverage, PricePair, PricePoint, PriceSource, user_agent};
 use crate::cmd::price::PriceResponse;
 use anyhow::{Context, Result};
 use rust_decimal::Decimal;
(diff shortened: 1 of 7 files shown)


robcohen:
**Round-1 deep review** (25 agents, adversarial verification): 10 findings (9 confirmed, 1 plausible hardened), all addressed in e82f9f78.

1. **ECB empty windows violated the `fetch_window` contract**: the SDMX API answers an empty selection with HTTP 404, which surfaced as a raw transport error instead of the dispatch's clean no-quote message — a frozen HRK window read as a network outage. 404 now maps to the contract's empty vec.
2. **alphavantage could serve a live in-progress bar as a settled close**: the dispatch's past-vs-today split uses the local calendar while `TIME_SERIES_DAILY` days are US/Eastern — a UTC+8 user's local "yesterday" can be the still-open US session. The window path now drops the Eastern-today bar (pure helper, unit-tested).
3. **URL-parameter injection closed at the dispatch**: tickers/currencies flow into provider URLs unescaped in every source, and a crafted ticker (`000001&endDate=2019-01-01` from shared-ledger `price:` metadata) could rewrite the window parameters. `reject_url_metacharacters` now guards both values once, canonically, before any routing — real ticker shapes (`BRK.B`, `CRYPTO:BTC`, `WIKI/AAPL`, `EURUSD=X`) pass, and the pre-existing latest-path exposure closes with it.
4. **Two latest-path regressions restored**: the series refactors had dropped the LATEST-fetch today-fallback for undateable rows in ecb and quandl, turning rates main served into misleading hard errors. Both parsers now take an explicit undated-fallback (today on the latest path, skip on the historical path — the same split `feed_date_or` documents).
5. **eastmoneyfund hardening**: `DWJZ` parses via the canonical `price_decimal_from_json` (string *and* number shapes, like every sibling parser); `ErrCode` tolerates string typing; and windows wider than one 49-entry page refuse explicitly instead of silently truncating (paging can come with a real consumer).
6. **The dispatch-level hermetic pin is restored**: `dated_forex_refuses_through_the_dispatch` exercises alphavantage's forex refusal through `fetch_price`, not just a direct `fetch_window` call — a future dispatch reorder can't silently unhermeticize it.
7. **Canonical-function discipline**: the three inline date slice-and-parse copies (quandl/tsp/eastmoneyfund) now call the new strict `feed_date` helper, which `feed_date_or` delegates to.

628 tests green, clippy/doc clean.

Issue #1766, opened 2026-07-12
WIT contract: entries-accepting surfaces classify/render with default ledger options


robcohen:
Design-debt tracker from the 2026-07 WIT deep-dive review.

Every surface that accepts already-loaded directives instead of source text runs with **default ledger options**, because the contract has no way to carry them alongside an entry list:

- `builder.query-entries` — `AccountTypes::default()`, so POSSIGN misclassifies renamed-root ledgers (the L5 note in convert.rs).
- `session.from-entries` — held entries carry no options (documented at the constructor).
- `format.format-loaded` (3.6.0) — `FormatConfig::default()`, ignoring `display-precision`/`render-commas`.

The debt compounds with each new entries-accepting function. Candidate designs, roughly in order of appeal:

1. **Session-held options**: `session.from-entries-with-options(entries, options: ledger-options)` (additive), making the session the canonical carrier — new entry-consuming capability then goes resource-first and inherits options for free.
2. Optional `options` parameter on each free function — additive but N repetitions of the same fix.
3. A breaking 4.0 that removes the free-function tier and makes the session the only entries surface.

Related: the free functions `builder.filter`/`clamp`/`query-entries` duplicate session methods; if (1) lands, their docs should soft-deprecate them in favor of the session.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

robcohen:
PR #1805 implements design 1: WIT 3.7.0 adds session.from-entries-with-options(entries, options), making the session the canonical options carrier. query now classifies with the ledger's own name-* roots (the L5 POSSIGN note), info() echoes the provided options, and the builder free functions are doc-soft-deprecated in favor of the session.

Two scope corrections from implementation, so this issue stays accurate:

1. The format-loaded row is deeper than an options-threading gap: the rendering pipeline (FormatConfig is alignment+indent only; canonicalize_directives takes no DisplayContext) has no display-precision/render-commas machinery anywhere, CLI included. A session.format honoring those options first needs the formatter to grow precision support — leaving that tracked here rather than shipping a method that ignores the options it advertises.

2. session.clamp's synthesized Equity:Opening-Balances accounts looked like the same debt, but ops::clamp hardcodes them for ALL callers (from-source sessions and the CLI too) — a global gap needing an ops-level config, not an entries-surface asymmetry. Worth its own line item.

rustfava adoption (passing load-time options into the filtered-query path) rides the next release train.

robcohen:
[...]

PR #1807, opened 2026-07-18
feat: formatter honors per-currency display precision (#1766)

robcohen:
## What

The remaining #1766 blocker: the rendering pipeline had **no precision machinery anywhere** (CLI included), which is why `format-loaded` couldn't honor `display-precision` and `session.format` was deliberately not shipped in #1805. This PR grows it at the canonical layer:

- `FormatConfig` gains `number_display: Option<DisplayContext>` — reusing the canonical precision machinery the query/report renderers already use (per-currency fixed precisions from `option \"display_precision\"` + observed distributions), rather than duplicating it.
- Every amount/cost/price number emission flows through one `render_number` chokepoint and config-threaded formatters, so the two behaviors (context-padded vs historical own-scale) cannot drift per call site. `None` is byte-identical to the historical rendering.
- **Deliberate decision**: thousands separators are NOT emitted in ledger text even when the context has `render_commas` — canonical form carries no separators (the CST canonicalizer strips them by definition, matching `bean-format`), and `render_number` strips any grouping so direct emitters agree byte-for-byte with the canonicalize shim. Commas remain a report/query display concern, honored where they always were. This surfaced from a failing test: pass 1 emitted `1,234.50`, pass 2 canonicalized it away — the test now pins the resolved semantics.
- Bare interpolation-target numbers (no currency) keep their own scale.

## Tests

Core `render_number` + directive-threading tests; a parser-level test pins precision padding through the two-pass canonicalize shim and byte-fidelity under the default config. Full workspace green (3937 tests).

## Next

The follow-up PR adds WIT `session.format` building the `DisplayContext` from held options — completing #1766's format surface.

Part of #1766.

https://claude.ai/code/session_013gJH7uBBF9Xw9et9QFRcsU

### DIFF
--- a/crates/rustledger-core/src/directive.rs
+++ b/crates/rustledger-core/src/directive.rs
@@ -1981,7 +1981,13 @@ mod tests {
     fn meta_value_int_display_and_kind() {
         assert_eq!(MetaValue::Int(42).to_string(), "42");
         assert_eq!(MetaValue::Int(-7).to_string(), "-7");
-        assert_eq!(crate::format::format_meta_value(&MetaValue::Int(42)), "42");
+        assert_eq!(
+            crate::format::format_meta_value(
+                &MetaValue::Int(42),
+                &crate::format::FormatConfig::default()
+            ),
+            "42"
+        );
         assert_eq!(meta_value_kind(&MetaValue::Int(0)), "int");
     }
 
--- a/crates/rustledger-core/src/display_context.rs
+++ b/crates/rustledger-core/src/display_context.rs
@@ -419,27 +419,6 @@ impl DisplayContext {
     /// precision via `round_dp(dp)`. That was correct for under-scale
     /// padding but wrong for over-scale truncation — it lost
     /// arithmetic precision that bean-query preserved (closes #1103).
-    /// Ledger-text variant of [`Self::format`] (#1766): pads a TRACKED
-    /// currency's value to the tracked precision (never rounding an
-    /// over-precise value away), and returns an UNTRACKED currency's
-    /// value at its own scale, byte-faithful — no `normalize()`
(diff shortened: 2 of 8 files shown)


robcohen:
**Deep review** (24 agents, adversarial verification): 10 findings, all addressed in 6a3dfe15 + 86909fac.

The one real correctness bug: **untracked currencies lost trailing zeros**. `render_number` went through `DisplayContext::format`, whose untracked branch calls `normalize()` — so with a context tracking only USD, a `100.50 EUR` posting reformatted to `100.5 EUR`, silently widening the balance assertion's implicit tolerance 10×. Fixed with a new `DisplayContext::format_plain`: tracked currencies pad (never round), untracked currencies stay byte-faithful, and separators are never emitted — which also deletes the comma add-then-strip inefficiency outright.

The chokepoint now genuinely covers every number: **balance tolerances** render with the amount's currency precision (mixed precision on one line read as a typo), and **amount-typed metadata and custom-directive values** thread through `format_amount_with` (the config now flows through `format_meta_value`/`metadata_lines`, with posting metadata keeping its deeper indent).

Also fixed: a doc splice that had cut `format_directives`' rustdoc in half and attached its head to `render_number` (both restored, and the pad-never-round / untracked-faithful semantics documented at the field, the chokepoint, and `format_plain`); `format_incomplete_amount` calls the canonical instead of re-deriving it; dead `format_amount` removed; constructors use `..Self::default()`; exact-equality asserts replace substring proxies; and a new parser-level drift guard pins padded **cost and price** numbers surviving the canonicalize shim's pass 2 (proving the CST re-canonicalization preserves trailing zeros) plus pass-1/shim agreement on rendered numbers.

Workspace green (3938 tests), clippy/doc clean.

Issue #1814, opened 2026-07-20
feat: investment returns reporting (rledger report returns — XIRR + TWR)


robcohen:
## Summary

Add investment **returns reporting** to rledger — money-weighted (XIRR) and time-weighted (TWR) returns per investment group, computed by a reusable engine that the CLI (and later BQL/FFI/rustfava) all share. Requested in discussion #1789.

This plan is grounded in (a) a code survey of rledger's existing primitives and (b) research across the plain-text-accounting + portfolio-tool ecosystem (beangrow, fava-portfolio-returns, fava-portfolio-summary, hledger `roi`, hoostus/portfolio-returns, Portfolio Performance). The closest prior art is **`hledger roi`** — a first-class built-in command computing both IRR and TWR with a query-defined investment boundary and auto-inferred cash flows — and rledger's command should take essentially that shape. We deliberately do **not** port beangrow's external-protobuf-config model, which is the ecosystem's most-cited pain point (see beangrow#10).

## Architecture — a shared engine, not CLI-buried

Per the repo's canonical-function discipline (no consumer re-derives pipeline logic), the returns math lives in a **new `rustledger-returns` leaf crate** (deps: `rustledger-core` + `rustledger-booking`), sitting below all consumers so none re-derives it.

- [ ] **New crate `rustledger-returns`**: cash-flow extraction + XIRR (Newton/bisection root-find, no external solver) + TWR. Deps: core + booking only.
- [ ] **Relocate `PriceDatabase` (`rustledger-query/src/price.rs`) down into `rustledger-core`** and re-export from `rustledger-query` for back-compat. It's a price index over `Price` directives — a core concern alongside `Inventory` — and it's coupled only to core today, so this inverts the layering cleanly and lets the returns crate use it without dragging in the query engine.
- [ ] **CLI surface**: `rledger report returns` — a `Report::Returns { … }` variant (`report_cmd/mod.rs`) + `report_cmd/returns.rs`, wrapping the engine. First-class command, hledger-`roi`-shaped.

## Grouping UX — query-pattern first, metadata as persistence

Ecosystem evidence: hledger (`--inv`/`--pnl` queries) and hoostus (`--account` regex) define the boundary **at invocation**; fava-portfolio-summary uses **account metadata** (`portfolio:` on the open directive) for persistent groups; beangrow uses an external protobuf file and is criticized for it. Query-scoping is zero-config, zero-annotation, and more flexible than metadata (metadata forces one holding into one group).

[...]

PR #1816, opened 2026-07-20
feat(returns): cash-flow extraction from a booked ledger

robcohen:
## What

Phase 1, part 2 of investment returns reporting (#1814), building directly on the XIRR engine merged in #1815. This adds **cash-flow extraction**: the layer that turns a booked ledger into the `CashFlow` series `xirr` consumes.

## The model — boundary crossing

Every account is classified by a `Scope` into one of three roles:

| Role | What it is | Contributes a flow? |
|------|-----------|---------------------|
| **Investment** | holdings + settlement cash | No — value tracked *inside* the portfolio |
| **Income** | dividends / gains / fees generated by the investment | No — internal to the return |
| **External** | bank, outside cash | **Yes** — money crossing the boundary |

For a transaction touching an investment or income account, the flow is the sum of its **external** postings, converted to one reporting currency at the transaction's date:

```
cashflow_t = Σ { external postings }  =  −Σ { investment + income postings }   (equal, by double-entry)
```

Summing the external side directly makes the sign land **investor-centric** with no manual negation.

## Three correctness properties fall out of the model, not special cases

- **Purchase** (`Assets:Bank -1000 USD`) → **−1000** (outflow) ✓
- **Sale / dividend-to-bank** (`Assets:Bank +500 USD`) → **+500** (inflow) ✓
- **Internal transfer** — shares/cash moved between two in-scope accounts — has no external posting, so its flow is **0** and it does **not** register. Double-counting internal transfers is the single most error-prone mistake in returns math (flagged as *the* trap in #1814); here it is excluded **structurally**.

A synthetic **terminal** flow values the position still held on the end date at market — realized through the booking engine (the same realization as `report_cmd::account_balances`, so lots keep cost and reductions match each account's booking method) and priced via the oracle.

## `PriceOracle` — keeping the crate a leaf

Prices arrive through a new `PriceOracle` trait, not a concrete price database. This:

- keeps `rustledger-returns` a **leaf** (no dependency on `rustledger-query`, which owns the price index);
- makes the correctness core **unit-testable** against a hand-specified rate table — no price-DB fixtures leaking the index's own behavior into extraction tests;
- is a **pass-through** to wire up: the trait method mirrors `PriceDatabase::convert`, so the CLI consumer implements it in one line.

The full core-relocation of `PriceDatabase` that #1814 sketches can still happen on its own merits; the trait means extraction doesn't have to wait on it.

## Fails loud

`extract` returns `ExtractError::MissingPrice` rather than dropping a flow it can't state in the reporting currency — a silently omitted flow doesn't error, it just yields the wrong rate.

## Tests (15 extraction + 16 existing xirr = 31)

Sign conventions; the internal-transfer and reinvested-dividend traps; multi-currency conversion; the horizon cutoff; terminal valuation (held position, broker cash, fully-sold → `None`); missing-price errors on both flow and terminal paths; segment-aware classification; and an end-to-end `extract_cash_flows` → `xirr` sanity check (~10%).

## Deferred to follow-ups (per #1814)

- Dividend / ex-dividend breakout (total vs. ex-income return)
- Time-weighted return (built on the same realization primitive as `terminal_value`)
- The `rledger report returns --investments <pattern>` CLI surface + the real `impl PriceOracle for PriceDatabase` (integration-tested there, where both crates are in scope)

Part of #1814.


robcohen:
## Deep review — round 1

Ran the workflow-backed review (4 finder angles, adversarial verify pass, high effort) against the branch diff. It surfaced 7 verified findings (3 more were correctly refuted). Dispositions below.

### The headline finding — and why its framing inverts the correctness

The finders flagged that `terminal_value` ignores `Pad`/`Balance` directives, and that "no single input stream is correct for both `extract_flows` and `terminal_value`" because a pad's `Equity:Opening-Balances` leg would inject a "spurious" cash flow. I worked through it and the opposite holds:

With the **booked, pad-expanded** stream (`Ledger::balance_view` — exactly what the canonical `report_cmd::account_balances` is fed), both halves are *coherent*. A pad seeding `Assets:Broker:Cash 500 / Equity:Opening-Balances -500` produces a **-500 opening flow** (capital in) that **pairs** with the **+500** that account contributes to the terminal value. `xirr([-500, +500])` -> ~0%, which is correct for an untouched opening balance. Dropping that flow is the *actual* bug: a +500 terminal with no offsetting outlay is an undefined return.

So the opening flow is not spurious — it is *required*. The real defect the finders correctly exposed is that **my input contract was unspecified** (I never stated "pass `balance_view`") and I had no test proving pad coherence.

**Fixed** (`419ca1cc`): a precise `# Input contract` on `extract_cash_flows` + module docs (booked + pad-expanded; from-inception semantics; opening balances are genuine capital flows), and two tests pinning the coherence — `opening_balance_pairs_with_terminal_to_a_zero_return` (0%) and `opening_balance_then_internal_growth_is_a_positive_return` (+10% via an internal dividend). This is the same booked-input precondition `apply`/`account_balances` already carry.

### Units vs. balance-weight (CONFIRMED — doc corrected)

External postings are summed by **units** (the money that actually moved through the outside account). The module's double-entry identity was stated over units; it holds over balance *weights* / plain postings, not raw units in general. Corrected the claim, and documented the deliberate limitation: an external leg carrying its own `@`price or `{cost}` (an in-kind transfer, or a cash leg with an explicit rate) is valued by units through the oracle, not its stated weight. Common-case external legs are plain cash; honoring per-posting weights is a noted follow-up.

### None-amount external posting (PLAUSIBLE — clarified)

The defensive `continue` on an external posting with no resolved amount is unreachable under the booked-stream contract (interpolation fills every posting), so it is not a silent drop of a real flow. Comment clarified to say so and tie it to the contract.

### Declined / deferred, with reasons

- **`Decimal += overflow panic` (PLAUSIBLE)** — declined. The overflow ceiling is ~7.9e28; a monetary sum reaching it exceeds world GDP by ~14 orders of magnitude, and `checked_add` on every `+=` would diverge from how the rest of the codebase (booking, inventory) does `Decimal` arithmetic. Not worth an error variant for an unreachable case.
- **Double directive pass / re-book (CONFIRMED, perf)** — deferred. Not a correctness issue; `extract_flows` iterates once and `terminal_value` rebuilds the engine once, called once per report. Optimizing before the consumer exists is premature; a future refactor can have `terminal_value` accept a pre-realized inventory (also lets the caller own pad expansion).

### Refuted (agreed)

Three findings claimed the `!net.is_zero()` guard drops legitimate zero-netting flows. Correctly refuted: a zero-valued flow is neutral to `xirr` (contributes 0 to NPV at every rate and is not a sign change), so dropping it is harmless.

34 tests; clippy/doc/fmt clean.

robcohen:
## Deep review — round 2

Second workflow-backed pass (4 finder angles + adversarial verify), pointed at the post-round-1 state with deeper angles (real BookingEngine lot semantics, mixed-currency positions, classify precedence under nested trees, short terminals, same-date flows vs xirr, determinism, vacuous-test check). 9 verified findings, 3 refuted. All addressed in `e195f799`.

### Real correctness bug — overlapping-prefix classification (CONFIRMED)

`classify()` used "investment always wins" on an account matching both role lists. With income scoped **under** an investment prefix (income `Assets:Broker:Dividends`, investment `Assets:Broker`), a reinvested dividend's income leg was classified Investment and summed into the terminal value, **cancelling the asset it created** — understating the return toward ~0%.

**Fixed:** `classify()` now takes the role of the **longest** (most-specific) matching prefix across both lists, so nesting resolves either direction correctly (Investment breaks an exact-length tie). Added a classification test and an end-to-end nested-income terminal regression test.

### Net-flat held portfolio reported as None (CONFIRMED)

`terminal_value` returned `None` when a held portfolio's market value netted to exactly zero, indistinguishable from fully-liquidated. **Fixed:** a held position is always reported (keyed off `held`); `None` only when nothing is held. Test added (offsetting AAPL/GOOG positions → `Some(0)`).

### Determinism + canonical drift (CONFIRMED / PLAUSIBLE)

`terminal_value` iterated `into_inventories()` (an `FxHashMap`), so which currency a `MissingPrice` named was hash-order dependent and differed 32/64-bit. **Fixed:** collect into a `BTreeMap` like the canonical `account_balances`. Added cross-reference comments at *both* this loop and `account_balances` documenting the deliberate re-derivation (a leaf crate can't call the CLI-side helper) and flagging the drift-guard parity test for the returns CLI PR.

### Docs (CONFIRMED)

- `# Panics` added to `terminal_value` / `extract_cash_flows`: an un-booked stream (a contract violation the leaf can't detect) can make the booking engine `debug_assert` rather than return `Err`.
- `PriceOracle` now requires conversion be **linear in the amount** (`convert(-x) == -convert(x)`) — the property that makes a short position value negatively. (This grounds a finding the verifier had refuted as unstated-contract; I tightened the contract rather than leave the reliance implicit.)

### Test gaps (CONFIRMED) + cleanup

Added the three missing coverage cases the finders flagged as vacuously-passable: **net-short** terminal (−1500), **multi-position** terminal sum (AAPL+GOOG → 2500), and a **same-end-date** flow+terminal fed through `xirr` (~50%). Also: `extract_flows` now classifies each posting **once** (single pass) instead of twice per transaction.

### Refuted (agreed)

Two duplicate FxHashMap-nondeterminism entries (kept one), and the PriceOracle sign-linearity concern — which I nonetheless closed by tightening the trait doc.

**40 tests** (+6); clippy/doc/fmt clean; CLI crate builds with the cross-reference comment.

robcohen:
## Deep review — round 3

Third pass, split between auditing the round-2 changes for regressions and new angles. It found **one live bug — a regression round 2 introduced** — plus the already-documented deferred drift-guard. Two findings refuted (both of which I'd reasoned through). The round-2 changes it audited (longest-prefix classify, BTreeMap determinism, currency comparison, date arithmetic, docs) all held up.

### Regression: zero-valued terminal flow fabricates a return (CONFIRMED)

Round-2 finding-2 asked terminal_value to emit a flow even when the held value is zero (to distinguish "net-flat holding" from "liquidated"). I took it — and it was wrong. A zero-valued terminal on a date *later* than the real flows is **not neutral to xirr**: it defeats xirr's all-same-date degenerate guard. Concrete failure the finder gave: a same-date deposit/withdrawal wash (`-1000`/`+1000` on one date) plus a worthless residual holding yields `[-1000, +1000]` (same date) + `[0 @ end_date]`; the later zero flow bypasses the guard, NPV is identically zero, and Newton returns its **0.10 seed → a fabricated +10% return** for a genuinely-undefined series.

**Fixed** (`cce68559`): reverted to emitting a terminal flow only when the held market value is nonzero (`None` otherwise — covering empty *and* net-flat/worthless portfolios). Added a regression test asserting `xirr` returns `None` for the wash, and corrected the round-2 net-flat test to expect `None`. The informational "net-flat vs liquidated" distinction round 2 wanted must not be smuggled through a zero cash flow; if a consumer ever needs it, it belongs in a richer report API.

I also **reverted round-2's single-pass `extract_flows` rewrite**. It allocated a `Vec` per *every* transaction — including the common irrelevant ones — to save a re-classify on only the few relevant ones. The original allocation-free two-pass (the relevance gate short-circuits, so an irrelevant transaction costs only the scan) is better for the common path; this also moots the refuted-but-real allocation concern.

### Deferred (PLAUSIBLE): canonical drift-guard test

`terminal_value` re-derives `account_balances`' realization loop. I've documented the divergence at both sites (round 2); the parity drift-guard test genuinely needs the CLI crate (`account_balances` is CLI-side), so it lands with the `rledger report returns` consumer PR. Deferral stands.

### Follow-up noted (out of scope): xirr degenerate edge

The underlying xirr behavior — returning its Newton seed when NPV is identically zero across ≥2 distinct dates (reachable only by a pathological multi-date wash series) — is a pre-existing edge in the merged engine, not introduced here. This PR removes the extraction-side trigger; I'll file a small xirr-hardening follow-up (evaluate NPV at a second rate to detect a flat-zero function) rather than expand this PR into the merged solver.

### Refuted (agreed)

The single-pass None-skip (unreachable under the booked contract) and the `Vec`-vs-`SmallVec` allocation (moot after the revert above).

**41 tests** (+1 net regression test); clippy/doc/fmt clean.

robcohen:
### Re: Copilot inline review

Thanks — all three checked against the current head (`cce68559`); Copilot reviewed an earlier commit, and the manual review rounds already cover these:

1. **`sort_by_key` "unstable" (extract.rs:263)** — `slice::sort_by_key` is a **stable** sort in Rust (`sort_unstable_by_key` is the unstable variant). Same-`date` flows therefore keep insertion order (real flows before the terminal), so the series is deterministic as documented. Independently, `xirr` is order-independent within a date, so ordering can't affect the result either way.

2. **"single positive terminal cash flow" (extract.rs:271)** — already corrected (round 2): the docs now say "a single terminal cash flow" and explicitly document that a net-short position values negatively. The only remaining "positive terminal" text is a comment on one test describing that specific positive case.

3. **`amount() == None` silently skipped (extract.rs:250)** — documented as unreachable under the input contract: extraction requires the **booked, pad-expanded** stream (interpolation fills every posting's units), the same precondition `BookingEngine::apply` and `account_balances` carry. The skip is defense-in-depth, not a silent drop of a real flow; the contract is spelled out on `extract_cash_flows` and in the module docs.

Resolving these threads.

PR #1818, opened 2026-07-20
feat(returns): rledger report returns (money-weighted / XIRR)

robcohen:
## What

The consumer that makes investment returns reachable from the command line — phase 1's user-facing piece of #1814, built on the merged XIRR engine (#1815) and cash-flow extraction (#1816):

```
rledger report returns --investments Assets:Investments [--income Income:Dividends] \
                       [--currency USD] [--end 2020-12-31]
```

## How it wires together

- **`Report::Returns` clap variant + `report_cmd/returns.rs`.** `--investments` (repeatable, required) and `--income` (repeatable) define the portfolio boundary; `--currency` defaults to the ledger's first `operating_currency` option; `--end` defaults to today. `Report::Returns` is added to the `needs_balance_view` gate so extraction receives the **booked, pad-expanded** stream its input contract requires.
- **`impl PriceOracle for PriceDatabase` via a `PriceDbOracle` newtype in the CLI** (the composition root). The `convert` signatures are byte-identical, so it's a pass-through. This keeps `rustledger-returns` a leaf and `rustledger-query` free of a returns dependency, rather than putting the impl in query and pulling returns into every BQL build.
- Reports the **money-weighted return** plus supporting figures (capital invested, distributions received, current market value, flow count) in **text / json / csv**. An undefined series (no sign change) reports `n/a`, never a fabricated rate.

## The boundary model, visible in the tests

The `--income` scope decides whether a dividend-to-bank counts. On the same ledger (buy 10 AAPL for 1000, a 20 dividend to the bank, hold to a 130 close):

| invocation | flows | money-weighted return |
|---|---|---|
| `--investments Assets:Broker --income Income:Dividends` | −1000, +20, +1300 | **32.36%** |
| `--investments Assets:Broker` (no `--income`) | −1000, +1300 | **30.00%** |

Without `--income`, the dividend transaction touches no investment account, so it isn't a flow — the boundary model, working as designed.

## Tests

- **Drift-guard unit test** (owed from #1816 round 2, per CLAUDE.md Canonical-Function Discipline): `terminal_value`'s realization must equal the market valuation of `report_cmd::account_balances`' inventories for the same scope/date. The two deliberately re-derive the same realization loop across a crate boundary; this pins their agreement.
- **5 integration tests**: the boundary model (both rows above), the reporting-currency default, the json fields, and the missing-currency error path.

## Scope

XIRR (money-weighted) only, whole-scope. **Time-weighted return** and **per-commodity grouping** remain follow-ups per #1814 (TWR isn't in the engine yet). No new publishable crate (returns landed in #1815).

Part of #1814.


robcohen:
## Deep review — round 1

Workflow-backed review (4 finder angles + adversarial verify) on the CLI consumer. Two confirmed findings, one refuted; it also cleared the JSON-`NaN` worry I flagged (the engine's `xirr` only returns `Some` for finite values, so the unquoted `money_weighted_return_pct` can never be a bare `NaN`/`Infinity`). All addressed in `a32eddbe`.

### Canonical-Function Discipline (CONFIRMED)

`report_returns` hand-combines `extract_flows` + `terminal_value` into the `xirr` series — duplicating the engine's canonical `extract_cash_flows`. I build it from the parts deliberately (the summary needs `invested`/`distributions`/`current_value` separately, and calling `extract_cash_flows` too would re-realize inventory a second time), but nothing pinned the equivalence. **Fixed:** a drift-guard test `series_matches_extract_cash_flows` asserts the manual combine equals `extract_cash_flows` on the same input, plus a cross-reference comment. Per CLAUDE.md rule 3.

### Per-format rate precision (CONFIRMED)

The money-weighted return was formatted at 2 decimals for text/csv but 4 for JSON — the same ledger reported different rates across formats. **Fixed:** unified to 2 decimals everywhere (JSON emits the bare 2dp number).

### Surfaced while adding coverage

- **`-0.00%`**: the new pad test exposed that a 0% return (capital returned unchanged) converges to a tiny signed epsilon rendering as `-0.00%`. **Fixed** in `rate_pct` (a rate rounding to zero now shows `0.00%` in every format) — a real output-quality bug, not just a test tweak.
- **Pad coverage**: my fixtures used only explicit buys, so nothing proved the `needs_balance_view` addition (the whole reason `Report::Returns` is in that gate). Added `pad_seeded_holding_is_valued` — a `pad`+`balance`-seeded position valued end-to-end (invested 500, current value 500, 0%), which fails if the pad-expanded stream doesn't reach extraction.

### Refuted (agreed)

Money fields not `json_escape`d in the JSON branch — correct to refute: `format_amount_number` emits only digits/commas/dots/minus (no `"`/`\`/control chars), and holdings.rs's JSON doesn't escape them either.

8 returns tests (2 unit drift-guards + 6 integration); clippy/doc/fmt clean.

robcohen:
Good catch on the vacuous `contains("USD")` — fixed in b3ad6833: the currency-default test now asserts the **Reporting currency** line's value is `USD` (via the field extractor), so it can't pass on the fixture's other USD literals if the default were wrong. Resolving.

Issue #1820, opened 2026-07-20
returns: per-group breakdown for report returns (per-commodity + returns-group: metadata)


robcohen:
## Summary

Add **per-group breakdown** to `rledger report returns` — today it reports a single money-weighted return for the whole `--investments` scope. Phase-2 follow-up to #1814 (the MVP whole-scope report shipped in #1818).

## Auto per-commodity grouping

- [ ] Within the investment scope, compute a return **per commodity/holding** in addition to the whole-scope total, so a user sees which positions drove the return (the fava-portfolio-returns / beangrow shape).
- [ ] Design note: the boundary-crossing extraction produces reporting-currency cash flows for the *whole* scope, not per-commodity. Per-commodity attribution needs either (a) a sub-scope per commodity (the accounts/lots holding it) or (b) splitting flows by the commodity each transaction touches. (a) reuses the existing `Scope` machinery and is the cleaner first cut; capture the decision in the code.
- [ ] Output: a row per group plus a total row, across text/json/csv (mirror the holdings report's multi-row shape).

## Metadata-persisted named groups

- [ ] `returns-group:` metadata on `open` / `commodity` directives for saved, named groups (the fava-portfolio-summary idiom; precedent in rledger: the `precision:` commodity metadata). Read via a `rustledger_core` parse helper so every consumer parses it identically.
- [ ] Whole-portfolio + multi-group reports (report every named group at once).
- [ ] Do **not** add an external group-config file — that is beangrow's documented anti-pattern (beangrow#10), explicitly rejected in #1814.

## Context

- Report: `crates/rustledger/src/cmd/report_cmd/returns.rs` (currently single whole-scope summary).
- Engine: `crates/rustledger-returns` (`Scope`, `extract_flows`, `terminal_value`).
- Tracking: #1814. Pairs with the TWR follow-up (each group would report both metrics).

PR #1830, opened 2026-07-20
feat(returns): time-weighted return (TWR) in engine and report

### LINKED ISSUE #1819: returns: add time-weighted return (TWR) to the engine and report
## Summary

Add **time-weighted return (TWR)** to the returns engine and surface it in `rledger report returns`. Phase-1 follow-up to #1814; the money-weighted return (XIRR) shipped in the engine (#1815 / #1816) and the CLI report (#1818).

TWR complements the money-weighted return already reported: MWR (XIRR) answers "what did *I* earn given when I moved money", TWR answers "how did the *investments* perform, independent of my contribution timing" — the GIPS/manager-comparison metric. The plan (#1814) calls for **both, on by default, MWR as the headline**.

## Engine (`rustledger-returns`)

- [ ] Add a `twr` (and/or `modified_dietz`) function. True TWR values the position at **each cash-flow date**, computes the sub-period return between consecutive valuations, and chains them geometrically. Modified Dietz (beangrow's choice) is a closed-form single-period approximation and an acceptable simpler first cut.
- [ ] The building blocks already exist: `terminal_value`'s realization (booking engine over directives ≤ a date) generalizes to "value the position as of date D", and `PriceOracle` does the conversion. Factor out a `value_position_at(directives, scope, ccy, oracle, date)` helper (currently inlined in `terminal_value`) so both TWR sub-period valuations and the terminal reuse it.
- [ ] Annualize before returning (matching the MWR/hledger convention).
- [ ] Unit tests, including the case where TWR and MWR diverge (a large late contribution into a rising position).

## CLI (`rledger report returns`)

- [ ] Add a `Time-weighted return` line to the text summary and a `time_weighted_return_pct` field to json/csv. MWR stays the headline.
- [ ] Consider a `--metric mwr|twr|both` flag (default both), mirroring the fava/hledger toggle — though in compiled Rust the perf argument for defaulting TWR off is likely moot.

## Validation

- [ ] Cross-check TWR against `hledger roi` and beangrow's `example/` outputs on the same ledger (the reference implementations named in #1814).

## Context

- Engine: `crates/rustledger-returns/src/{lib.rs,extract.rs}` — `terminal_value` is the realization to generalize.
- Report: `crates/rustledger/src/cmd/report_cmd/returns.rs`.
- Tracking: #1814. Blocked-by: nothing (engine + report both exist).

robcohen:
## What

Adds the **time-weighted return (TWR)** alongside the money-weighted return, closing #1819 (phase-1 of #1814). The two answer different questions:

- **MWR (XIRR)** — "what did *I* earn, given *when* I moved money" (the personal-performance metric, already shipped).
- **TWR** — "how did the *investments* perform, independent of my contribution timing" (the GIPS / manager-comparison metric).

Report both, MWR as the headline.

## Engine (`rustledger-returns`)

`twr(directives, scope, currency, oracle, end_date) -> Result<Option<f64>, ExtractError>` — the annualized time-weighted return via the **unit-value (mutual-fund NAV) method**: value the portfolio at every external cash-flow date, chain the sub-period returns geometrically, annualize (actual/365, matching `xirr`). Each contribution/withdrawal is netted out at its date, so only market movement — and income the holdings generate (reinvested dividends, internal reshuffling, which aren't external flows) — drives the result. `None` when undefined (no flows, a single-day span, or a non-positive portfolio value where unit accounting breaks down — a documented MVP limitation).

Refactored `terminal_value`'s realize-and-value loop into a shared **`investment_value_at`** primitive that both `terminal_value` (value at the end date) and `twr` (value at each flow date) use.

## CLI (`rledger report returns`)

A **`Time-weighted return`** line in text and a **`time_weighted_return_pct`** field in json/csv, next to the money-weighted return.

## The divergence, in the tests

A ledger that sits flat for the first half-year, then takes a large contribution right before a **+20%** second half:

```
Money-weighted return   27.13%
Time-weighted return    20.00%
```

TWR is exactly **20%** — the investments returned 0% then +20%, chained, regardless of timing. MWR is **higher** because most of the money was present only for the winning half. That gap *is* the point of having both.

## Tests

- **4 engine tests**: single-flow = holding-period return (10%); the timing-neutralization case above (TWR 20% while MWR is materially higher); `None` for no-flows and single-day spans.
- **1 CLI integration test** asserting TWR = 20.00% and MWR > 25% on the divergence fixture.

45 engine tests + 7 returns integration tests; clippy/doc/fmt clean.

## Scope

True TWR via the NAV method. Still open per #1814: dividend / ex-dividend breakout, and per-group breakdown (#1820).

Closes #1819. Part of #1814.

### DIFF
--- a/crates/rustledger-returns/README.md
+++ b/crates/rustledger-returns/README.md
@@ -22,10 +22,15 @@ re-deriving it.
   `PriceOracle` + end date → the `CashFlow` series, with structural
   internal-transfer exclusion and a terminal market valuation of the position
   still held.
-- [ ] Dividend / ex-dividend breakout (total vs. ex-income return).
-- [ ] Time-weighted return (Modified Dietz / true TWR) — needs per-date
-  portfolio valuations, built on the same realization primitive as
+- [x] `twr` — annualized time-weighted return via the unit-value (NAV) method:
+  values the portfolio at each cash-flow date and chains the sub-period returns,
+  so it measures the investments' performance independent of contribution
+  timing. Shares the `investment_value_at` realization primitive with
   `terminal_value`.
+- [ ] Dividend / ex-dividend breakout (total vs. ex-income return).
+- [ ] Per-commodity / named-group breakdown ([#1820]).
+
+[#1820]: https://github.com/rustledger/rustledger/issues/1820
 
 ## Sign convention
 
--- a/crates/rustledger-returns/src/extract.rs
+++ b/crates/rustledger-returns/src/extract.rs
@@ -511,11 +511,12 @@ fn investment_value_at(
 /// # Returns `None`
 ///
 /// When the return is undefined: no cash flows; the whole span is a single day
-/// (nothing to annualize); or the portfolio value is non-positive at an
-/// *intermediate* valuation point (a mid-stream full liquidation-then-refund, or
-/// a net-short position, where unit accounting across a zero breaks down — a
-/// documented limitation of this MVP). A position simply **closed out on the
-/// report date** is *not* `None`: its holding-period return is reported.
+/// (nothing to annualize); a **net-short** portfolio (negative value, at any
+/// valuation point) where unit accounting is undefined; or a mid-stream full
+/// **liquidation-then-refund** (a zero value between two flows, breaking the
+/// chain) — both documented limitations of this MVP. A position simply **closed
+/// out on the report date** (value zero only at the very end) is *not* `None`:
+/// its holding-period return is reported.
 ///
 /// # Performance
 ///
@@ -601,7 +602,14 @@ pub fn twr(
             }
             cumulative *= r;
         }
-        Some(_) => {} // fully liquidated at the last flow: no final-segment return
+        // Net short (negative value): unit accounting across a non-positive value
+        // is undefined — return None, consistent with the in-loop guard that
+        // returns None for a mid-stream vp <= 0. (Not the same as a clean close.)
+        Some(vp) if vp < 0.0 => return Ok(None),
+        // Exactly zero: fully liquidated at the last flow, so the final span holds
+        // nothing and contributes no return — the whole-period figure is already
+        // in `cumulative` (a position closed out on the report date keeps its TWR).
+        Some(_) => {}
         None => return Ok(None), // unreachable: `flows` non-empty ⇒ the loop ran
     }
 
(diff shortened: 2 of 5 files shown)


robcohen:
## Deep review — round 1

Workflow-backed review (4 finder angles + adversarial verify) on the new TWR math. It found **two genuine correctness regressions** plus doc/test gaps; two findings refuted; one false positive I verified and corrected. All addressed in `6a4fabc0`.

### 1. TWR error aborted the whole report (CONFIRMED — fixed)

TWR values the portfolio at *every* cash-flow date, so it needs more prices than MWR (which only needs them at transaction dates + the end). `twr`'s `MissingPrice` was propagated with `?`, so a price gap at any intermediate flow date failed the **entire** `rledger report returns` — killing the already-computed money-weighted output too. **Fixed:** a missing price now degrades TWR to `n/a`; the headline MWR still renders.

### 2. A closed-out position reported n/a instead of its return (CONFIRMED — fixed)

Buy then sell-all on the report date: the final segment saw `v_prev == 0` and returned `None`, discarding a well-defined ~10% TWR for the very common "what did my sold-out position return" case. **Fixed:** the final segment now skips (adds no return) when the position was closed, so the whole-period figure already accumulated is reported. Verified end-to-end — the closed-out fixture now shows `Time-weighted return 10.03%` (was `n/a`). The regression test also exercises the withdrawal/outflow sign path, which finding 5 flagged as untested.

### Refined / documented

- **Sub-year annualization (CONFIRMED)** — the doc now warns that a short span extrapolates (a +10% month → ~+214%/yr; standard, but surprising).
- **Quadratic realization (CONFIRMED, perf)** — labeled correctness but it's purely performance (`O(flow_dates × directives)`; no wrong answer). Added a `# Performance` note and filed **#1832** for the single-pass linear rewrite (needs a small `BookingEngine` borrowing iterator). Deferred rather than rush a risky rewrite of new math into this PR.
- **`.expect()` panic path** — removed (`first_date` captured in the loop).

### False positive (verified)

The "no drift-guard test exists" finding was incorrect: `terminal_value_matches_account_balances_realization` (added in #1818, on main) values `account_balances`' inventories and compares to `terminal_value`, which now delegates to `investment_value_at` — so the shared realization *is* guarded end-to-end. The finder missed it (it's a unit test in returns.rs, not the tests/ dir). Comment wording made precise.

### Also refuted (agreed)

`to_f64().unwrap_or(0.0)` — unreachable for portfolio-value decimals.

46 engine tests + 7 returns integration tests; clippy/doc/fmt clean.

robcohen:
## Deep review — round 2

Second workflow pass, auditing the round-1 fixes and going deeper. **One real correctness bug** (which I'd flagged as a risk when launching this round), its test gap, plus two cleanups. No re-litigation of round 1. All addressed in `c994faf7`.

### Net-short reported 0.00% instead of n/a (CONFIRMED — fixed)

The round-1 closed-out fix was too broad: the final segment's `Some(_) => {}` skip matched a **negative** (net-short) `v_prev` the same as a fully-liquidated **zero**. A held short position therefore annualized `cumulative = 1.0` into a **fabricated 0.00% TWR** instead of the documented `None` — asymmetric with the in-loop guard that returns `None` for a mid-stream `vp <= 0`.

**Fixed:** the final segment now distinguishes `vp < 0` (net short → `None`, undefined) from `vp == 0` (fully liquidated → skip, keep the accumulated return). Doc aligned. **Test added:** a single net-short flow → `None` (fails on the old skip).

### Dividend-to-bank sub-period untested (CONFIRMED — test added)

The marquee TWR path — crediting income the investment paid out — had no test. Added one that pins the two-sub-period case: 10 AAPL flat but a $50 dividend paid out → sub-period `(1000+50)/1000 = 1.05`, then +10% → chained **15.5%**. Confirms the dividend is credited exactly once (no double-count / mis-scale). The code was already correct; now it's guarded.

### Deferred (cleanup)

The CLI computes `extract_flows` + the end-date valuation once for MWR and again inside `twr`. Distinct from the per-flow-date quadratic, but same family — folded into the `twr` efficiency rework (#1832, noted there).

48 engine tests + 7 returns integration tests; clippy/doc/fmt clean.

robcohen:
Thanks — addressed in `b35a169f`. The three `to_f64().unwrap_or(0.0)` sites (v_i, contribution, v_end) now use `let Some(..) = ..to_f64() else { return Ok(None) }`, so an unrepresentable valuation makes TWR **n/a** rather than a silently-fabricated 0. Behavior is identical for real portfolio decimals; this just hardens the impossible-conversion path. The O(flow_dates × transactions) perf note is tracked in #1832 (single-pass rewrite + CLI dedup). Resolving these threads.

Issue #1832, opened 2026-07-20
returns: single-pass realization for twr + shared per-group extraction (currently ~O(groups × flows × directives))


robcohen:
## Summary

The returns report re-realizes the portfolio from scratch far more than necessary. Two compounding sources:

1. **`rustledger_returns::twr`** re-realizes the whole portfolio at each distinct cash-flow date — a fresh `BookingEngine` pass over every directive `<= date` — so it is **O(flow_dates × directives)**.
2. **The CLI (`report_cmd/returns.rs::compute_group`)**, since #1841, does per group: `extract_flows` + `terminal_value` + `twr` (which re-runs `extract_flows` internally and its D+1 realizations), plus a full transaction scan for the self-containment warning (`first_shared_inscope_account`). With `--by-group` this runs **N+1 times** (N groups + TOTAL), so the path is roughly **O(groups × flow_dates × directives)**.

Correctness is unaffected — this is a scalability follow-up (from the #1830 TWR review, amplified by #1841).

## Fix

1. **Single-pass `twr`**: walk the (date-ordered, booked) directive stream once through one `BookingEngine`, snapshotting market value each time a flow date is crossed → `O(directives + flow_dates × positions)`.
2. **Combined entry point**: a `compute_returns(...)` that walks once and returns `{invested, distributions, terminal, mwr, twr}` together, so the CLI stops calling `extract_flows` twice + `terminal_value` + `twr`'s internal extract per group.
3. **One realization shared across groups**: `BookingEngine` realizes *all* accounts and `investment_value_at` filters by scope afterward, so a single whole-ledger forward pass can serve every group's TWR snapshots (snapshot at the union of all groups' flow dates; each group filters by its own scope). Collapses `O(groups × D × N)` → `O(N + Σflow_dates × positions)`.
4. **Fold in the self-containment scan**: classify per-transaction roles during the same single walk instead of re-scanning per group.

Needs a small primitive: add a borrowing `BookingEngine::inventories(&self) -> impl Iterator<Item = (&Account, &Inventory)>` so the value can be snapshotted mid-walk without consuming the engine (today only `inventory(&account)` and the consuming `into_inventories(self)` exist).

## Correctness caveats (each needs a guard)

[...]

PR #1844, opened 2026-07-21
perf(returns): share one realization across all --by-group scopes

robcohen:
## What

`report returns --by-group` computes one summary for the whole scope plus one per group. After #1843 each is a single-pass `compute_returns`, but that still re-runs the **booking pass** once per scope — even though booking (applying transactions, realizing inventories) is entirely **scope-independent**; only the valuation filters by scope.

This shares **one** realization across all scopes. Adds `compute_returns_multi`, which books the ledger once — over the union of every scope's flow dates — and reuses that single forward pass for every scope's valuation. Per-group booking goes from **`O(scopes × directives)` → `O(directives)`**.

## How

- **`investment_values_multi`** — one `BookingEngine` forward pass; at each union date it values every scope that needs that date (per-scope cursors into each scope's ascending date list, handling a duplicated `end_date`). The engine state at any date is exactly `transactions ≤ date` — identical to what per-scope `investment_values_at` builds — so each scope's values are unchanged. Falls back to per-scope `investment_values_at` on an unsorted stream.
- **`compute_returns_multi`** — preps each scope (`extract_flows`, which may fail independently), runs the one shared realization, then assembles each `Returns`. A scope whose flow can't be priced sits out the shared pass and reports its own error without affecting the others.
- **CLI** — `--by-group` now calls `compute_returns_multi` once for `[TOTAL, ...groups]`.

## Correctness

Each scope's result is **identical** to `compute_returns(that scope)` — the booking is shared, the per-scope flow extraction and valuation are not. Drift guards (the per-scope path is the oracle):
- `compute_returns_multi_matches_per_scope` — asserts `compute_returns_multi[i] == compute_returns(scope_i)` across a happy single-holding scope, a scope whose EUR flow can't be priced (independent error → **skipped from the shared pass**, exercising the alignment), and an income-only scope (zero-capital TWR guard).
- `investment_values_multi_matches_per_scope` — the lower-level pass vs per-scope `investment_values_at`.
- All 19 `report_returns` integration tests pass unchanged.

## Scope

Part of #1832 — this is the shared-realization increment (the last big perf lever). Remaining minor item: the self-containment warning scan (`first_shared_inscope_account`) still walks transactions per group; folding it into the same pass is small and warning-only.

Part of #1832. Part of #1814.

### DIFF
--- a/crates/rustledger-returns/src/extract.rs
+++ b/crates/rustledger-returns/src/extract.rs
@@ -580,6 +580,84 @@ fn investment_values_at(
     values
 }
 
+/// Value SEVERAL scopes at their (each ascending) dates in ONE booking pass over
+/// the whole ledger.
+///
+/// The booking forward pass is **scope-independent** — it applies every
+/// transaction once and realizes every account — so N scopes share a single
+/// realization; only the per-date valuation ([`value_investment_scope`]) filters
+/// by scope. The engine state at any date is exactly `transactions ≤ date`,
+/// identical to what per-scope [`investment_values_at`] would build, so each
+/// scope's values match `investment_values_at` for that scope — but the O(N)
+/// booking is paid once instead of once per scope. Returns, per scope (in input
+/// order), its value at each of its dates. Falls back to per-scope
+/// `investment_values_at` on an unsorted stream, matching that function's own
+/// fallback. Pinned by `investment_values_multi_matches_per_scope`.
+fn investment_values_multi(
+    directives: &[Directive],
+    scoped_dates: &[(&Scope, &[NaiveDate])],
+    reporting_currency: &str,
+    prices: &impl PriceOracle,
+) -> Vec<Vec<Result<Decimal, ExtractError>>> {
+    let transactions = || {
+        directives.iter().filter_map(|directive| match directive {
+            Directive::Transaction(txn) => Some(txn),
+            _ => None,
+        })
+    };
+    if !transactions().is_sorted_by_key(|txn| txn.date) {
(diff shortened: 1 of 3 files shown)


robcohen:
## Deep review → fixes (82cb565a)

The review confirmed the core: **no correctness regression — the multi==per-scope equivalence holds** (union/cursor logic, ready/failed alignment, all branches). Findings were diagnosability, coverage, and docs:

| # | Finding | Resolution |
|---|---------|-----------|
| 3 | No test for the riskiest path (duplicate `end_date == last flow date` → union collapse must still emit both values), nor empty-flows | Added `compute_returns_multi_matches_per_scope_edge_cases`: a scope with a sale ON `end_date` + an empty-flows scope, each asserted == per-scope `compute_returns` |
| 4 | `compute_returns_multi` missing `# Panics`/`# Errors`/`#[must_use]` | Added all three (its `.expect`s are internal alignment invariants, mirroring `compute_returns`) |
| 1 | Whole-portfolio TOTAL error context read "for TOTAL" in the by-group path but "for the scope" elsewhere | `compute_group` now wraps with the row label too — consistent across paths ("...for TOTAL" / "...for Tech") |
| 2 | Group warnings now print to stderr *before* a whole-portfolio pricing error aborts (old path computed TOTAL first) | **Accepted**: inherent to sharing the realization across TOTAL + groups (the total is computed after `build_groups`); benign — more diagnostics, not wrong ones |

Gauntlet green (58 returns + 19 CLI tests, clippy -D warnings, doc, fmt).

PR #1861, opened 2026-07-24
feat(report): realized IRR for capgains (--irr)

robcohen:
## Realized IRR for `report capgains`

Adds `--irr`: the annualized money-weighted return of each **closed** tax lot,
plus a pooled rate per term and per currency.

```text
Sold       Commodity / account        Units  Acquired    Term       Proceeds           Gain       IRR
-----------------------------------------------------------------------------------------------------
2020-12-31 AAA Stock                     10  2020-01-01    ST           1100            100    10.00%
2021-12-31 BBB Stock                      5  2020-01-01    LT            605            105    10.00%
-----------------------------------------------------------------------------------------------------
Short-term    1 disposals   proceeds            1100   gain             100 USD   IRR 10.00%
Long-term     1 disposals   proceeds             605   gain             105 USD   IRR 10.00%
TOTAL       net realized gain             205 USD   IRR 10.00%
```

Both lots read 10% although the second gained 21% in total — the rate is
*annualized*, so 1.21× over two years compounds to 10%/year. That is the thing the
raw gain column cannot tell you.

### Design

- **Per-lot**: each closed lot is a round trip (`-cost_basis` at acquisition,
  `+proceeds` at sale), so it has a well-defined IRR.
- **Aggregate**: the summary rates pool every eligible lot's flows into one series
  and solve once — a money-weighted return over all the capital that actually
  cycled through — rather than averaging per-lot rates.
- **Reuses the canonical solver**: `rustledger_returns::xirr` (365-day basis,
  beangrow-compatible). No IRR solver is re-derived. The text column reuses
  `returns`' `fmt_rate_pct` so the two rate columns cannot drift in formatting.
- **Opt-in**: keeps the default a lean tax report, and leaves the CSV/JSON schema
  unchanged unless `--irr` is passed. Machine output is a **2-decimal percent** in
  an `irr_pct` field (empty / `null` when undefined) — the same unit as `report
  returns`' `money_weighted_return_pct`, so the two reports' rates are directly
  comparable rather than 100x apart. JSON summaries also carry `irr_lots` /
  `irr_lots_total`, the rate's coverage.
- **Bounded reporting**: a very short hold annualizes to an enormous number (a
  one-day 20% gain compounds to ~8e30 %/yr) — too wide for the text column and more
  digits than this project's own decimal type can parse back. Rates above 9999% show
  as `>9999%` in text and are omitted (empty/`null`) in machine output rather than
  emitting a fabricated figure.

### Positioning vs `report returns`

This is a **realized-only** return — it sees only lots you actually closed, so it
is *not* the portfolio's total return (a position you still hold contributes
nothing, however it has performed). `report returns` remains the beangrow-shaped
total-return report (money- **and** time-weighted, including still-held positions
valued at market). This adds the **per-lot** view that `returns` cannot give. The
docs state the distinction and cross-reference both ways.

### Undefined cases

`n/a`, and excluded from the pooled rates: **short sales** (money-in-then-out makes
an IRR unconventional and misleading), lots with **no acquisition date** (AVERAGE
booking merges lots and drops dates), **same-day round trips** (a zero-day span
cannot be annualized), a non-positive cost basis, negative proceeds, and any lot
whose rate does not solve. A **total loss** is *not* excluded — it is exactly
`-100%` at any horizon and stays in the pool, since dropping it would flatter the
aggregate by hiding capital that never came back.

When a bucket's rate covers fewer lots than the line's disposal count, the summary
says so: `IRR 25.00% (1 of 2 lots)`. The rate and the per-lot cells use one
predicate, so a summary can never contradict the rows above it.

### Testing

Verified against an independent bisection XIRR — per-lot and pooled rates agree to
6 decimals (a 1-year 1000→1100 and a 2-year 500→605 both annualize to
10.000000%). Tests cover annualization/compounding, the opt-in default (schema
unchanged), and every undefined case, at both unit and end-to-end CLI level.
Full `rustledger` suite green; clippy + docs clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ ______________________________________________________________________
 | **31 built-in plugins** | Plus Python plugin compatibility via WASI sandbox |
 | **Bank import** | CSV/OFX import with auto-detection, dedup, and categorization |
 | **Investment returns** | Money-weighted (XIRR) and time-weighted returns, robust over imperfect ledgers: a built-in beangrow replacement |
-| **Capital gains** | Realized gains per tax lot, short vs long term, with exact `@@` proceeds and calendar-correct holding periods |
+| **Capital gains** | Realized gains per tax lot, short vs long term, with exact `@@` proceeds, calendar-correct holding periods, and optional per-lot annualized return (`--irr`) |
 
 <details>
 <summary><strong>Comparison with other tools</strong></summary>
--- a/crates/rustledger/src/cmd/report_cmd/capgains.rs
+++ b/crates/rustledger/src/cmd/report_cmd/capgains.rs
@@ -39,7 +39,7 @@
 //! rules beyond the long-term threshold are out of scope (run `rledger check` to
 //! validate the ledger first).
 
-use super::returns::fmt_rate_pct;
+use super::returns::{fmt_rate, fmt_rate_pct, json_rate};
 use super::{OutputFormat, csv_escape, json_escape};
 use anyhow::Result;
 use rust_decimal::Decimal;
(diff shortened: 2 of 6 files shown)


robcohen:
Addressed the Copilot review (commit `2bfb5781`):

**Fixed now:**
- `aggregate_irr` no longer re-solves every lot on each of the four summary cells per currency — it keys off the already-solved `d.irr`. Beyond the redundant Newton/Brent runs, this makes the aggregate use the *literal* value shown in each cell, so a summary rate cannot disagree with the rows above it.
- The "no IRR by default" assertions matched a substring of the whole output (a commodity or payee containing `irr` could flip them). They now assert on the schema: the CSV header and its column count, the JSON keys, and the text table header.
- A stale docs paragraph survived an earlier edit and still claimed CSV/JSON carry "the exact value" and mentioned a `<-9999%` clamp. Corrected: machine output is a 2-decimal percent, rates above 9999% are omitted entirely (empty/`null`), and there is no lower clamp since a round trip cannot lose more than its basis.
- PR description updated to the actual output contract (percent in `irr_pct`, not a raw fraction).

**Already fixed in earlier commits on this branch** (these comments are against superseded revisions):
- The stale "raw fraction" CSV code comment — corrected in `e464708f`.
- `ends_with(',')` proxy assertions — replaced with exact field-index checks in `a9d1854a`/`e464708f`.
- The `0.1` stringify-flakiness concern is moot: output is now a rounded 2-decimal percent (`25.00`), which is deterministic, and the fixtures deliberately avoid 0.10 because that is `xirr`'s Newton seed.

PR #1890, opened 2026-07-28
fix(core): report Decimal overflow instead of panicking the CLI

### LINKED ISSUE #1863: Decimal overflow in Inventory::add panics the CLI on valid ledger input
`Inventory::add` sums position units with `+=`, which panics on `Decimal`
overflow. Because it runs in the loader's booking phase, a ledger that
triggers it aborts **every** command, including `check`.

## Reproduce

```beancount
2024-01-01 open Expenses:Food
2024-01-01 open Assets:Cash

2024-02-10 * "huge"
  Expenses:Food   79228162514264337593543950335 USD
  Assets:Cash    -79228162514264337593543950335 USD

2024-02-11 * "huge again"
  Expenses:Food   79228162514264337593543950335 USD
  Assets:Cash    -79228162514264337593543950335 USD
```

```
$ rledger check big.beancount
thread 'main' panicked at rust_decimal-1.42.1/src/arithmetic_impls.rs:161:18:
Addition overflowed
$ echo $?
101
```

`report balances`, `report income` and `check` all abort identically
(exit 101). The amounts are `Decimal::MAX`, so each transaction balances
and the ledger is otherwise well-formed.

## Where

`crates/rustledger-core/src/inventory/mod.rs:526`

```rust
*self
    .units_cache
    .entry(position.units.currency.clone())
    .or_default() += position.units.number;
```

Backtrace: `Inventory::add` → `BookingEngine::try_apply_posting` →
`BookingEngine::apply` → `run_booking` → `process`.

## Why it matters

CLAUDE.md's checklist requires that ledger input never panic the CLI — a
malformed or hostile file should produce a diagnostic, not an abort. This
is reachable from any file the user can write, and it takes out `check`,
which is the command a user would run to find out what is wrong.

## Suggested fix

Use `checked_add` and surface an overflow as a `LedgerError` naming the
account and currency, rather than aborting. The same treatment was applied
to the budget report's accrual in #1862 (`accrue` returns `Option<Decimal>`
and the report renders `n/a`), so there is a recent precedent for the
shape of the fix.

Found while reviewing #1862; not fixed there because it is in the booking
phase and unrelated to that PR.

robcohen:
Closes #1863.

`Inventory` summed position units with `+=`, which **panics** on `Decimal` overflow. It runs in the loader's booking phase, so a ledger holding amounts near `Decimal::MAX` aborted every command — including `check`, the command a user runs to find out what's wrong with their file:

```
$ rledger check big.beancount
thread 'main' panicked at rust_decimal … "Addition overflowed"
$ echo $?
101
```

Now:

```
error[BOOK]: Arithmetic overflow in adding a position for USD in Expenses:Food:
the amount exceeds the representable range (about 7.9e28)
exit 1
```

## Seven sites, not five

The issue named five in `inventory/mod.rs`. Two more turned up while testing, and **both are multiplications** — fixing only the additions would have moved the panic rather than removed it.

| file | site | op |
|---|---|---|
| `inventory/mod.rs` | units cache in `add` | `+=` |
| `inventory/mod.rs` | merge into existing position | `+=` |
| `inventory/mod.rs` | `rebuild_index` | `+=` |
| `inventory/mod.rs` | `book_value` totals | `+=` |
| `inventory/mod.rs` | `at_cost` units×cost | `*` |
| `cost.rs` | `Cost::total_cost` | `*` ← found by the new test |
| `booking/lib.rs` | `cost_number_weight_generic` | `*`, `+=` ← found by the repro |

The last is in a different crate on the interpolation path, and it's what the issue's **second** repro actually hit — my review claimed that fixture proved `at_cost` reachable, which was wrong. `WeightNum` gains saturating ops; its `BigDecimal` impl is the plain operation, since arbitrary precision has no ceiling, so the precise-residual tier is unaffected.

## Shape

Arithmetic saturates and sets a sticky `Inventory::overflowed()`, propagated by `clone`, serde, and `at_cost`.

*(`at_cost` did **not** propagate it in the first commit — it constructs a fresh `Inventory` rather than deriving one, so a source poisoned by `add` produced a derived inventory reporting clean while holding figures computed from clamped units. Found by probing the claim rather than trusting it; fixed in `9b816c4b` with a test that fails if the propagation is removed. `book_value` returns a map with nowhere to carry a flag — the caller must check the source inventory, which is a real limitation of flag-on-container and is called out below.)* `try_add` turns that into `BookingError::Overflow`, and the booking engine calls it — so the user-reachable path **reports**. A clamped figure presented as exact would be worse than the panic for an accounting tool.

`add` keeps its signature: ~250 call sites, and making it fallible would bury this change. The saturation is never presented as a result on its own — every site that clamps also flags.

`apply_checked` reports overflow while still ignoring a failed *reduction*, whose contract (transaction must be pre-booked) the existing `debug_assert` guards. Overflow isn't a caller-contract violation but a property of the user's numbers, so it belongs in the loader's error list.

New code **E4004**, with spec and docs entries — the `error_codes_documented_in_spec` drift guard requires them, and the exhaustive `title`/`explanation` matches force the rest. That guard working as designed is why they're here.

## Tests

Seven, one per operation, plus:

- ordinary amounts are neither perturbed nor flagged — a false positive would condemn good data;
- negative overflow clamps to `MIN`, not `MAX` — a sign error there would turn a huge liability into a huge asset, and nothing would look wrong.

Sabotage-verified: reverting the units-cache check fails four tests, reverting `total_cost` fails the book-value one.

## Note on the TLA+ spec

`spec/tla/Conservation.tla` is cited directly on `add` and models **unbounded** integers, so it cannot express this failure. The model will not catch a regression here — the tests are the guard. I said so at the call site so the citation isn't mistaken for coverage.

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -633,7 +633,12 @@ impl BookingEngine {
             // `CostSpec::resolve`). `apply`'s callers book first, which fills the
             // inferred currency into `cost_spec.currency`; direct-`apply` tests use
             // explicit-currency fixtures, which need no inference.
-            inv.add(Position::from_posting(units, posting.cost.as_ref(), date));
+            // `try_add`, not `add`: this is the user-reachable path into an
+            // inventory, and an overflow here used to panic and take out every
+            // command including `check` (#1863). The error carries the currency
+            // and is wrapped with the account by the caller.
+            inv.try_add(Position::from_posting(units, posting.cost.as_ref(), date))
+                .map_err(|e| convert_core_booking_error(e, &posting.account))?;
         }
         Ok(())
     }
@@ -655,8 +660,35 @@ impl BookingEngine {
     /// at market** (`rustledger-returns`), never `apply`, so a re-merged
     /// booking-failed transaction cannot over-sell it.
     pub fn apply(&mut self, txn: &Transaction) {
+        let _ = self.apply_checked(txn);
+    }
+
+    /// [`Self::apply`], returning an arithmetic overflow instead of dropping it.
+    ///
+    /// Only overflow is reported. A failed REDUCTION is still ignored here (see
+    /// `apply`'s contract: the transaction must already be booked, and the
+    /// `debug_assert` below catches a violating caller in debug builds).
+    /// Overflow is a different thing — not a caller-contract violation but a
+    /// property of the numbers in the user's file — so it must reach the
+    /// loader's error list rather than abort or vanish (#1863).
+    ///
+    /// # Errors
+    /// [`BookingError::Overflow`] if a posting's amount took an inventory
+    /// total outside `Decimal`'s range.
+    pub fn apply_checked(&mut self, txn: &Transaction) -> Result<(), BookingError> {
+        let mut overflow = None;
         for posting in &txn.postings {
             let reduced = self.try_apply_posting(posting, txn.date);
+            if let Err(e) = &reduced
+                && matches!(
+                    e,
+                    BookingError::Inventory(a)
+                        if matches!(a.error, rustledger_core::BookingError::Overflow { .. })
+                )
+            {
+                overflow.get_or_insert_with(|| e.clone());
+                continue;
+            }
             debug_assert!(
                 reduced.is_ok(),
                 "apply() reduction failed — the transaction must be booked \
@@ -668,6 +700,7 @@ impl BookingEngine {
             // Release builds keep the historical ignore-and-continue behavior.
             let _ = reduced;
         }
+        overflow.map_or(Ok(()), Err)
     }
 
     /// Book and interpolate a transaction.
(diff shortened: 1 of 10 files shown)


robcohen:
Closing unmerged. Two independent review passes plus direct reproduction found this makes things worse, not better: `rust_decimal::MIN == -MAX`, so a clamped positive and a clamped negative sum to exactly zero — the balance residual reads 0, the two-tier check never escalates, and a transaction off by 1e40 USD passes `rledger check` with exit 0 where main panicked. Same for an elided posting, which gets a fabricated interpolated amount. Saturation cannot be made safe here incrementally.

It also only covered 7 of ~15 panicking sites, so #1863's symptom survives on the reduce ladder, residual accumulation, disposal, interpolation, BQL `cost()` and `sum_account_and_subaccounts`.

Full site list, reproducers and constraints for the next attempt are recorded on #1863.

PR #1891, opened 2026-07-29
fix(core): report Decimal overflow instead of panicking or clamping

### LINKED ISSUE #1863: Decimal overflow in Inventory::add panics the CLI on valid ledger input

robcohen:
Closes #1863.

`rust_decimal` is 96-bit with a hard ~±7.9e28 ceiling and its `+`/`*` **panic** on overflow. Inventory and weight arithmetic runs in the loader's booking phase, so a ledger holding large amounts aborted **every** command with exit 101 — including `check`, the command you run to find out what is wrong with your file.

All arithmetic reachable from ledger input is now **checked**. Where a correct answer is available it is computed; where none exists the transaction fails with a diagnostic. Nothing is clamped.

## This is not the saturating fix

[#1890](https://github.com/rustledger/rustledger/pull/1890) saturated and was closed. `Decimal::MIN == -Decimal::MAX` **exactly**, so a clamped debit and a clamped credit cancel to a residual of *precisely zero* — the fast residual reads 0, the two-tier check never escalates, and a ledger off by 1e40 passed `check` with exit 0 where main panicked. For an accounting tool, trading a loud abort for a silently wrong ledger is the wrong direction.

`min_is_exactly_negative_max` pins that property as a test, so anyone reaching for `saturating_*` here meets the reason first.

## Escalate where an exact answer exists

Python beancount has no such limit: its `decimal` context keeps 28 significant digits but `Emax` is 999999, so magnitude is effectively unbounded. rustledger already carries a `BigDecimal` tier for the precise residual check (#1240) — so the `Decimal` tier now reports "cannot represent" and the balance validator **escalates** instead of skipping.

A fully-explicit transaction therefore gets beancount's exact answer:

| fixture | main | #1890 | this PR |
|---|---|---|---|
| off by 1e40 | panic, 101 | `✓ No errors found`, **0** | `E3001 … residual -1e40 USD` |
| two 4e28 postings | panic, 101 | panic, 101 | `E3001 … residual 8e28 USD` |
| elided posting | panic, 101 | **exit 0, fabricated amount** | `E4004`-class, exit 1 |
| sell a large lot | panic, 101 | panic, 101 | reported, exit 1 |

The 1e40 residual is reported as `-10000000000000000000000000000000000000000 USD`, matching Python's `-1.000000000000000000000000000E+40`.

**E4004** is reserved for where a representable result is genuinely required: an interpolated posting amount (it is written *into* the posting and is thereafter indistinguishable from user input), an inventory total, or a cost basis.

## Fallible signatures

`Inventory::add` and friends now return `Result`. #1890 rejected this over "~250 call sites" — that count was ~95% tests. Production sites are about **33**, and roughly half were already inside `Result`-returning functions.

- `add` computes both running totals *before* committing either, so a rejected add leaves the inventory untouched rather than half-applied.
- **`FromIterator<Position> for Inventory` is removed** in favor of `try_from_positions`. `from_iter` cannot report failure, so it had to swallow the overflow and return an inventory holding a wrong total with nothing to indicate it. A `collect()` that can silently lie is worse than no `collect()`.
- `apply` gains an error path for overflow **only**. A failed *reduction* keeps its debug-assert-then-continue contract — that means the caller applied an unbooked transaction, which is a caller bug rather than ledger data. One API, not a parallel `apply_checked`.

## Verification

Every test was **sabotage-verified** — fix reverted, test observed to fail:

- restoring saturation → the 1e40 fixture prints `✓ No errors found` again; caught.
- short-circuiting the escalation → caught by two tests.
- restoring the unchecked `total_cost` in `reduce_from_lot` → caught.

One test was **rewritten after it passed under sabotage**: its fixture tripped the units-cache check before reaching the merge check it claimed to pin. The replacement uses a costed `+MAX` against an uncosted `−MAX` so only the merge overflows, and it now fails when the commit is moved before the check.

The FIFO/LIFO ladder's cost-basis accumulation is checked as **defense-in-depth and documented as such** — no ledger was found that reaches it, because the reducing posting's own weight exceeds the ceiling first and is reported earlier. Removing that check does not make any fixture panic, so the comment says so rather than implying coverage.

## Notes for review

- **No `CACHE_VERSION` bump.** main *panics* everywhere behavior changes, so it can never have written a cache entry holding a wrong answer. Verified by warming the cache with a main-built binary and re-running.
- **Breaking API changes** in `rustledger-core` / `-booking` / `-ops`: `Inventory::{add, merge, at_cost, at_units, book_value}`, `Cost::total_cost`, `sum_account_and_subaccounts`, `calculate_residual`, `cost_number_weight`, `price_weight`, `clamp`/`clamp_indexed`, and the removed `FromIterator`.
- **One known gap, deliberately not fixed here:** the WIT contract (`rustledger:ledger@3.0.0`) returns a bare list from `clamp`, with no error channel. On overflow the FFI degrades to returning the entries *unclamped* rather than emitting a summary built from a clamped total — real directives, never a fabricated opening balance. Surfacing it as a typed error needs a WIT change; worth a follow-up issue.

4279 tests pass · clippy clean at the CI-pinned 1.97.0 · `cargo doc -D warnings` clean.

### DIFF
--- a/crates/rustledger-booking/fuzz/Cargo.lock
+++ b/crates/rustledger-booking/fuzz/Cargo.lock
@@ -710,7 +710,7 @@ checksum = "94300abf3f1ae2e2b8ffb7b58043de3d399c73fa6f4b73826402a5c457614dbe"
 
 [[package]]
 name = "rustledger-booking"
-version = "0.16.5"
+version = "0.21.0"
 dependencies = [
  "bigdecimal",
  "rust_decimal",
@@ -733,7 +733,7 @@ dependencies = [
 
 [[package]]
 name = "rustledger-core"
-version = "0.16.5"
+version = "0.21.0"
 dependencies = [
  "imbl",
  "jiff",
--- a/crates/rustledger-booking/fuzz/fuzz_targets/fuzz_booking.rs
+++ b/crates/rustledger-booking/fuzz/fuzz_targets/fuzz_booking.rs
@@ -166,9 +166,13 @@ fuzz_target!(|input: FuzzTransaction| {
             .with_synthesized_posting(posting)
             .with_synthesized_posting(counter);
 
-        // Ignore errors — we're building up state, some combos may fail
+        // Ignore errors — we're building up state, some combos may fail.
+        // That includes `apply`'s overflow error (#1863): the fuzzer generates
+        // arbitrary units and costs, so an out-of-range product is a legitimate
+        // input to explore past, not a crash. Unwrapping here would abort the
+        // run and report a false finding — the opposite of this target's job.
         if let Ok(result) = engine.book_and_interpolate(&txn) {
-            engine.apply(&result.transaction).expect("fixture fits in Decimal");
+            let _ = engine.apply(&result.transaction);
         }
     }
 
--- a/crates/rustledger-booking/fuzz/fuzz_targets/fuzz_booking.rs
+++ b/crates/rustledger-booking/fuzz/fuzz_targets/fuzz_booking.rs
@@ -168,7 +168,7 @@ fuzz_target!(|input: FuzzTransaction| {
 
         // Ignore errors — we're building up state, some combos may fail
         if let Ok(result) = engine.book_and_interpolate(&txn) {
-            engine.apply(&result.transaction);
+            engine.apply(&result.transaction).expect("fixture fits in Decimal");
         }
     }
 
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -694,6 +694,42 @@ impl BookingEngine {
     /// non-fatal preserves every existing caller's release-build behavior;
     /// only the overflow path is new.
     pub fn apply(&mut self, txn: &Transaction) -> Result<(), BookingError> {
+        // Snapshot every account this transaction touches, so an overflow
+        // partway through cannot leave the earlier postings applied.
+        //
+        // Without this, a transaction the loader then DROPS (`run_booking`'s
+        // `failed_indices`) still moved the running balances, and every later
+        // transaction booked against the corruption — silently wrong reports
+        // and lot matching, which is the outcome this whole change exists to
+        // prevent. `Inventory`'s positions are an `imbl::Vector`, so each
+        // clone is O(1) structural sharing.
+        //
+        // `None` records an account that did not exist yet, which must be
+        // REMOVED rather than restored to an empty inventory: `apply` is the
+        // only thing that creates entries here, and leaving a phantom empty
+        // one changes `into_inventories`' output.
+        let mut snapshot: FxHashMap<rustledger_core::Account, Option<Inventory>> =
+            FxHashMap::with_capacity_and_hasher(txn.postings.len(), Default::default());
+        for posting in &txn.postings {
+            snapshot
+                .entry(posting.account.clone())
+                .or_insert_with(|| self.inventories.get(&posting.account).cloned());
+        }
+        let rollback =
+            |engine: &mut Self,
+             snapshot: FxHashMap<rustledger_core::Account, Option<Inventory>>| {
+                for (account, prior) in snapshot {
+                    match prior {
+                        Some(inv) => {
+                            engine.inventories.insert(account, inv);
+                        }
+                        None => {
+                            engine.inventories.remove(&account);
+                        }
+                    }
+                }
+            };
+
         for posting in &txn.postings {
             let reduced = self.try_apply_posting(posting, txn.date);
             if matches!(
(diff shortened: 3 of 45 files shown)


robcohen:
Thanks — acting on all six.

**Two were live defects:**

- **`fuzz_booking` unwrapping `apply`** — correct and the most valuable of the batch. The comment directly above literally says *"Ignore errors — we're building up state, some combos may fail"*, and my mechanical test-fixup pass appended `.expect(...)` to the line beneath it. Since the fuzzer generates arbitrary units and costs, that converts an input worth exploring past into an abort, and into a false crash report from a target whose whole job is finding real ones. Restored to `let _ =` with a note. (Checked the harness's own `-units * cost` too: `units: u16` and cents at scale 2 cap it near 1.4e12, so that one is safe.)

- **"smaller units" → "larger units"** — right on the substance, and you caught that it contradicted E4004's own explanation text, which already said *"use larger units (thousands, millions)"*. Fixed in all three places; they now read *"split the transaction, or denominate it in larger units (thousands, millions) so the number is smaller"*.

**Two were already fixed** in c2ed1c17, pushed before this review landed: the partial pad mutation in both `pad.rs` and `validators/balance.rs`. Both halves of a pad now apply as a unit — if the source half overflows, the target half is undone. Sabotage-verified: without the undo the new test reports `Balance failed for Assets:A: expected 0 USD, got 40000000000000000000000000000 USD`, a phantom pad satisfying a later assertion.

While re-reading my own comments I also found three claims stronger than the code and corrected them: `OverflowError` said "every reachable site reports this" (several return a plain `Option`), `accumulate_residual` said "when every posting is explicit" (the gate also fires on an unresolved `{}` cost spec), and `apply` asserted "every existing caller relies on" continuing past a failed reduction.

robcohen:
The `book.rs` finding is correct and was the most serious defect in the PR — thank you. Fixed in f8f98ca2.

Confirmed it before fixing: a two-posting transaction whose second posting overflows left `Assets:A = 10.00` in the engine after `apply` returned an error. Since the loader then drops that transaction via `failed_indices`, an entry the user is told was *rejected* had still moved the running balances, and every later transaction booked against it — wrong lot matching and wrong balances presented as fact, which is exactly the outcome this branch exists to prevent.

Took the rollback option you suggested: `apply` snapshots every account the transaction touches and restores them on overflow. `Inventory`'s positions are an `imbl::Vector`, so each clone is O(1) structural sharing. One wrinkle worth noting — accounts that did not exist beforehand are *removed* rather than restored to an empty inventory, since `apply` is what creates those entries and a phantom empty one would change `into_inventories`' output. A failed *reduction* keeps its historical debug-assert-then-continue contract; only the overflow path rolls back.

Sabotage-verified: dropping the rollback fails the new test on "an account first seen in the failed transaction must not survive it".

You also spotted that this is the same defect as the pad partial-application — that was the right connection to draw. I had fixed the pad instance and not gone looking for others of the same shape, which is what let this one through.

The `validators/balance.rs` thread is the pad instance, fixed in c2ed1c17 (both halves now apply as a unit, with the target undone if the source overflows).

Issue #1892, opened 2026-07-29
Is it intentional that `format` strips thousands separators, even with `option "render_commas" "TRUE"`?


ayalavalva:
Hi Rob, we've moved a two-entity plain-text accounting setup (~10k transactions, multi-currency, plus a consolidated group ledger) onto it and it's been excellent.

This is a **question rather than a bug report**: I'd like to understand whether `format`'s handling of thousands separators is a deliberate canonicalization choice, because it differs from Beancount's `bean-format` and from rledger's own `report`/`query`.

## What I see (rledger 0.21.0, Windows x86_64)

`format` strips thousands separators out of amounts, **even when the ledger sets
`option "render_commas" "TRUE"`**:

```beancount
option "operating_currency" "USD"
option "render_commas" "TRUE"

2020-01-01 commodity USD
2020-01-01 open Assets:Bank USD
2020-01-01 open Income:Sales USD

2020-01-02 * "A client" "Big invoice"
  Assets:Bank    1,234,567.89 USD
  Income:Sales  -1,234,567.89 USD
```

```console
$ rledger format --check --diff probe.beancount
@@ line 10 @@
-  Assets:Bank    1,234,567.89 USD
+  Assets:Bank    1234567.89 USD
@@ line 11 @@
-  Income:Sales  -1,234,567.89 USD
+  Income:Sales  -1234567.89 USD
$ echo $?
1
```

It also strips them from `@@` totals, `price` and `balance` lines.

## Why it looks inconsistent

**1. `report`/`query` honour `render_commas` on the very same file.** Same ledger as above, same run:

```console
$ rledger report probe.beancount balsheet
  1,234,567.89 USD  Assets:Bank
```

So within one ledger, `report` renders separators and `format` removes them from the source.

**2. Beancount's `bean-format` preserves them.** It only aligns the decimal column, leaving the numerals alone:

```console
$ python -m beancount.scripts.format probe.beancount
  Assets:Bank                       1,234,567.89 USD
```

Beancount's *parser* accepts separated numerals, and its `printer` (which re-serializes parsed entries) does drop them, so there's arguably precedent either way inside Beancount, but the user-facing `bean-format` tool keeps them.

**3. The formatter already consults display options elsewhere.** #1807 ("formatter honors per-currency display precision") and #1750 ("render numbers through the ledger DisplayContext") suggest display options do reach the formatter, which is why `render_commas` being ignored there reads more like an oversight than a policy. Happy to be told otherwise!

## The question

[...]

Issue #954, opened 2026-04-30
BQL output ignores DisplayContext, breaks per-currency precision


robcohen:
## Summary

`rledger query` text output renders `Value::Number` via `n.normalize().to_string()`, which strips trailing zeros from `Decimal` values. Python `bean-query` formats Decimal output through the file's `DisplayContext`, which preserves the per-currency precision recorded during loading. Result: `SELECT SUM(number)` returns `"USD 0"` where bean-query returns `"USD 0.00"` — values are equal, the rendered precision differs.

The same root issue affects the JSON output path.

Since #938 (parser fix), bare integer literals parse as `Value::Integer`, not `Value::Number`. So this issue only affects fractional Decimals and aggregate results — the impact is bounded by how many BQL queries produce non-integer numeric output.

## Reproducer

```bash
cat > /tmp/x.bean <<'BEAN'
2020-01-01 open Assets:Bank
2020-01-01 open Equity:Open
2020-01-01 * "opening"
  Assets:Bank   100.00 USD
  Equity:Open  -100.00 USD
BEAN

rledger query /tmp/x.bean "SELECT currency, SUM(number) GROUP BY currency"
# rledger:    USD  0
# bean-query: USD  0.00
```

The values are equal; only the rendered precision differs.

## Code pointers

- `crates/rustledger/src/cmd/query/output.rs:223` — `Value::Number(n) => n.normalize().to_string()`
- `crates/rustledger/src/cmd/query/output.rs:356` — JSON path: `Value::Number(n) => serde_json::json!(n.to_string())`. Same root issue, addressed by the same fix.
- `crates/rustledger-core/src/display_context.rs:42` — `DisplayContext` is already plumbed to `format_value` as `ctx`; only the `Number` branch ignores it.
- `DisplayContext::format(number, currency)` already does the right thing when a currency is known. The fix needs to handle the bare-Number case.

## Correcting an earlier claim

An earlier draft of this issue said "bean-query doesn't normalize either; it just inherits whatever DisplayContext-applied stringification produced upstream." That's wrong: bean-query's `display_context.format_decimal` rounds and pads to per-currency precision. The Decimal values reach the renderer pre-formatted to the right scale. So a fix that just drops `normalize()` (option 3 below) doesn't actually match bean-query — it relies on `rust_decimal` happening to preserve scale through the arithmetic, which works for some cases and diverges in others.

## Recommended fix: DisplayContext-driven default precision

[...]

PR #1893, opened 2026-07-29
fix(query,docs): apply render_commas consistently across output surfaces

robcohen:
Answers #1892 and fixes the two inconsistencies that investigating it turned up.

## The question: is `format` stripping separators deliberate?

**Yes, and it is pinned by tests** — not an oversight. `rustledger-ffi-component/src/convert.rs` states the policy outright, and two tests in `rustledger-core/src/format/mod.rs` set `render_commas(true)` and assert the formatter emits none (`number_display_context_pads_without_separators`, `number_display_context_threads_through_directives`). The loader only calls `set_render_commas` on the display path, never the format path.

The reporter's inference from #1807/#1750 was reasonable — display options *do* reach the formatter — but `render_commas` is excluded on purpose while precision is not.

## Two things that were genuinely wrong

**`SUM(position)` ignored the flag while `SUM(number)` honored it**, in the same query and the same output format:

```
SELECT account, sum(number)    ->  Assets:Bank    1,234,567.89
SELECT account, sum(position)  ->  Assets:Bank   1234567.89 USD
```

Per-column `DisplayContext`s start empty, and the ledger context was merged in **only on the first `Value::Number` cell** — that inheritance exists to give naked decimal columns a precision fallback (#954). `render_commas` was riding along as a side effect, so any column holding a Position or Inventory never received it. It is presentation policy for the whole table, so it is now set on every column context directly; the Number-only precision inheritance is untouched.

**CSV emitted separators; JSON did not.**

```
csv  -> Assets:Bank,"1,234,567.89"
json -> "1234567.89"
```

Both are machine interchange. The separator forces the field to be quoted and is then rejected by ordinary decimal parsers — it broke the reporter's reconciliation script. CSV now matches JSON. Precision is untouched; only the separator is suppressed.

## Documentation

`docs/reference/options.md` described `render_commas` with no scope at all, which is why expecting `format` to honor it was reasonable. It now carries a per-surface table and the rationale, including the consequence the reporter identified: a separator-bearing file is permanently "unformatted" to `format --check` (exit 1), so a CI gate needs `rledger format` run once to canonicalize. The parser accepts separators on input either way, so existing files keep loading.

## Verification

| surface | before | after |
|---|---|---|
| `report` | `1,234,567.89 USD` | unchanged |
| `query` text `sum(number)` | `1,234,567.89` | unchanged |
| `query` text `sum(position)` | `1234567.89 USD` | **`1,234,567.89 USD`** |
| `query -f csv` | `"1,234,567.89"` | **`1234567.89`** |
| `query -f json` | `1234567.89` | unchanged |
| `format` | `1234567.89` | unchanged (by design) |

Both tests sabotage-verified — each fails with its fix removed.

With `render_commas` unset (every existing ledger), output is byte-identical to main across text/csv/json for both query shapes; with it set, only the two intended surfaces change. The BQL compat harness does not set the option, so it is unaffected.

4288 tests · clippy clean at the CI-pinned 1.97.0 · `cargo doc -D warnings` clean.

### DIFF
--- a/crates/rustledger-core/src/display_context.rs
+++ b/crates/rustledger-core/src/display_context.rs
@@ -337,11 +337,21 @@ impl DisplayContext {
     /// Only `render_commas` is affected — precision is a property of the data
     /// and is identical on every surface. See [`OutputSurface`] for why the
     /// distinction exists.
+    ///
+    /// Borrows unless the flag actually has to change, so the common cases
+    /// cost nothing: a ledger without `render_commas` (almost all of them) and
+    /// any human-facing surface both borrow. Only suppressing separators for a
+    /// machine or ledger-text surface clones, and that clone carries the
+    /// per-currency histograms — worth avoiding on a REPL's hot path, and
+    /// wasted entirely on the JSON writer, which ignores the context.
     #[must_use]
-    pub fn for_surface(&self, surface: OutputSurface) -> Self {
+    pub fn for_surface(&self, surface: OutputSurface) -> std::borrow::Cow<'_, Self> {
+        if !self.render_commas || surface.renders_thousands_separators() {
+            return std::borrow::Cow::Borrowed(self);
+        }
         let mut ctx = self.clone();
-        ctx.render_commas = self.render_commas && surface.renders_thousands_separators();
-        ctx
+        ctx.render_commas = false;
+        std::borrow::Cow::Owned(ctx)
     }
 
     /// Set the `render_commas` flag.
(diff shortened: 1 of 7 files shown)


robcohen:
## Deep review — found a third affected surface and a design flaw

Pushed as `c978a3b7`.

### `query --format beancount` was putting separators in ledger text

`1,234,567.89 USD` — emitted as beancount source, by the same binary whose `format` command strips separators precisely because ledger text has one canonical form. Pre-existing on main, but the first commit's own documentation turned it into a contradiction.

### `report -f csv` / `-f json` had the identical defect

I fixed `query`'s CSV and stopped there. `report` has its own format enum and eight call sites passing the raw ledger context, so `report -f json` emitted `"amount": "1,234,567.89"` — a separator inside a JSON numeric string, which is the worse of the two.

### The cause was structural

Five places independently decided whether separators apply — the core formatter plus each of the four query writers — with nothing asserting agreement. They had already drifted. Fixing CSV in isolation was treating a symptom.

`OutputSurface` (Human / Machine / LedgerText) now states the rule once and `DisplayContext::for_surface` applies it; both commands resolve the context at their single dispatch point. Each `From<OutputFormat>` is an exhaustive match, so a new output format has to declare which kind of consumer it serves instead of inheriting whatever the previous arm did. `for_surface` narrows only the separator flag — precision is a property of the data and is identical everywhere, which a test asserts.

### One test superseded rather than deleted

`csv_amounts_with_render_commas_are_quoted` (#1750) failed under the change, and it encoded a real earlier review catch: separator-bearing CSV fields must be quoted per RFC-4180 or rows grow extra columns. That quoting was correct — #1892 just showed that a correctly-quoted `"2,357.65"` is still rejected by ordinary decimal parsers. Omitting separators subsumes the concern (no comma, no quoting, no split column), so the test is rewritten with its column-count assertion retained and both issues cited. Deleting it would have been how the original bug comes back.

### Final matrix

| | text | csv | json | beancount |
|---|---|---|---|---|
| `query` | `1,234,567.89` | `1234567.89` | `1234567.89` | `1234567.89` |
| `report` | `1,234,567.89` | `1234567.89` | `1234567.89` | — |
| `format` | — | — | — | `1234567.89` |

Sabotage-verified (mapping `Beancount` to `Human` fails the new test). With `render_commas` unset — every existing ledger — output is byte-identical to main across 20 command/format combinations.

4290 tests · clippy clean at 1.97.0 · `cargo doc -D warnings` clean.

robcohen:
Both review points addressed.

**The `docs/reference/options.md` comment is already fixed.** It landed at 14:38Z against `ce540774` (14:32Z); `c978a3b7` an hour later fixed exactly that — `report`'s CSV/JSON were using the raw ledger context via eight call sites and did emit separators. Worth saying that Copilot found that gap independently, at the same time my own deep review did, which is decent corroboration that it was the real problem rather than a stylistic quibble. Verified on the current head:

```
report -f csv   ->  1234567.89
report -f json  ->  1234567.89
```

So the docs sentence now holds for every machine-readable surface rather than needing to be narrowed.

**The suppressed low-confidence note about `for_surface` cloning was correct**, and is fixed in `567616a3`. It cloned the whole context — per-currency histograms included — on every query and report, even when the flag didn't change, and the JSON writer never reads the context at all. It now returns `Cow` and borrows unless separators actually have to be suppressed, so the two common cases are free: a ledger that never set `render_commas` (almost all of them) and any human-facing surface. A test pins which cases borrow so it can't silently regress.

Behaviour is unchanged across all seven command/format combinations. 4291 tests, clippy clean at 1.97.0, docs clean.

PR #1894, opened 2026-07-29
fix(parser): keep a sign that is separated from its number

robcohen:
Started as the leading-comma bug from #1892's discussion; the root cause turned out to be broader and more likely to bite.

## A sign separated from its number was silently dropped

```beancount
Assets:A  - 7.50 USD      ; booked as +7.50   ← valid beancount
Assets:A  -,123.00 USD    ; booked as +123.00 ← malformed
```

`starts_amount` required the operand to be **adjacent** to the sign. Anything in between — a space, a stray comma — meant no `AMOUNT` node opened at the sign, so the sign became a flat `POSTING` child. Nothing reads flat children, so it vanished:

```
-123.00   -> [POSTING ACCOUNT [AMOUNT MINUS NUMBER CURRENCY]]       ✓
- 7.50    -> [POSTING ACCOUNT [AMOUNT MINUS] [AMOUNT NUMBER CURRENCY]]
-,123.00  -> [POSTING ACCOUNT MINUS COMMA [AMOUNT NUMBER CURRENCY]]
```

The user saw only a downstream "transaction does not balance" — or, in a transaction that balanced for other reasons, nothing at all.

**The space form is the more serious one.** It is legal beancount, it is far likelier to appear in a real file than a stray comma, and the lima fixture `Arithmetic.NumberExprNegative` already contains `- 7.5` and asserts it parses. That fixture only checks for parse *errors*, never the value, so it has been pinning the silently-wrong behaviour.

## Fixes

- **`starts_amount`** looks past whitespace for the operand, and **`emit_amount`** consumes that whitespace *inside* the node. The second half matters: without it the sign opened an `AMOUNT` that closed immediately, producing two sibling amounts and still losing the sign.
- A token stranded between the account and the amount is now **reported** rather than discarded, so `-,123.00` and `,123.00` fail with a parse error naming the actual problem — a thousands separator must be inside the number.

Well-formed grouping is untouched — `1,234,567.89`, `-1,234.00`, `1,234`, `0,123.00` all parse to the same values. The lexer's grouping regex was already strict and is unchanged; every genuinely malformed grouping (`1,23,456`, `1,,234`, `123,`, …) was already rejected and still is.

## Both conversion paths

`orphaned_amount_prefix` (red) and `green_has_orphaned_amount_prefix` (green) answer the same question over the two representations; green bails to red so red owns the diagnostic. Without the green half it would have kept converting the malformed posting silently and red would never run — `fuzz_green_eq_red` is what holds the two together, and it passes.

## CACHE_VERSION 15 → 16

Parser output changes and a stale entry holds the **wrong sign**. Serving one to a fixed binary reproduces the bug on every already-cached ledger, and a flipped sign is silent — nothing downstream would flag it. Verified by warming a cache with a pre-fix binary and confirming the bumped version rejects it.

## Verification

Sabotage-verified: reverting the `starts_amount` lookahead fails the new test. One of my own test expectations was wrong on the first run (`0,123.00` renders normalized as `123.00`, not `0123.00`) — corrected rather than the code.

4293 tests · clippy clean at the CI-pinned 1.97.0 · `cargo doc -D warnings` clean.

Refs #1892


robcohen:
Both inline comments were right and are fixed in `89700213`.

**`convert.rs:1650`** — the comment claimed "the amount (or anything structured) ends the prefix" while the code stops only at `AMOUNT`. The code is what I want: a `COST_SPEC` or `PRICE_ANNOTATION` can legitimately precede the units, and a stray sign or comma between one of those and the amount is just as orphaned as one right after the account. Their own commas live *inside* their nodes, so it can't false-positive on `{100.00 USD, 2020-01-01}` — verified. So I fixed the comment to state that, with the reason.

**`convert.rs:1082`** — correct, and it became wrong precisely *because* I narrowed the guard in `31aa4672`: it now fires on a stray `+`/`-` with no operand, where a message about thousands separators describes nothing the user wrote. Reworded to cover both. (`Assets:A  -` on its own line reaches it — confirmed.)

## The deep review also found a green/red divergence

`is_trivia_kind` counts `NEWLINE` as trivia, so red's trivia check swallowed the newline and `continue`d — which also made the explicit `NEWLINE => break` beneath it **unreachable dead code**. Red therefore scanned past the end of the posting line while the green mirror stopped at it.

I could not construct an input where that changes an answer (a POSTING's remaining direct-child tokens after the newline are trivia, and its metadata is wrapped in `META_ENTRY` nodes), so this is a latent divergence rather than a live bug. Fixed regardless: the two paths are held together only by `fuzz_green_eq_red` happening to find a distinguishing input, and a predicate that disagrees *by construction* is one input away from a real one.

## One inconsistency I am NOT fixing here

`-,7.50` is now rejected in a posting but still accepted in a `balance` directive, which reads it as `-7.50`:

```
balance Assets:A  -,7.50 USD   ->  expected -7.50 USD   (accepted)
  Assets:A        -,7.50 USD   ->  parse error          (this PR)
```

Balance/price take a different path (`node_has_minus_before_number`), which scans for a `MINUS` before the first `NUMBER` without caring what sits between. That path never had the silent sign loss this PR fixes — it reads the sign correctly — so what remains there is only a strictness gap, and it predates this PR. Extending the guard to it means a different node shape and a different risk profile, so I would rather do it deliberately than bolt it on here. Flagging it as a follow-up candidate.

4293 tests · clippy clean at 1.97.0 · `cargo doc -D warnings` clean.

PR #1895, opened 2026-07-29
fix(parser): stop reading balance/price values as "the first NUMBER token"

robcohen:
The follow-up flagged on #1894. The gap was **wider than the `-,7.50` case that prompted it** — two of these three were silently wrong, all with exit 0:

```beancount
balance Assets:A  (1 + 5) / 2.1 USD   ; asserted against 1, not 2.857…
balance Assets:A  1,23,4.50 USD       ; asserted against 1 — a 1000× error
price   HOOL      -,7.50 USD          ; accepted (sign right, shape wrong)
```

Both directives fall back to "take the first `NUMBER` token and apply a leading sign" when the value isn't an evaluable expression, and that fallback read **only the first token**. Everything after it was discarded without a word. Postings have rejected the same shapes all along — a split numeral never forms one `NUMBER` token, and the lexer's grouping regex is strict.

## Parenthesized arithmetic never evaluated

This is the one that matters most, and it had nothing to do with commas.

`directive_arithmetic_value` skipped the directive header with `skip_while(kind != NUMBER)` — which also swallowed a leading `(`. So `(1 + 5) / 2.1` became `1 + 5 ) / 2.1`, failed to parse, and fell through to the first number:

```
(1 + 5)        ->  1        (should be 6)
(1 + 5) / 2.1  ->  1        (should be 2.857…)
6 / 2.1        ->  2.857…   ✓ (no parens, so it worked)
```

It now skips to the first token that can *begin* a value — `NUMBER`, `L_PAREN`, or a sign. No header token can be any of those, so it cannot over-skip.

## Malformed values are reported

`malformed_directive_value` refuses a value carrying more than the single signed number these directives can represent — extra `NUMBER` tokens, or a misplaced `,`. Shared by both call sites so they can't drift.

**Deliberately not flagged:** a `~ tolerance` clause, whose second number is legitimate (the scan stops at the `TILDE`), and arithmetic, which is evaluated before the guard is consulted.

## Verification

Every legal shape is pinned: `7.50`, `-7.50`, `- 7.50`, `1,234.50`, `0.25 + 0.75`, `(1 + 5) / 2`, `100.00 ~ 0.05`.

Sabotage-verified twice — reverting the header skip fails both new tests; removing the malformed guard fails the truncation test alone.

**CACHE_VERSION 16 → 17.** Cached directives hold the wrong *values*, and the malformed cases also hold an empty `errors` list — both silent. Verified by warming a cache with a pre-fix binary (`expected 1 USD`) and confirming the bumped version re-parses it (`expected 6 USD`).

Two baseline lines drift, both `(1 + 5) / 2.1` fixtures. **Hand-edited, not regenerated** — `BASELINE_UPDATE=1` rewrites against the local corpus, which differs from CI's (it added 21 unrelated entries when I tried it on #1894). Their source hashes are unchanged, confirming the fixtures themselves didn't move.

4295 tests · clippy clean at the CI-pinned 1.97.0 · `cargo doc -D warnings` clean.

Refs #1892


robcohen:
Both points correct, fixed in `274e8027` — and the second one caught a message I shipped two PRs ago.

**The guard's trigger set is wider than its message said.** It fires on extra `NUMBER` tokens as well as stray commas, so malformed arithmetic drew a thousands-separator message that described nothing the user wrote:

```
(1 + 2   -> "a thousands separator must be inside the number"
1 2      -> "a thousands separator must be inside the number"
```

Now: *"expected one number, optionally signed, or an arithmetic expression. A thousands separator must be inside the number, as in `-1,234.00`"*.

**"Attached" was outright wrong**, and you're right that spaced signs are valid — `balance Assets:A - 7.50 USD` parses as `-7.50`. What makes that notable is that #1894 is what *made* it valid, and I wrote "must be attached to a number" in that same PR. So the posting-level message merged on main carries the same error; since I was already editing this file I corrected it there too — *"must be **followed** by a number"*. Thanks for catching a wording bug that had already shipped.

No behavior change. 4295 tests, clippy clean at 1.97.0, and the parser baseline is unaffected — it hashes error kinds and spans rather than message text, which the baseline test staying green confirms.

PR #1913, opened 2026-08-01
fix(booking): solve an elided units number from the cost or price

### LINKED ISSUE #1911: cannot interpolate a missing units number when a cost is present
rledger interpolates a missing units number correctly on its own, but not when the posting carries a cost. beancount solves `units = residual / cost_per_unit`; rledger leaves the posting unfilled and then reports the transaction as unbalanced.

Found by widening the booked-value comparison (#1910) to files whose beancount errors are non-fatal — this file's only errors are `ValidationError` (accounts never opened), so its values are perfectly comparable.

## The three related vectors

All three are listed in `spec/impl/test-vectors.md`.

**`IncompleteInputs.UnitsMissingNumber` — works**

```beancount
2010-05-28 *
  Assets:Account1     100.00 USD
  Assets:Account2            USD
```
Both fill `Assets:Account2 = -100.00 USD`.

**`IncompleteInputs.UnitsMissingNumberWithCost` — BROKEN**

```beancount
2010-05-28 *
  Assets:Account1            HOOL {300.00 USD}
  Assets:Account2    -600.00 USD
```

| | `Assets:Account1` |
|---|---|
| beancount | `2 HOOL {300.00 USD}` — i.e. `600.00 / 300.00` |
| rledger | not booked at all |

rledger then reports:

```
E3001  Transaction does not balance: residual -600.00 USD
```

The residual is the *whole* other leg, which is the tell: the HOOL posting contributed nothing because its number was never filled.

**`IncompleteInputs.UnitsMissingNumberWithPrice` — ambiguous, not claimed as a bug**

```beancount
2010-05-28 *
  Assets:Account2            CAD @ 1.2 USD
  Assets:Account1     100.00 USD @
```

beancount rejects the transaction outright; rledger keeps `Assets:Account1 100.00 USD`. Both refuse to book it meaningfully and the input has a dangling `@`, so this looks like divergent error recovery on genuinely invalid input rather than a booking bug. Noting it for completeness.

## Scope

The missing rule is specifically "units number absent, cost present": the number is recoverable as `residual / cost.number`. The equivalent with a price already works (`10 STK @ 3 USD` fills the cash leg), so this is a gap in one branch rather than a missing feature.

## Not silent

Worth saying plainly: rledger refuses rather than booking something wrong, which puts this well below #1908 and #1909 in severity. A user gets an error, not a bad number.

robcohen:
## What

A posting whose units number is elided beside a **per-unit cost** is now solved as `residual / cost_per_unit` and written in the posting's own commodity. The same inversion covers a per-unit **price**, which had the identical gap.

```beancount
2010-05-28 *
  Assets:Account1            HOOL {300.00 USD}
  Assets:Account2    -600.00 USD
```

| | `Assets:Account1` |
|---|---|
| before | nothing booked; `E3001 residual -600.00 USD` (the whole other leg) |
| after | `2 HOOL {300.00 USD}` |

## Why it was broken

One wrong premise, not a missing feature. A units-missing posting was registered under its **units** currency, but a posting carrying a cost weighs in the **cost** currency. It therefore absorbed a residual that was none of its business. That fabricated lots rather than merely failing to fill them, which makes this worse than the issue reported:

| input | before | after |
|---|---|---|
| `A HOOL {300.00 USD}` / `B -600.00 USD` / `D 5.00 HOOL` | `A = -5.00 HOOL` — wrong sign **and** magnitude, taken from `D`'s residual | `A = +2.00 HOOL` |
| `A {300.00 USD}` / `B -600.00 USD` | `A = 600.00 USD {300.00 USD}`, an invented lot whose imbalance was then reported as an invented `179400.0000 USD` | refused, with a reason |

Registering the unknown under the weight currency also fixes the count rule: a cost posting now collides with a plain elided posting in that same currency, exactly as beancount's `InterpolationError` does.

## What is deliberately refused

Only a **per-unit** factor is inverted. A total cost `{{600 USD}}`, a total price `@@`, a zero factor, an empty `{}` and a compound `{a # b}` are refused with a reason naming the shape. The first four contribute a weight that does not vary with the units at all, so the input is genuinely ambiguous rather than merely hard; the compound form admits two roots and I would rather refuse than pick one.

These **must** error rather than be left unfilled. `validate_transaction_balance` deliberately skips a transaction containing an unfilled posting, on the documented assumption that interpolation already reported the real failure. An early draft of this PR left them unfilled and silently turned four erroring cases green.

## Not copied from beancount

beancount crashes on much of this space rather than diagnosing it:

- `TypeError: bad operand type for abs(): 'type'` on a total cost, a zero cost
- `AssertionError: Internal error; residual currency different than missing currency` on a zero price

So the rule here is derived from the weight algebra in `cost_number_weight_generic` rather than from beancount's output. Where beancount does produce an answer, this agrees with it.

## Verification

- **Value oracle**, 560-file comparable corpus: **4 divergences to 3**, the one removed being `test-cases_IncompleteInputs.UnitsMissingNumberWithCost` (`beancount=2 rledger=None`). No new divergence, identical eligibility (560 compared / 198 skipped on both sides).
- **All 7 new tests confirmed to fail against the pre-fix logic** (neutered `units_weight` to its old behavior, watched them go red, restored). The 120 pre-existing tests stayed green under the neuter, so they pin new behavior without over-constraining the old.
- `solved_units_reproduce_the_canonical_cost_weight` is the drift guard: it feeds the solved number back through the canonical `cost_number_weight` and asserts it cancels the cash leg, so the inverse cannot drift from the forward function.
- Workspace: **4411 passed, 0 failed**. Clippy clean under the CI pin (1.97.0) with `-D warnings`. `cargo fmt` clean.
- Three of the four official `IncompleteInputs.*WithCost` lima vectors improve; the fourth is unchanged.

## Two separate defects found, filed not fixed here

1. **#1914** — E3002 rejects two elided postings in *different* currency groups. The validator builds a per-currency map and then sums it, discarding the grouping. `interpolate` itself is correctly per-currency, so the validator shadows it.
2. **#1915** — a bare price sigil (`100.00 USD @`) is treated as "no price" instead of an unresolved unknown. This PR makes it newly visible: `IncompleteInputs.UnitsMissingNumberWithPrice` went from rejected-for-the-wrong-reason to accepted, because beancount counts that bare `@` as a second unknown in the group and we do not. The parse is unchanged by this PR (main reads it the same way); only the solve improved.

Closes #1911

### DIFF
--- a/crates/rustledger-booking/src/interpolate.rs
+++ b/crates/rustledger-booking/src/interpolate.rs
@@ -62,7 +62,9 @@ pub enum InterpolationError {
 
     /// A posting's units number is missing and cannot be recovered from the
     /// balance, because the posting's weight does not vary with that number
-    /// (#1911). See [`UnitsWeight::Undetermined`] for the cases.
+    /// (#1911). The cases are enumerated on the private `UnitsWeight` classifier
+    /// in this module: a total cost or price, a zero factor, an empty `{}` spec,
+    /// a compound cost, and a cost with no units currency to write into.
     ///
     /// Reported rather than left unfilled: `validate_transaction_balance` skips
     /// a transaction that still has an unfilled posting, so staying silent here
--- a/crates/rustledger-booking/src/interpolate.rs
+++ b/crates/rustledger-booking/src/interpolate.rs
@@ -60,6 +60,21 @@ pub enum InterpolationError {
         candidates: String,
     },
 
+    /// A posting's units number is missing and cannot be recovered from the
+    /// balance, because the posting's weight does not vary with that number
+    /// (#1911). See [`UnitsWeight::Undetermined`] for the cases.
+    ///
+    /// Reported rather than left unfilled: `validate_transaction_balance` skips
+    /// a transaction that still has an unfilled posting, so staying silent here
+    /// would accept an unbalanced transaction.
+    #[error("cannot interpolate the units number for the {account} posting: {reason}")]
+    UnsolvableUnits {
+        /// The account of the posting whose units could not be solved.
+        account: rustledger_core::Account,
+        /// Why the balance does not determine the number.
+        reason: &'static str,
+    },
+
     /// Cannot infer currency for a posting.
     #[error("cannot infer currency for posting to account {account}")]
     CannotInferCurrency {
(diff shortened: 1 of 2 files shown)


robcohen:
Superseded, not abandoned: this branch's two commits are already on main.

#1918 was stacked on this branch, so its squash (`45cc9b961`) carried both changes in together, and this PR is now `DIRTY` against a main that already has its content. Verified on main:

```
$ rledger query i1911.beancount "SELECT account, units, cost_number"
Assets:Account1  2 HOOL   300.00
Assets:Account2  -600.00 USD
```

`UnsolvableUnits`, `scaled_missing` and `elided_unknown_groups` are all present in `crates/rustledger-booking/src/interpolate.rs` on main.

My mistake in the stacking: I armed auto-merge on both, and the child cleared its checks first while this one was still in mutation testing. Next time the parent should merge before the child is armed.

Issue #1966, opened 2026-08-05
WEIGHT() ignores the price annotation, returning units in the wrong currency


robcohen:
Found by the deep review of #1965 (which fixes the adjacent scale problem, #1963).

The canonical weight ladder in `rustledger_booking::posting_weight` has three rungs:

1. cost, if the posting has one
2. **price annotation**, if it has one
3. units

`WEIGHT()` in `crates/rustledger-query/src/executor/mod.rs` implements rung 1 and rung 3 and skips rung 2 entirely — its no-cost branch is `Ok(Value::Amount(p.units.clone()))`.

## Reproduction

```beancount
2024-01-01 open Assets:Bank
2024-01-01 open Assets:Cash
2024-02-01 * "priced, no cost"
  Assets:Bank    10 EUR @ 1.10 USD
  Assets:Cash   -11.00 USD
```

```
$ rledger query p.beancount "SELECT account, position, weight, WEIGHT(position)"
account      position    weight      WEIGHT
-----------  ----------  ----------  ----------
Assets:Bank  10 EUR      11.00 USD   10 EUR
Assets:Cash  -11.00 USD  -11.00 USD  -11.00 USD
```

`weight` says `11.00 USD`. `WEIGHT(position)` says `10 EUR`.

This is worse than #1963: that was the same value carrying invented scale, this is a **different number in a different currency**. Anything summing `WEIGHT()` across a ledger with priced postings gets a silently wrong, currency-mixed total.

## Why it is structural

`rustledger_core::Position` is `{ units: Amount, cost: Option<Cost> }` — there is no price field, and `evaluate_function_on_values` receives only `Value`s, so there is no `Posting` to recover the annotation from. `WEIGHT()` *cannot* be made correct without a contract change. That is the same root cause as #1963 and why I am filing rather than folding it into #1965.

## Options

1. **Carry the price on the query-side position value.** Correct, and closes #1963's remaining scale gap too (a position that retained its cost spec could call the canonical outright). Widest change.
2. **Error on a priced position** rather than return a wrong-currency amount. Cheap and honest; breaks any query relying on the current fallthrough.
3. **Document the limitation.** Cheapest, but leaves a silently wrong number in a function whose whole purpose is to report a weight.

Leaning 1: it is the only option that makes `WEIGHT()` and the `weight` column the same computation, which is what CLAUDE.md's canonical-function rule asks for. 2 is a reasonable stopgap if 1 is not scheduled soon.

[...]

Issue #1963, opened 2026-08-05
[Bug]: BQL `WEIGHT()` re-derives the weight and disagrees with the `weight` column


robcohen:
Found by the canonical-function drift sweep (#1902 Phase 2, sweep 1).

CLAUDE.md names this exact pair: *"Consumers surfacing a per-posting weight (BQL `weight` column, `currency_accounts` grouping) MUST use this rather than re-derive the ladder, or they drift from `rledger check` on those shapes."*

The **column** obeys that. The **function** does not.

| site | implementation |
|---|---|
| `weight` column | `compute_posting_weight` → `rustledger_booking::posting_weight` ✓ |
| `WEIGHT()` function | `p.units.number * cost.number` — re-derived |

## Reproduction

```beancount
2024-01-01 open Assets:Broker
2024-01-01 open Assets:Cash
2024-02-01 * "total cost that does not divide evenly"
  Assets:Broker   3 HOOL {{100.00 USD}}
  Assets:Cash  -100.00 USD
```

```
SELECT account, weight, WEIGHT(position) AS w2
```

| account | `weight` | `WEIGHT(position)` |
|---|---|---|
| Assets:Broker | `100.00` | **`100.00000000000000000000000000`** |
| Assets:Cash | `-100.00` | `-100.00` |

Same value, different scale. `{{100.00 USD}}` resolves to a per-unit of `33.333…`, and multiplying that back by 3 accumulates 26 digits of spurious precision. The canonical avoids it by taking the preserved total rather than recomputing — the precision-loss issue #1106/#1113 fixed, reintroduced in the function path.

## Why it matters beyond aesthetics

- The **same posting reports two different weights** depending on whether you select the column or call the function. That is an internal inconsistency a user can hit in one query.
- The compat suite compares *rendered* output, so a scale difference is a real divergence from beanquery, not a cosmetic one.

## Note for whoever fixes it

Not a one-line swap. `WEIGHT()` receives a `Value::Position`, whose `cost` is already **resolved** to a per-unit `Decimal` — the original total is gone by then, which is precisely why recomputing loses it. The column path works because it sees the `Posting` and its `CostNumber`, where `Total` / `PerUnitFromTotal` still carry the preserved total.

So the fix is either to give `WEIGHT()` access to the posting, or to have `Position` retain enough of the cost spec to call `cost_number_weight`. Worth deciding deliberately rather than papering over the scale.

Issue #1902, opened 2026-07-31
Correctness review: verification coverage is inversely correlated with bug density


robcohen:
Tracking issue for a correctness review of the whole codebase, scoped by evidence rather than by file count.

## The finding that should drive it

Verification effort is inversely correlated with where bugs actually land. Fix commits over the last 6 months (650 total), normalised by source lines:

| crate | fixes | src loc | **fixes/kloc** | verification covering it |
|---|---:|---:|---:|---|
| rustledger-ffi-wasi | 25 | 1,905 | **13.1** | fuzz |
| **rustledger-loader** | 68 | 7,108 | **9.6** | mutation |
| rustledger-query | 103 | 16,026 | **6.4** | fuzz (miri excluded — rowan) |
| rustledger-validate | 28 | 5,998 | **4.7** | mutation |
| rustledger-booking | 29 | 7,220 | 4.0 | miri, kani, tla, mutation, fuzz |
| rustledger-wasm | 27 | 7,616 | 3.5 | — |
| rustledger-core | 57 | 19,425 | 2.9 | miri, kani, tla, mutation |
| rustledger-parser | 52 | 18,387 | 2.8 | miri, fuzz, mutation |
| rustledger-lsp | 63 | 23,977 | 2.6 | — |
| rustledger-plugin | 21 | 15,768 | 1.3 | mutation |

`core` and `booking` carry four or five tools each and have the lowest density. `loader` carries one and has nearly four times `core`'s. That inversion is the target, not any particular file.

**Caveat on the metric.** Fix-commit counts partly measure *attention*, not just defects — `query` may rank high because the BQL compat suite exercises it hard. Treat the ranking as a prior, not ground truth.

**Second gap.** These crates have no `tests/` directory, though they do have unit tests in `src/`:

| crate | src loc | `#[test]` in src/ | integration tests |
|---|---:|---:|---|
| rustledger-ops | 4,636 | 108 | none |
| rustledger-returns | 3,450 | 64 | none |
| rustledger-budget | 3,901 | 58 | none |
| rustledger-ffi-component | 3,449 | 30 | none |
| rustledger-completion | 1,067 | 33 | none |

## Phase 0 — restore the instruments

Adding findings while existing automation sits dark is misallocated effort.

- [x] miri: produced nothing for 8 weeks (4 real failures, then 4 six-hour timeouts). Fix in #1901.
[...]

PR #1982, opened 2026-08-07
test(query): algebraic properties, and a generator that reaches boundaries

robcohen:
#1902 Phase 1 — the last uncovered crate. Miri excludes `rustledger-query` for
rowan reasons that are not going away, so this is the compensating coverage.

## Three algebraic properties

The existing properties ask whether the same input gives the same answer twice.
These ask whether the answer is **right**, using relationships that hold whatever
the ledger contains — so they need no expected value, which is what makes them
checkable over generated input.

| property | what it catches |
|---|---|
| `WHERE p` and `WHERE NOT p` **partition** the unfiltered rows | a predicate that misevaluates in one direction only — the union stops reconstructing the whole |
| grouping **preserves the total** | the #1177 shape: rows right, group-key sidecar wrong |
| `ORDER BY` is a **permutation** | a comparator that is not a total order losing or duplicating rows |

Aimed where this crate actually breaks: #1177 was a grouping/parallel-path
divergence, and the two bugs found by hand this cycle (#1963, #1966) were both
in evaluation rather than parsing. The neighbouring null-ordering tests pin
*which* order; none pinned that the multiset survived it.

## The generator was the more interesting finding

Sabotaging the partition property is how it surfaced. Swapping `number > 0` for
`number > 1` — a filter that should lose every row between them — **changed
nothing.**

With a uniform `-1_000_000..1_000_000`, landing in `(0, 1]` needs
`|n| <= 10^scale`: at most 1000 of two million draws, roughly **0.3 expected
hits across an entire run**. There was nothing between 0 and 1 to lose.

The amount strategy now spans three magnitude bands, and that same sabotage
fails. This affects the pre-existing determinism properties too, which were
sampling the same narrow band.

## Sabotage-verified

| sabotage | result |
|---|---|
| the two halves overlap | partition fails |
| compare against a filtered total | grouping fails |
| drop a row from the ordered side | permutation fails |
| `number > 1` (boundary) | fails **now**; passed before the generator fix |

Worth noting the third and fourth: my first two attempts at sabotaging the
partition were *no-ops* — one because the generator filters zero amounts, one
because it never produced boundary values. A sabotage that doesn't fail is
either a bad sabotage or a real gap, and telling those apart is where the
generator weakness came from.

`cargo test --workspace` — 132 suites, 0 failures. Clippy clean.

### DIFF
--- a/crates/rustledger-query/tests/pipeline_invariants.rs
+++ b/crates/rustledger-query/tests/pipeline_invariants.rs
@@ -65,13 +65,24 @@ fn results_match(a: &QueryResult, b: &QueryResult) -> bool {
 
 /// A balanced two-posting transaction with a random leg account, amount,
 /// and day.
+/// Magnitudes spanning three scales, not one uniform range.
+///
+/// The original `-1_000_000..1_000_000` made a value near the boundary
+/// vanishingly rare: landing in `(0, 1]` needs `|n| <= 10^scale`, so at most
+/// 1000 of two million draws — roughly 0.3 expected hits across a whole run.
+/// A predicate bug at the boundary would have gone unseen, which sabotaging
+/// the partition property is how I found out: swapping `number > 0` for
+/// `number > 1` changed nothing, because there was nothing between them.
+fn amount_strategy() -> impl Strategy<Value = i64> {
+    prop_oneof![
+        3 => -1_000_000i64..1_000_000,  // ordinary magnitudes
+        2 => -1_000i64..1_000,          // small — puts values around 1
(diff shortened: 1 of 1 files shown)


robcohen:
Superseded by #1983 — same commits plus Copilot's per-key grouping assertion, on `chore/query-properties`. This branch was `test/`, which the branch-name gate rejects (allowed: feat, feature, fix, docs, chore, refactor, release, hotfix, perf, claude, dependabot, copilot). Renaming a branch closes its PR anyway, so reopening under the right prefix was the cheaper move.

PR #1983, opened 2026-08-07
test(query): algebraic properties, and a generator that reaches boundaries

robcohen:
#1902 Phase 1 — the last uncovered crate. (Supersedes #1982, which was opened on a `test/` branch that the branch-name gate rejects; the policy allows feat/fix/chore/docs/refactor/perf/release/hotfix.)

Miri excludes `rustledger-query` for
rowan reasons that are not going away, so this is the compensating coverage.

## Three algebraic properties

The existing properties ask whether the same input gives the same answer twice.
These ask whether the answer is **right**, using relationships that hold whatever
the ledger contains — so they need no expected value, which is what makes them
checkable over generated input.

| property | what it catches |
|---|---|
| `WHERE p` and `WHERE NOT p` **partition** the unfiltered rows | a predicate that misevaluates in one direction only — the union stops reconstructing the whole |
| grouping **preserves the total** | the #1177 shape: rows right, group-key sidecar wrong |
| `ORDER BY` is a **permutation** | a comparator that is not a total order losing or duplicating rows |

Aimed where this crate actually breaks: #1177 was a grouping/parallel-path
divergence, and the two bugs found by hand this cycle (#1963, #1966) were both
in evaluation rather than parsing. The neighbouring null-ordering tests pin
*which* order; none pinned that the multiset survived it.

## The generator was the more interesting finding

Sabotaging the partition property is how it surfaced. Swapping `number > 0` for
`number > 1` — a filter that should lose every row between them — **changed
nothing.**

With a uniform `-1_000_000..1_000_000`, landing in `(0, 1]` needs
`|n| <= 10^scale`: at most 1000 of two million draws, roughly **0.3 expected
hits across an entire run**. There was nothing between 0 and 1 to lose.

The amount strategy now spans three magnitude bands, and that same sabotage
fails. This affects the pre-existing determinism properties too, which were
sampling the same narrow band.

## Sabotage-verified

| sabotage | result |
|---|---|
| the two halves overlap | partition fails |
| compare against a filtered total | grouping fails |
| drop a row from the ordered side | permutation fails |
| `number > 1` (boundary) | fails **now**; passed before the generator fix |

Worth noting the third and fourth: my first two attempts at sabotaging the
partition were *no-ops* — one because the generator filters zero amounts, one
because it never produced boundary values. A sabotage that doesn't fail is
either a bad sabotage or a real gap, and telling those apart is where the
generator weakness came from.

`cargo test --workspace` — 132 suites, 0 failures. Clippy clean.

## Second commit — Copilot's point on #1982

The total alone is blind to the failure #1177 actually was. Permute the group
sums across the account labels and the total still reconciles perfectly; every
group is simply attributed to the wrong key. So each group is now also pinned
against the ungrouped rows carrying that key.

Sabotage-verified by doing exactly that — rotating each group's sum onto the
next key. The total assertion stays green; only the new per-key one fires.

### DIFF
--- a/crates/rustledger-query/tests/pipeline_invariants.rs
+++ b/crates/rustledger-query/tests/pipeline_invariants.rs
@@ -65,13 +65,24 @@ fn results_match(a: &QueryResult, b: &QueryResult) -> bool {
 
 /// A balanced two-posting transaction with a random leg account, amount,
 /// and day.
+/// Magnitudes spanning three scales, not one uniform range.
+///
+/// The original `-1_000_000..1_000_000` made a value near the boundary
+/// vanishingly rare: landing in `(0, 1]` needs `|n| <= 10^scale`, so at most
+/// 1000 of two million draws — roughly 0.3 expected hits across a whole run.
+/// A predicate bug at the boundary would have gone unseen, which sabotaging
+/// the partition property is how I found out: swapping `number > 0` for
+/// `number > 1` changed nothing, because there was nothing between them.
+fn amount_strategy() -> impl Strategy<Value = i64> {
+    prop_oneof![
+        3 => -1_000_000i64..1_000_000,  // ordinary magnitudes
+        2 => -1_000i64..1_000,          // small — puts values around 1
+        1 => -5i64..5,                  // tiny — puts values either side of 0
+    ]
+}
+
 fn txn_strategy() -> impl Strategy<Value = Transaction> {
-    (
-        1u32..28,
-        0usize..ACCOUNTS.len(),
-        -1_000_000i64..1_000_000,
-        0u32..3,
-    )
+    (1u32..28, 0usize..ACCOUNTS.len(), amount_strategy(), 0u32..3)
         .prop_filter("non-zero amount", |(_, _, n, _)| *n != 0)
         .prop_map(|(day, acct, n, scale)| {
             let amt = Decimal::new(n, scale);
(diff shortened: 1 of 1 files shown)


robcohen:
## Self-review — the sidecar assertion was still testing the wrong thing

Pushed a fix. My response to Copilot's comment added a per-key assertion that compared `group[0]` — the projected `account` **column** — against the ungrouped rows. But the group key travels separately, in `row_group_keys`, reachable only through the public `QueryResult::group_key(i)`. A row whose column is right and whose sidecar is wrong satisfies every assertion the test had, and that is precisely the divergence #1177 shipped.

Verified rather than argued. Rotating `group_key()` onto the next row, leaving every projected column untouched:

| | rotated sidecar | dropped sidecar |
|---|---|---|
| previous version of this test | **passes** | — |
| with the sidecar assertion | fails | fails |

So Copilot's point stood after my first answer to it, and only the direct `group_key(i) == group[..1]` assertion actually closes it.

Worth naming why the existing coverage didn't already cover this: `results_match` next door *does* compare sidecars — but only between two runs of the same query. It pins them as **deterministic**, not as **correct**, and a consistently-wrong sidecar passes it every time. That is a general trap with determinism properties, and it's the reason this PR exists.

Also dropped the Debug-scraping out of `number_of`, which parsed a `Decimal` back out of `{:?}` output and had already grown a branch to cope with two different renderings — a test reaching through a formatting impl that is free to change. `Value`'s variants are public; it now matches them.

robcohen:
Both Copilot comments were right; fixed in 1379d0da4.

**The partition really was weaker than its docstring.** Comparing multisets of bare numbers cannot see a WHERE path that drops one row and duplicates another carrying the same number — and that is the common case here, not a remote one: every generated transaction posts `amt` to a random account and `-amt` to `Assets:Bank`, so equal numbers abound.

Verified by modeling exactly that shape — find two rows sharing a number, drop one, duplicate the other:

| row identity | result |
|---|---|
| `number` alone | **passes** — the union still reconstructs |
| `date\|account\|number` | fails |

Rows are now identified by all three, so a coincidence has to agree on all of them.

**And the boundary arithmetic was wrong.** `scale` is `0u32..3` — exclusive, so 0, 1 or 2 — which puts the ceiling at `10^2`, not `10^3`. Landing in `(0, 1]` needs `0 < n <= 100`: at most 100 of two million values, so about **0.04** expected hits across a run, not 0.3. The conclusion survives and gets sharper — not one run in twenty-five would have produced a value between 0 and 1, which is why swapping `number > 0` for `number > 1` changed nothing.

Issue #1967, opened 2026-08-05
Regenerate rustfava snapshots after the meta.hash format change (#1961)


robcohen:
Tracking issue for the downstream break landed knowingly with #1961.

## What changed

#1961 length-prefixes every field fed into `compute_directive_hash`, fixing three real collisions — the worst being that a transaction *with* a payee hashed identically to one *without*. The fix is correct and not up for debate here.

But `compute_directive_hash` is the FFI identity contract: it is surfaced as `meta.hash` across the WIT boundary. Length-prefixing changes **every digest**, and rustfava pins them in committed snapshots.

## Symptom

`rustfava tests against PR component` gains two failures on top of the pre-existing baseline:

- `tests/test_internal_api.py::test_get_ledger_data`
- `tests/test_json_api.py::test_api[account_report_off_by_one_journal-/off-by-one/api/account_report?interval=day&conversion=at_value&a=Assets]`

Attribution was isolated against two sibling PRs with no hash change:

| | failing rustfava tests |
|---|---|
| #1960 (merged) | `test_api_errors`, `test_serialise_entry_types` |
| #1962 (merged) | the same 2 |
| **#1961** | those 2 **+ the two above** |

And the assertion diff is the payload itself:

```
-  'hash': '3f43843fea1277977ca64674e0255cf3ac68b8e59…'
+  'hash': '89a5cc60095860b4164bcfca2be2edff4f4d27374…'
```

(`entry_hash` is scrubbed to `ENTRY_HASH` in those snapshots. `hash` — the field that moved — is not.)

## To do

- [ ] Regenerate the rustfava snapshots (`pytest --snapshot-update`) against a component built from `main` at or after #1961.
- [ ] Confirm the two tests above go green and that nothing else moved in the regenerated snapshots — a diff touching anything other than `hash` values would mean this change did more than intended.
- [ ] Decide whether `meta.hash` needs an explicit stability signal. Right now a consumer can pin it with no way to know the format changed. This is the same shape of problem as a parser change without a `CACHE_VERSION` bump: the fix is right, but the contract needs a version or a documented no-stability-guarantee. See the `CACHE_VERSION` precedent.

## Not in scope

The other two failures (`test_api_errors`, `test_serialise_entry_types`) predate all of this — red on merged #1960 and #1962. They mean the rustfava check has been red on every PR long enough that nobody reads it, which is its own problem and deserves its own issue.

Related: #1902 Phase 1.

robcohen:
[...]

PR #1984, opened 2026-08-07
fix(ffi): metadata is part of a directive's identity

### LINKED ISSUE #1968: Decide whether posting/directive metadata is part of meta.hash identity
Split out of #1961, where Copilot correctly pointed out that the posting fields fed into `compute_directive_hash` were incomplete.

#1961 fixed the part that is not a judgment call: `cost`, `price` and `flag` are unambiguously identity-bearing — two purchases of the same stock at different costs are different transactions, and they were hashing identically. All seven cases collided before that fix.

What #1961 deliberately did **not** decide is whether the following are part of a directive's identity:

- `Posting::meta` and `Transaction::meta`
- `Posting::comments` / `trailing_comments`, `Transaction::trailing_comments`

## Why this is a contract question, not a bug

It cuts both ways, and both directions have a real cost:

- **Exclude them** (today's behavior): two postings differing only by a metadata key are the same object to rustfava and the desktop app. If any consumer treats metadata as meaningful — and metadata is how users attach their own structured data — that is a genuine conflation.
- **Include them**: adding a comment to a transaction changes its identity. Anything keyed on `meta.hash` across an edit — a UI selection, a cached render, a link — breaks on a purely cosmetic change. That is arguably worse day to day.

There is a defensible middle (`meta` in, comments out) on the grounds that metadata is structured user data while comments are presentation. That is still a decision someone has to make deliberately.

## What decides it

Whoever owns the consumer contract should answer: **is `meta.hash` an identity for the directive as a semantic object, or for the directive as written?** Everything above follows from that.

Worth answering alongside the same question raised on #1967 — whether `meta.hash` needs an explicit stability signal at all. If the answer there is "consumers must not pin it," this becomes much lower stakes.

Note that `compute_directive_hash` now matches `CostNumber` exhaustively, so adding a variant is a compile error rather than a silent omission. The same protection does not exist for adding a *field* to `Posting`, which is how `cost`/`price`/`flag` went missing in the first place.

robcohen:
Closes #1968 — with **option B**: metadata in, comments out.

## Why this way

`meta.hash` is not a label, it's a **write address**. rustfava resolves an entry
by it and then mutates that entry's line in the source file — `save_entry_slice`,
`delete_entry_slice`, `insert_metadata`.

Two of those carry a second factor (the sha256sum of the displayed source slice),
so a wrong resolution is caught. **`insert_metadata` carries nothing** — and it is
the endpoint whose entire job is *creating metadata-only differences*. The one
operation that could act on a metadata collision was the one operation unguarded
against it.

**Comments stay out.** They're presentation, they don't reliably survive a
parse/format round trip, and including them would mean fixing a typo in a comment
changes a directive's identity. `comments_are_deliberately_not_part_of_the_identity`
pins that, because a decision to *exclude* something is only recorded if a test
fails when someone quietly reverses it.

**This is closer to where fava started.** Before the rustledger hash was bolted on,
rustfava called `beancount.core.compare.hash_entry(entry)` — default
`exclude_meta=False`, i.e. metadata *including filename and lineno*. Beancount's
own docstring: *"when you're using the hashes to uniquely identify transactions,
you want to include the filenames and line numbers."*

## What this is not

**Uniqueness.** Two transactions identical in every hashed field still collide, and
duplicates like that are ordinary in real ledgers. Only hashing source position
would make it unique, and that costs stability under every edit above the entry.
The consumer-side half is rustfava#273, which stops `_get_entry` silently picking
the first match.

## Two details that aren't incidental

- **Keys are sorted.** `Metadata` is an `FxHashMap`; hashing it in iteration order
  would make the digest depend on insertion history — nondeterminism in the one
  field every consumer treats as stable. Sorting also matches
  `meta_entries_from_core` in the component converter, which sorts for the same
  reason.
- **Values are variant-tagged, not rendered.** `MetaValue`'s `Display` is lossy:
  `Number(42)` and `Int(42)` both print `42`. #1766 went to deliberate trouble to
  keep those distinct; hashing the rendering would throw it away one layer down.

## Separately — four fields reached no digest at all

Not the #1968 judgment call. Found by auditing every directive struct's fields
against the arms of `compute_directive_hash`; same class as the `cost`/`price`/
`flag` omission #1961 fixed.

| field | what collided |
|---|---|
| `Open::booking` | FIFO vs LIFO — and the method decides how every reduction against the account books |
| `Balance::tolerance` | `100 USD` and `100 USD ~ 0.05` are different assertions; one can pass where the other fails |
| `Document::tags` / `links` | a Document carries both exactly as a Transaction does |
| `Custom::values` | the values *are* the directive — every `custom "budget"` on a date hashed the same |

## Verification

Six sabotages, six failures, restored green: drop transaction metadata, flatten the
`MetaValue` variant tags, remove the key sort, un-hash `booking`, un-hash
`Custom::values`, and start hashing comments.

The map-order test **self-checks**: whether two maps iterate differently is a
property of the chosen keys and `FxHashMap`'s bucket layout, not something the test
controls — so it asserts the orders differ *before* asserting the hashes match.
Without that, a hasher change could leave it passing with the sort deleted.

`cargo test --workspace` — 132 suites, 0 failures. Clippy clean on 1.97.0.

## ⚠️ Ordering

No `CACHE_VERSION` bump (the digest is derived at convert time from cached
directives, which are unchanged) and no WIT bump (the type is untouched) — which is
exactly the thing worth flagging. **This is a breaking behavioral change to a wire
field that neither `cargo-semver-checks` nor the WIT gate can see.** rustfava's
snapshots are the only thing that catches it, which is #1967.

### DIFF
--- a/crates/rustledger-ffi-wasi/src/hash.rs
+++ b/crates/rustledger-ffi-wasi/src/hash.rs
@@ -200,7 +200,10 @@ fn incomplete_amount(hasher: &mut Sha256, amount: Option<&rustledger_core::Incom
 /// exhaustive match means adding a variant is a compile error here - which
 /// forces the identity question to be answered rather than silently defaulted.
 ///
-/// Posting `meta` and comments are deliberately NOT hashed; see #1968.
+/// Posting `meta` IS hashed and posting comments are not — the #1968 split.
+/// Metadata is structured user data with a write API premised on it being
+/// meaningful; comments are presentation, and do not reliably survive a
+/// parse/format round trip. See [`metadata`].
 fn posting_identity(hasher: &mut Sha256, p: &rustledger_core::directive::Posting) {
     use rustledger_core::CostNumber;
 
--- a/crates/rustledger-ffi-wasi/src/hash.rs
+++ b/crates/rustledger-ffi-wasi/src/hash.rs
@@ -85,6 +85,93 @@ fn decimal(hasher: &mut Sha256, d: rustledger_core::Decimal) {
     field(hasher, d.to_string().as_bytes());
 }
 
+/// Feed one metadata VALUE in, variant-tagged.
+///
+/// The tag is not decoration. `MetaValue`'s `Display` is lossy across variants -
+/// `Number(42)` and `Int(42)` both render `42`, and `Account("Assets:A")`
+/// renders the same as a `String` would if the quoting were dropped - so
+/// hashing the rendering would collide `key: 42` with `key: 42.0` where the
+/// parser went to deliberate trouble (#1766) to keep them distinct.
+///
+/// Written out variant by variant rather than through `Display` for the same
+/// reason `CostNumber` is below: an exhaustive match means adding a variant is
+/// a compile error here, which forces the identity question to be answered
+/// rather than silently defaulted. `MetaValue::Int` was appended once already.
+fn meta_value(hasher: &mut Sha256, v: &rustledger_core::MetaValue) {
+    use rustledger_core::MetaValue as V;
+
+    match v {
+        V::String(s) => {
+            hasher.update([1u8]);
+            field(hasher, s.as_bytes());
+        }
+        V::Account(a) => {
+            hasher.update([2u8]);
+            field(hasher, a.as_bytes());
+        }
+        V::Currency(c) => {
+            hasher.update([3u8]);
+            field(hasher, c.as_bytes());
+        }
+        V::Tag(t) => {
+            hasher.update([4u8]);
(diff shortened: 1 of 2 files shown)


robcohen:
Fixed in 7f64aaa3d. `posting_identity`'s doc comment still carried the pre-#1968 line saying posting metadata was deliberately excluded — left behind by the change that reversed it. A comment contradicting the code it sits on is worse than none, since it's what a reader trusts before reading the body. It now states the split (metadata in, comments out) and points at `metadata`.

Issue #1985, opened 2026-08-07
[Bug]: BQL BALANCES shows unmerged lots and a dangling negative for AVERAGE-booked accounts


robcohen:
Found by extending the `query_report_realization_parity_test` guard past the two booking shapes it covered (#1902 Phase 3).

## Repro

```beancount
2024-01-01 open Assets:Broker "AVERAGE"
2024-01-01 open Assets:Cash
2024-01-01 open Income:PnL

2024-01-05 * "buy lot1"
  Assets:Broker   10 AAPL {150.00 USD}
  Assets:Cash  -1500.00 USD

2024-02-05 * "buy lot2"
  Assets:Broker   10 AAPL {200.00 USD}
  Assets:Cash  -2000.00 USD

2024-06-10 * "sell 5"
  Assets:Broker   -5 AAPL {} @ 180.00 USD
  Assets:Cash    900.00 USD
  Income:PnL
```

```
$ rledger report x.beancount balances --no-pager
Assets:Broker
               15 AAPL { 175.00 USD}          <-- correct

$ rledger query x.beancount BALANCES
Assets:Broker  10 AAPL { 200.00 USD}   10 AAPL { 150.00 USD}   -5 AAPL { 175.00 USD}
```

Three rows, including a **negative holding**, for an account that holds 15 units. `SELECT SUM(number)` agrees with the report (15), so the units are right — it is the lot-keyed presentation that is wrong.

## Why

The parity guard's own docstring already names the mechanism, and it turns out to state an invariant that does not hold:

> The query executor re-aggregates already-booked postings with naive `Inventory::add` (lot-key netting), while reports realize through the booking engine's lot-matching `apply`. They agree today ONLY because both consume post-booking directives — the loader's book phase has already matched every reduction against its lot, so naive re-aggregation lands on the same keys.

For AVERAGE that last step fails. Booking **merges** lots, so the reduction is booked at the merged average cost (175.00) while the two augmentations keep their original costs (150.00 and 200.00). Naive re-aggregation by lot key therefore sees three distinct keys where the booking engine sees one, and the reduction has no key to net against — so it survives as a negative row.

This is the same shape as #1726 (separate augmentation/reduction rows). It survived because the guard only covered FIFO-via-empty-`{}` and an explicit-cost reduction; AVERAGE is the one method where the booked cost of a reduction is not the cost of any augmentation.

## Scope

Checked the whole matrix on the same fixture:

| method | query vs report |
|---|---|
[...]

PR #1986, opened 2026-08-07
test: extend the green/red and realization parity guards

robcohen:
#1902 Phase 3, the last open item. Both guards existed; both had a hole big
enough to hide a live bug, and one of them was hiding one.

## Green/red: the fuzzer cannot reach the thing it is named after

`fuzz_green_eq_red` mutates raw **bytes**. Measured before writing anything: of
20,000 random byte inputs, **zero** produced even one directive — let alone the
well-formed simple transaction the green converter is the only consumer of. It's
a fine guard on lexing and error recovery, and it is structurally incapable of
exercising the green path. That path's correctness rested entirely on ~25
hand-picked corpus entries, i.e. on what someone thought to write down.

A structured generator inverts that — **487 of 500** generated ledgers parse and
**166** are error-free transactions with postings. It also *crosses* features
(costs × prices × flags × tags) where the corpus has one entry per feature.
#1713 was a product bug.

**What that buys, measured rather than argued.** Deleting green's bail on an
arithmetic price amount:

| | |
|---|---|
| 317 lib tests | pass |
| `ast`, `fuzz_regressions`, `structured_directives` | pass |
| **the new property** | **fails** |

**The generator was itself fixed by sabotage.** Two candidate divergences —
removing the price bail, and latching the *last* amount rather than the first
(the #1713 shape one level down) — were caught by *nothing at all*, because
nothing generated an arithmetic price amount. Added those; the first is now
caught. The second still isn't: it needs two `AMOUNT` children in one
`PRICE_ANNOTATION`, which appears syntactically unreachable — a defensive
branch, not a coverage gap, and not claimed as covered.

## Realization: five of seven booking methods were unpinned, and one diverges

The guard pinned FIFO-via-empty-`{}` and an explicit-cost reduction. Running the
same fixture through the whole matrix found **AVERAGE diverging today — #1985**.

| method | query vs report |
|---|---|
| FIFO, LIFO, HIFO, NONE | agree |
| STRICT | both error identically |
| **AVERAGE** | **diverge** |

The file's own docstring names the mechanism and states an invariant that turns
out not to hold: *"they agree ONLY because both consume post-booking directives
… so naive re-aggregation lands on the same keys."* AVERAGE is the one method
where a reduction is booked at a cost belonging to **no** augmentation (the
merged average), so it has no key to net against and survives as a negative row.
`report balances` gives the correct `15 AAPL {175.00}`; BQL BALANCES gives
`10 {200}  10 {150}  -5 {175}`.

The old `broker_holding` helper couldn't express any of this — it asserts
exactly *one* AAPL line, right for a netted holding and wrong for every method
that legitimately leaves two lots standing. Comparison is now on the **multiset**
of `(units, cost)`, dropping the lot date the report shows and BQL doesn't, for
the same reason the original helper dropped it.

**AVERAGE is pinned as a characterization test, not `#[ignore]`d.** It asserts
the wrong current behaviour on purpose, so fixing #1985 *fails* it and forces
AVERAGE into the agreeing set. An ignored test would let the fix land with
nothing noticing, which is how a known divergence becomes a permanent one.

That test also turns out to be the comparator's self-check: sabotaging
`aapl_lots` to return a constant makes every surface compare equal, so the
agreeing-methods test passes vacuously — and only the AVERAGE assertion catches
it.

## Sabotage-verified

| sabotage | result |
|---|---|
| green maps `@@` to per-unit | property fails (corpus guard does **not**) |
| green drops the last tag | property fails |
| green stops bailing on an arithmetic price | property fails; whole existing suite passes |
| report reads a different method's ledger | matrix fails |
| `aapl_lots` returns a constant | AVERAGE characterization fails |

`cargo test --workspace` — 133 suites, 0 failures. Clippy and fmt clean.

### DIFF
--- a/crates/rustledger-parser/tests/green_red_properties.rs
+++ b/crates/rustledger-parser/tests/green_red_properties.rs
@@ -237,12 +237,19 @@ fn ledger_source() -> impl Strategy<Value = String> {
     .prop_map(|parts| parts.concat())
 }
 
-/// Every observable on `ParseResult`, rendered for comparison.
+/// The `ParseResult` fields `fuzz_green_eq_red` compares — deliberately that
+/// list and not "everything".
 ///
-/// The same list the fuzz target asserts, kept together so a new field added
-/// to `ParseResult` is one edit rather than a silent coverage hole. Compared
-/// via `Debug` for the same reason the fuzz target does: it avoids requiring
-/// `PartialEq` on every field type.
+/// `syntax_root` is excluded, which Copilot was right to want said out loud:
+/// both paths build the SAME lossless CST and only differ in how they walk it,
+/// so comparing the tree would pass by construction while saying nothing about
+/// the conversion under test. Mirroring the fuzz target's list also keeps the
+/// two guards' notions of "the parity contract" from drifting apart, which is
+/// the failure mode this whole file exists to catch.
+///
+/// Kept in one place so a new field on `ParseResult` is one edit rather than a
+/// silent coverage hole. Compared via `Debug` for the same reason the fuzz
+/// target does: it avoids requiring `PartialEq` on every field type.
 fn observables(r: &rustledger_parser::ParseResult) -> Vec<(&'static str, String)> {
     vec![
         ("directives", format!("{:?}", r.directives)),
--- a/crates/rustledger-parser/tests/green_red_properties.rs
+++ b/crates/rustledger-parser/tests/green_red_properties.rs
@@ -1,0 +1,360 @@
+//! Green/red parser parity over GENERATED beancount source (#1902 Phase 3).
+//!
+//! The green transaction converter returns `Some` only when it exactly
+//! replicates the red path and bails to red otherwise, so any divergence is a
+//! green bug. Two things already pin that: `fuzz_green_eq_red`, and the
+//! `parse_green_eq_red_corpus` list inside `cst/green.rs`.
+//!
+//! Neither reaches the green path with any regularity, which is the gap this
+//! closes.
+//!
+//! `fuzz_green_eq_red` mutates raw BYTES. Measured before writing this: of
+//! 20 000 random byte inputs, **zero** produced even one directive — let alone
+//! the well-formed simple transaction the green converter is the only consumer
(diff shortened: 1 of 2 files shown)


robcohen:
All three Copilot comments fixed in 524b17815 — and the second one was hiding a bug.

**1. `unwrap_or_default()` on the lot cost was exactly the vacuity trap.** A change to how costs render would have handed *both* surfaces an empty string, and they would have compared equal — a guard whose entire job is comparing costs, passing precisely when it had stopped seeing them. It now panics with the offending line unless the cost parses as a number. A lot with no `{...}` at all stays a legitimate shape (NONE booking, and the unbooked reduction below) and is recorded as such rather than conflated with a cost that failed to parse.

Sabotage-verified: making the cost unreadable fails both the matrix and the AVERAGE pin.

**2. The STRICT test said "must FAIL on both surfaces" and only checked that both PRINTED the message.** Following the suggestion to assert exit status found that they do not fail alike at all:

| surface | exit | behaviour |
|---|---|---|
| `query BALANCES` | **0** | prints `-5 AAPL` for an account holding 15 units |
| `report balances` | **101** | **panics** in `apply()` |

Filed as **#1987**. The panic is `apply()`'s own assertion about unbooked reductions doing its job — the ambiguity is emitted as a *warning*, the pipeline continues, and realization is then handed a reduction that booking declined to book. Reached from a normal command on ordinary input, that is a crash rather than a guard. So the two surfaces diverge in the worst way available: one crashes, the other silently answers wrongly, and neither signals failure through its exit code.

Pinned as a characterization test naming the issue, like AVERAGE — a fix changes those exit codes and trips it, forcing it to be rewritten as the agreement assertion it was always meant to be. Sabotage-verified by pretending the fix landed.

**3. `observables` overclaimed.** It said "every observable on `ParseResult`" while mirroring the fuzz target's list, which omits `syntax_root`. It now says so and says why: both paths build the *same* lossless CST and differ only in how they walk it, so comparing the tree would pass by construction while saying nothing about the conversion under test.

Worth noting that this is the second time in this PR that a Copilot comment about a *test's* precision turned into a production bug — the same shape as the AVERAGE finding. `cargo test --workspace`: 133 suites, 0 failures.

Issue #1897, opened 2026-07-30
perf(booking): apply()'s unconditional snapshot costs 2.4x heap churn and 6% CPU


robcohen:
## Summary

The nightly profile regressed on 2026-07-30 (`54e47170`): **+6.08% instructions** and **heap total 86.8 MB → 210.5 MB (+142%)** on the 10k-transaction workload. Heap *peak* is byte-identical (27,452,151), so this is allocator churn rather than memory pressure.

All of it is the snapshot that `BookingEngine::apply` takes to make itself all-or-nothing.

## Data

From `origin/profiling`'s `history.jsonl`:

| date | commit | instructions | heap total | heap blocks |
|------|--------|-------------:|-----------:|------------:|
| 2026-07-29 | `b6ea6097` | 808,832,902 | 86,820,807 | 504,243 |
| 2026-07-30 | `54e47170` | 858,033,545 | 210,549,215 | 569,106 |

Four sites account for **100.0%** of the +123,728,408 byte delta, obtained by diffing `latest/dhat-heap.json` at `68f819d3` vs `ae732158` (normalize frames first — the raw strings carry build-specific addresses, which will make every site look new):

| delta bytes | delta blocks | site |
|------------:|-------------:|------|
| +112,940,512 | +18,287 | `UniqueArcUninit<imbl_sized_chunks::Chunk<Position, 64>>` via `Arc::make_mut` |
| +5,960,000 | +10,000 | `HashMap<Account, Option<Inventory>>` — the snapshot map, one per transaction |
| +2,706,476 | +18,287 | `HashMap<Currency, Decimal>` — `Inventory`'s cached totals, cloned |
| +2,121,292 | +18,287 | `HashMap<Currency, usize>` — ditto |

18,287 clones / 10,000 transactions ≈ 1.83, i.e. one per account a transaction touches.

Allocation stack for the dominant site:

```
Inventory::add                     (core/src/inventory/mod.rs:637)
  imbl GenericVector …
    SharedPointer::make_mut
      Arc<Chunk<Position, 64>>::make_mut     <-- copy-on-write
BookingEngine::try_apply_posting   (booking/src/book.rs:660)
BookingEngine::apply               (booking/src/book.rs:734)
run_booking::{closure#2}           (loader/src/process.rs:828)
```

## Cause

[...]

PR #1988, opened 2026-08-07
fix(booking): report a failed reduction instead of panicking in debug and lying in release

### LINKED ISSUE #1987: [Bug]: `report balances` panics on an ambiguous STRICT lot match; `query BALANCES` prints a wrong row
Found while addressing a Copilot comment on #1986 — it pointed out that the STRICT parity test asserted only that both surfaces *print* the ambiguity message, never that they failed. Checking the exit codes turned up a panic.

## Repro

```beancount
2024-01-01 open Assets:Broker "STRICT"
2024-01-01 open Assets:Cash
2024-01-01 open Income:PnL

2024-01-05 * "buy lot1"
  Assets:Broker   10 AAPL {150.00 USD}
  Assets:Cash  -1500.00 USD

2024-02-05 * "buy lot2"
  Assets:Broker   10 AAPL {200.00 USD}
  Assets:Cash  -2000.00 USD

2024-06-10 * "sell 5"
  Assets:Broker   -5 AAPL {} @ 180.00 USD
  Assets:Cash    900.00 USD
  Income:PnL
```

An empty `{}` reduction under STRICT is ambiguous — two lots match — which is correct and expected. What happens next is not.

```
$ rledger report x.beancount balances --no-pager ; echo "exit=$?"
warning: BOOK: Ambiguous lot match for AAPL: 2 lots match in Assets:Broker (2024-06-10, "sell 5")

thread 'main' panicked at crates/rustledger-booking/src/book.rs:813:13:
apply() reduction failed — the transaction must be booked before apply() (postings
filled, costs resolved); applying an unbooked reduction silently over-sells
inventory. A caller that cannot guarantee booked input must not derive a figure
from this realization (...)
exit=101
```

```
$ rledger query x.beancount BALANCES ; echo "exit=$?"
BOOK: Ambiguous lot match for AAPL: 2 lots match in Assets:Broker (...)

Assets:Broker  10 AAPL { 200.00 USD}   10 AAPL { 150.00 USD}   -5 AAPL
exit=0
```

## Two problems

**1. A user ledger panics the CLI.** `report balances` exits 101 with a Rust panic message and a backtrace hint. The assertion at `crates/rustledger-booking/src/book.rs:813` is doing its job — it exists precisely to stop an unbooked reduction silently over-selling inventory — but it is being reached from a normal command on ordinary input, which makes it a crash rather than a guard. CLAUDE.md: *"Use `Result<T, E>` for fallible operations, not panics."*

The root cause is upstream: the ambiguous lot match is emitted as a **warning** and the pipeline continues, so `apply()` is handed a reduction the booking phase declined to book. Either that diagnostic should be an error that stops realization, or the report path should decline to realize accounts whose booking failed.

**2. `query BALANCES` exits 0 and prints a wrong row.** `-5 AAPL` with no cost, alongside both original lots — a negative holding for an account that holds 15 units. Same shape as #1985 but reached by a different route: there the reduction was booked at a cost matching no augmentation, here it was never booked at all.

So the two surfaces diverge in the worst possible way: one crashes, the other silently answers wrongly. Neither reports failure through its exit code.

## Suggested direction

Whatever the fix, the two surfaces should agree on the *outcome*, which is the realization-family concern in the duplication registry. My inclination is that an ambiguous STRICT reduction should be a hard error at load time — it is not a situation any downstream figure can be derived from, which is exactly what the panic message says.

#1986 pins the current behaviour as a characterization test naming this issue, so a fix trips it.

robcohen:
Closes #1987.

## The bug was worse in release than the issue said

`apply()` ended its posting loop with `debug_assert!(reduced.is_ok())` followed by
`let _ = reduced`, commented *"release builds keep the historical
ignore-and-continue behavior"*. So the two build profiles disagreed about a
user's ledger:

| profile | on an ambiguous STRICT reduction |
|---|---|
| debug | `report balances` **panicked**, exit 101 |
| release | the reduction was **dropped** — 20 AAPL reported for an account holding 15 |

The release half is the worse one, and it's the one real users got. An
over-stated holding isn't a crash you can see; it's a number you act on. CI never
saw it because CI runs debug, where the same input panics instead.

Reduction failures are now **returned**. The precondition is unchanged and still
documented — `apply` wants booked input — but violating it is reported rather
than asserted in one profile and ignored in the other. Every existing caller
already handles the `Err`: `book()` records it and marks the transaction failed,
the wasm entry checks `is_ok()`.

## Atomicity

Returning mid-loop would leave earlier postings applied — exactly #1976's shape.
The snapshot gate widens from `overflow_is_possible` to `rollback_needed` =
overflow **or** the transaction contains a reduction.

Kept as two predicates so #1897's reasoning stays legible: snapshotting *every*
transaction cost 2.4× heap churn and 6% CPU, and reductions are a minority of
transactions, so this widens the gate without reopening that regression.
`has_reduction` calls the canonical `Inventory::is_booking_reduction` rather than
re-deriving "is this a reduction" from the sign.

Deliberately *not* relying on the coincidence that `overflow_is_possible` returns
true for an unfilled posting — a filled-but-unbookable transaction (ambiguous
STRICT with explicit amounts) would otherwise mutate half its accounts and then
return `Err`.

## The CLI half

The root cause is upstream of `apply`: `run_booking` partitions failures out, and
`finalize` re-merges them so the user still sees their own input — documented as
*"for output"*, with nothing stopping the same directives flowing into
computation. `account_balances` even documents the precondition the pipeline then
fails to establish.

`bail_on_booking_errors` now gates `query` and `report`, mirroring
`bail_on_parse_errors` one phase later. That also fixes the **query** half, which
never touched `apply`: BALANCES re-aggregates by lot key and printed a dangling
`-5 AAPL`, exiting 0.

```
before:  report → panic, exit 101      query → wrong row, exit 0
after:   both → "1 transaction(s) could not be booked; refusing to derive
                 figures … Run `rledger check x.beancount`", exit 1
```

`rledger check` still reports it, and a clean ledger is untouched (verified).

## Two tests changed because the behavior did

**`strict_ambiguity_is_reported_by_both_surfaces`** was the characterization test
added with the guard in #1986. It tripped — which is what it was for — and is now
the agreement assertion it was always meant to be.

**`ambiguous_strict_reduction_yields_no_rows`** asserted capgains *succeeded* with
a header-only CSV. Not fabricating a disposal was right; exiting 0 was not — a
header-only capital-gains CSV is indistinguishable from "you had no disposals this
year". It now asserts the refusal **and** that no row is fabricated.

## A third finding

The parity guard's NONE fixture was comparing two **wrong** answers. `-5 AAPL {}`
cannot book under NONE (cost tracking off → "2 unknowns"), so both surfaces were
realizing an unbooked transaction and agreeing. Making failures fatal is what
exposed it. NONE now reduces by price, the only shape that means anything without
lots.

## Verification

| sabotage | result |
|---|---|
| restore ignore-and-continue | 2 tests fail |
| return `Err` without rolling back | atomicity test fails |

The new tests pass identically under `--release`, which is the point — the old
`debug_assert` is exactly why debug and release disagreed.

`cargo test --workspace` — 134 suites, 0 failures. Clippy and fmt clean.

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -683,6 +683,49 @@ impl BookingEngine {
     /// what any one account can consume, which is the safe direction, and it
     /// avoids grouping postings by account and currency — an allocation, which
     /// is the thing being removed.
+    /// Whether `apply` must snapshot before mutating.
+    ///
+    /// Two reasons a posting can fail partway through, and both now abort the
+    /// transaction, so both need the earlier postings rolled back:
+    ///
+    /// * an `add` that overflows (#1863) — [`Self::overflow_is_possible`];
+    /// * a `reduce` with no matching lot, which since #1987 is an error rather
+    ///   than something to ignore.
+    ///
+    /// The reduction half is checked separately because `overflow_is_possible`
+    /// is about magnitudes and answers `true` for an unfilled posting only by
+    /// coincidence. Relying on that coincidence would leave a filled-but-
+    /// unbookable transaction (an ambiguous STRICT match with explicit
+    /// amounts) mutating half its accounts and then returning `Err`.
+    ///
+    /// Kept as two predicates rather than folded into one so the #1897
+    /// perf reasoning stays legible: snapshotting EVERY transaction cost 2.4x
+    /// heap churn and 6% CPU, which is why neither half is unconditional.
+    /// Reductions are a minority of transactions in a real ledger, so this
+    /// widens the gate without reopening that regression.
+    fn rollback_needed(&self, txn: &Transaction) -> bool {
+        self.overflow_is_possible(txn) || self.has_reduction(txn)
+    }
+
+    /// Whether any posting reduces an existing cost-bearing lot.
+    ///
+    /// Uses the canonical [`Inventory::is_booking_reduction`] rather than
+    /// re-deriving "is this a reduction" from the sign, which is the drift the
+    /// duplication review kept finding.
+    fn has_reduction(&self, txn: &Transaction) -> bool {
+        txn.postings.iter().any(|posting| {
+            let Some(units) = posting.amount() else {
+                // Unfilled: booking did not complete, so this transaction is
+                // exactly the unbookable kind that needs the rollback.
+                return true;
+            };
+            let method = self.method_for(&posting.account);
+            self.inventories
+                .get(&posting.account)
+                .is_some_and(|inv| inv.is_booking_reduction(units, posting.cost.as_ref(), method))
+        })
+    }
+
     fn overflow_is_possible(&self, txn: &Transaction) -> bool {
         let mut sum_abs = rustledger_core::Decimal::ZERO;
         for posting in &txn.postings {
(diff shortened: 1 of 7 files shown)


robcohen:
Both Copilot comments fixed in 3cf8d1964 — and the first was a real hole in the fix.

**`--no-errors` bypassed the guard entirely.** It sat *inside* the `!args.no_errors` block, so:

```
$ rledger query --no-errors x.beancount BALANCES
Assets:Broker  10 AAPL { 200.00 USD}   10 AAPL { 150.00 USD}   -5 AAPL
exit=0
```

The dangling `-5 AAPL` row with exit 0 — the exact bug this PR exists to prevent, reachable through a flag documented as suppressing the *printing* of diagnostics. Verbosity is not permission to compute figures from a ledger that did not book. Hoisted out, with a regression test.

Worth recording how my first attempt to reproduce it **"passed"**: I ran `query FILE QUERY --no-errors`, where the flag lands after the positionals and is swallowed into the BQL text, so `no_errors` stayed false and the guard fired. I nearly wrote the comment off on that basis. Instrumenting the flag rather than trusting the output is what settled it, and the test now says the argument order matters.

**The fixture inconsistency was real too** — both buys paid `-1500.00` while the second lot is 10 @ 200.00. `apply` doesn't check balance today, so nothing failed, which is exactly why it was worth fixing before something starts depending on it. The cash leg is now derived from the lot.

Sabotage-verified: nesting the guard back inside `--no-errors` fails the new test. `cargo test --workspace` — 134 suites, 0 failures.

PR #1989, opened 2026-08-08
test(compat): compare WHY each tool rejects a file, not just whether

robcohen:
#1902 Phase 3's last item — closes the issue.

## The gap

The oracle's error axis compared `bc_err == rl_err` — **two booleans**. So two
tools rejecting a file for entirely unrelated reasons counted as agreement:
beancount refusing over a bad lot match while rledger refuses over an unopened
account was indistinguishable from the two agreeing.

A common family vocabulary makes the incompatible error names comparable —
`parse`, `account`, `balance_assertion`, `pad`, `transaction_balance`, `booking`,
`currency`, `option`, `document`.

## The table is the risk, so every row is measured

`--self-test` writes a minimal fixture per family and checks that each side still
lands where the table says. A mapping table is a copy of someone else's
vocabulary, and a stale copy turns real divergence into "agree" — the drift the
duplication review keeps finding, one layer out.

Two measurements shaped the design:

- **`ValidationError` is a catch-all.** One beancount class covers an unopened
  account (E1001), a closed one (E1003), a transaction that doesn't balance
  (E3001) and a disallowed currency (E5002) — four rledger families. So a kind
  maps to a **set**, and agreement is a **non-empty intersection**. Demanding
  equality would have reported every `ValidationError` file as divergent.
- **`BalanceError` exists twice** — `core.interpolate` and `ops.balance`, meaning
  different things. Kinds are now qualified by module so the two can't be
  conflated.

## Found: the oracle couldn't see rledger's booking errors at all

The code extractor matched `[A-Z]\d{4}`, and the booking phase emits
`error[BOOK]` — a **word**. Every booking failure came back with an *empty* code
set.

Invisible while the axis compared booleans. Adding the kind comparison surfaced
it immediately, as 8 corpus files reporting "both reject, kinds undecidable" with
**zero unmappable kinds** — a set that was empty, not unrecognized. It also
degraded the older rledger-only diagnostics, which printed `(no code)` for
exactly those files.

## The first table produced three false positives

`e09`, `e14`, `e17` came back divergent. Reading them is what corrected the
table: both tools report the same interpolation failure — *"Too many missing
numbers for currency group 'EUR'"* against *"multiple postings missing amounts …
(2 unknowns)"* — beancount from `booking_full`, rledger from its booking pass. So
`InterpolationError` maps to **both** families. That case is now a measured
fixture, so the calibration can't silently regress.

Worth saying plainly: those three were the table's error, not the tools'. An
oracle that reports a divergence it caused is worse than one that reports
nothing, and the only defence is calibrating against reality rather than
asserting from a doc.

## Blind spots are reported, not counted as agreement

An unmappable kind is `undecidable`, and the unmapped counts are printed. An axis
that treats what it cannot classify as agreement is the vacuous-guard shape —
reporting zeros forever while its vocabulary drifts underneath it.

## Verification

| sabotage | result |
|---|---|
| stale `ValidationError` mapping | self-test fails ("kind table stale") |
| comparator always returns agree | self-test fails |
| blind spots treated as agreement | self-test fails |
| `BOOK` mismapped to `document` | corpus reports 8 findings |

Corpus: **8 files both reject and agree on the reason, 0 undecidable, 0
divergences.** CI already hard-fails on `--self-test` (it's the one step that
isn't `continue-on-error`), so the new axis is gated there.

### DIFF
--- a/scripts/compat-values.py
+++ b/scripts/compat-values.py
@@ -44,7 +44,7 @@ rledger raised E3001 on a file beancount accepts.
 """
 from __future__ import annotations
 
-import argparse, csv, io, json, re, subprocess, sys
+import argparse, csv, io, json, re, subprocess, sys, tempfile
 from collections import defaultdict
 from decimal import Decimal, InvalidOperation
 from pathlib import Path
@@ -111,6 +111,16 @@ NON_FATAL_ERRORS = {"ValidationError", "BalanceError", "DocumentError", "PadErro
 KNOWN_ERROR_DIVERGENCES: dict[tuple[str, str], str] = {}
 
 
+# Deliberate deviations in WHICH KIND of error the two tools raise, for files
+# both already reject. Keyed by (basename, beancount families, rledger
+# families) so a pin covers one file and one direction of disagreement — the
+# same surgical-pin rule the registry above follows, for the same reason.
+#
+# EMPTY ON PURPOSE at introduction. Entries get added only with a written reason
+# after a corpus run shows them.
+KNOWN_KIND_DIVERGENCES: dict[tuple[str, tuple[str, ...], tuple[str, ...]], str] = {}
+
+
 def beancount_errors(path: Path):
     """(errored, kinds) per Python beancount, or (None, set()) if it crashed.
 
@@ -128,7 +138,12 @@ def beancount_errors(path: Path):
         _, errors, _ = loader.load_file(str(path))
     except Exception:
         return None, set()
-    return bool(errors), {type(e).__name__ for e in errors}
+    # QUALIFIED by module, because the bare class name is ambiguous:
+    # `BalanceError` exists in BOTH `beancount.core.interpolate` and
+    # `beancount.ops.balance`, and they mean different things. Comparing kinds
+    # (below) with the bare name would silently conflate a failed balance
+    # ASSERTION with a transaction that does not balance.
+    return bool(errors), {f"{type(e).__module__}.{type(e).__name__}" for e in errors}
 
 
 def rledger_errors(binary: str, path: Path):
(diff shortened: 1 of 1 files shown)


robcohen:
All three fixed in 82a782aad — and the second was a factual error of mine, not just a stale comment.

**The comment claimed something the table didn't do.** It said module qualification separated two different `BalanceError` classes, in `core.interpolate` and `ops.balance`, while the table mapped both to one family. Checking it: **`ops.balance` imports the class from `core.interpolate`** — they're the same object, and `__module__` always reads `core.interpolate`. A package scan lists it under both module names, and that's what I misread as two classes.

So `beancount.ops.balance.BalanceError` was a **dead key**. It looked thorough — that module really does raise the error — and no error can ever carry that name.

**Nothing would have caught it**, because a key that never matches simply never fires; it reads as coverage while contributing none. The self-test now rejects any key that isn't a class's own qualified name, which is a strictly stronger check than the family fixtures: those prove the keys that *do* fire are right, this proves there are no keys that *can't*. Verified by putting the dead key back and by inventing a class:

```
SELF-TEST FAILED: kind table key 'beancount.ops.balance.BalanceError' is a RE-EXPORT
  of 'beancount.core.interpolate.BalanceError'; errors carry the defining module,
  so this key can never match
SELF-TEST FAILED: kind table key 'beancount.ops.pad.MadeUpError': no such class
```

The qualification stays, for the reason that survives measurement rather than the one I'd written: beancount reuses short error names across modules (`CommodityError`, `ConfigError`, `BookingError` each appear more than once), so a bare name is ambiguous in general even where it happened not to be here.

**CodeQL, twice:** the module-level `tempfile` import is unused — `self_test` already imports it at function scope. Removed.

Corpus unchanged: 8 files both reject and agree on the reason, 0 undecidable, 0 divergences.

Worth noting this is the third time in this PR that the *table* was wrong rather than the tools — first the three interpolation false positives, now a key that could never match. That's the argument for measuring a mapping against live fixtures instead of asserting it from documentation.

PR #1990, opened 2026-08-08
fix(query): realize BALANCES through the booking engine, not a bare add

### LINKED ISSUE #1985: [Bug]: BQL BALANCES shows unmerged lots and a dangling negative for AVERAGE-booked accounts

robcohen:
Closes #1985.

## The bug

BQL accumulated per-account balances with `Inventory::add` — **unconditionally,
with no reduction branch at all**. Reports realize through `BookingEngine`, which
resolves the account's booking method and then *reduces against a lot* or *adds a
new one*.

The two agreed for FIFO/LIFO/HIFO **by coincidence**: booking has already
resolved a reduction's cost, so its lot key matches an existing lot and `add`
nets it. Under AVERAGE it doesn't — the reduction is booked at the *merged
average* cost, a key belonging to no augmentation — so it survived as a dangling
negative:

```
query BALANCES   10 AAPL {200.00}   10 AAPL {150.00}   -5 AAPL {175.00}
report balances  15 AAPL {175.00}
```

…for an account holding 15. `SELECT SUM(number)` agreed with the report all
along at 15, so the units were right and only the lot-keyed presentation was
wrong.

BQL now calls `BookingEngine::apply_posting`, newly public because the query
engine needed exactly this decision and was re-deriving a broken half of it. One
realization rather than two that agree by luck — the duplication registry's
realization family, and this was the drift.

## Why no differential test could have found this

**Python beancount does not implement AVERAGE.**
`beancount.parser.booking_method.booking_method_AVERAGE` raises
`AmbiguousMatchError("AVERAGE method is not supported")`, with the real
implementation left commented out (*"DISABLED - This is the code for AVERAGE,
which is currently disabled"*). Meanwhile `Booking.AVERAGE` stays in the enum and
is accepted in an `open` directive — so a ledger declaring it parses and then
books nothing. Run this PR's fixture through beancount and 20 units remain.

So the compat oracle **cannot referee this**. Correctness rests on the definition
— 10 @ 150 plus 10 @ 200 is 3500 over 20 units, so 175.00 per unit; less 5 leaves
15 at 175.00 — and on the two surfaces agreeing.

That's the general point rather than a footnote: **a divergence in a feature the
reference implementation lacks is invisible to differential testing by
construction.** The internal parity guard was the only thing that could have
caught it, which is a decent argument for #1902's whole approach. Documented at
`reduce_average`, where someone will next be tempted to "match beancount".

## Test churn

`average_booking_diverges_today` was the characterization test added with the
guard in #1986 — it asserted the wrong behavior on purpose so a fix would trip
it. It did. Replaced by `average_booking_nets_to_a_single_merged_lot`, and
AVERAGE joins the agreeing-methods loop.

## Verification

| sabotage | result |
|---|---|
| restore the unconditional `add` | 2 tests fail |
| build the engine without registering account methods (AVERAGE falls back to the default) | 2 tests fail |

`cargo test --workspace` — 134 suites, 0 failures. Clippy and fmt clean. The
beancount oracle reports 0 divergences on every axis, and its self-test passes.

### DIFF
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -619,16 +619,6 @@ impl BookingEngine {
         })
     }
 
-    /// Apply ONE posting to the running inventories, returning an error if it
-    /// reduces a lot that is not held.
-    ///
-    /// The per-posting core of [`Self::apply`], which ignores a reduction failure
-    /// per its booked-input contract; the fallible signature keeps the over-sell
-    /// detectable by the `debug_assert` there.
-    ///
-    /// # Errors
-    /// Returns the reduce error when the posting reduces a lot that is not held
-    /// (an over-sell / unbooked-input contract violation).
     /// Apply ONE booked posting to the running inventories.
     ///
     /// The canonical "what does this posting do to an inventory" decision:
@@ -651,14 +641,6 @@ impl BookingEngine {
     ///
     /// Same as [`Self::apply`]: the posting must already be booked.
     pub fn apply_posting(
-        &mut self,
-        posting: &Posting,
-        date: rustledger_core::NaiveDate,
-    ) -> Result<(), BookingError> {
-        self.try_apply_posting(posting, date)
-    }
-
-    fn try_apply_posting(
         &mut self,
         posting: &Posting,
         date: rustledger_core::NaiveDate,
@@ -872,7 +854,7 @@ impl BookingEngine {
             // other. Callers already handle this — `book()` records the error
             // and marks the transaction failed, the wasm entry point checks
             // `is_ok()`, and the CLI refuses to derive a figure at all.
-            if let Err(e) = self.try_apply_posting(posting, txn.date) {
+            if let Err(e) = self.apply_posting(posting, txn.date) {
                 // `snapshot` is `Some` whenever this arm is reachable:
                 // `rollback_needed` covers both failure modes. Restoring
                 // nothing would be silent corruption — precisely the bug being
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -629,6 +629,35 @@ impl BookingEngine {
     /// # Errors
     /// Returns the reduce error when the posting reduces a lot that is not held
     /// (an over-sell / unbooked-input contract violation).
+    /// Apply ONE booked posting to the running inventories.
+    ///
+    /// The canonical "what does this posting do to an inventory" decision:
+    /// resolve the account's booking method, then REDUCE against a lot or ADD a
+    /// new one. Public because the query engine needs exactly this and was
+    /// re-deriving a broken half of it — `Inventory::add` unconditionally, with
+    /// no reduction branch at all (#1985). That looked right for FIFO/LIFO,
+    /// where booking has already resolved the reduction's cost so its lot key
+    /// matches an existing lot and `add` nets it by coincidence, and produced a
+    /// dangling negative position under AVERAGE, where the reduction is booked
+    /// at the merged average cost — a key belonging to no augmentation.
+    ///
+    /// # Errors
+    ///
+    /// [`BookingError`] when the reduction finds no matching lot, or an `add`
+    /// overflows. Unlike [`Self::apply`] this is a SINGLE posting, so there is
+    /// nothing to roll back — the caller owns transaction atomicity.
+    ///
+    /// # Precondition
+    ///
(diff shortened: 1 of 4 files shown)


robcohen:
All four fixed in 29e3a27fa — and one of them led somewhere more interesting.

**1. Duplicated rustdoc.** My insertion landed between `try_apply_posting`'s existing doc and its signature, orphaning the old one onto the new. That old doc was *also* stale in its own right: it described the reduction failure as ignored and detectable via a `debug_assert`, which #1988 removed. Fixed by deleting the **wrapper** rather than the duplicate text — the public `apply_posting` was a one-line call through to a private function with no other caller worth the indirection, so it's now one function with one doc.

**2. Two passes over the directive stream.** Registration now reads the vec `directive_iter` already builds. `register_account_methods` only looks at `Open` directives, so order is irrelevant and it's the same registration one pass earlier.

**3. `resolved` computed unconditionally.** Before this PR the per-account accumulation used it, so unconditional was right; `apply_posting` resolves the position itself, leaving `needs_balance` as the only consumer. Now gated on that.

**4. "Four methods"** with AVERAGE added.

---

**The one that mattered:** chasing #4 turned up that the *file docstring* described an architecture this PR removes. It explained that the two surfaces agree

> ONLY because both consume post-booking directives … so naive re-aggregation lands on the same keys

— true for FIFO/LIFO/HIFO, false for AVERAGE, and the whole reason #1985 existed. A comment asserting an invariant nobody checks, with a live exception, is precisely the stale-invariant class #1902 names. It now says what's true and records why the old claim held for as long as it did.

The section header had the same problem in the present tense ("AVERAGE diverges today"), rewritten to note the matrix stays so the next method added is pinned from the start rather than five releases later.

134 suites, 0 failures. Clippy and fmt clean.

Issue #1964, opened 2026-08-05
[Bug]: FFI `get-account-type` and the LSP classify accounts with hardcoded roots; report commands use the configured names


robcohen:
Found by the canonical-function drift sweep (#1902 Phase 2, sweep 1).

CLAUDE.md lists `AccountTypes` as the canonical for *"config-aware account classification — never match root prefixes by string."* There are in fact **two** classifiers in core, and consumers split between them:

| classifier | config-aware? |
|---|---|
| `AccountTypes::kind(&self, account)` | **yes** — matches the configured `name_*` values |
| `account_type(account) -> &'static str` | **no** — hardcodes `Assets`/`Liabilities`/`Equity`/`Income`/`Expenses` |

## Who uses which

| surface | classifier |
|---|---|
| `report balsheet`, `income`, `networth`, `holdings`, `budget` | `AccountTypes::kind` ✓ |
| **FFI `get-account-type`** (WIT contract) | `account_type` ✗ |
| **LSP** hover / definition (`is_account_type`) | `ACCOUNT_TYPES.contains` ✗ |
| plugin `sell_gains` | raw `starts_with("Income:")` / `starts_with("Expenses:")` ✗ |

So for a ledger with `option "name_expenses" "Depenses"`, `report balsheet` classifies correctly while the FFI answers `"unknown"` and the LSP stops offering account affordances. **Same ledger, different answers depending on which surface you ask.**

That is the shape #1902 calls *"policy threaded to some consumers only"*, and the shape CLAUDE.md's canonical rule exists to prevent.

## One nuance worth recording

`split_expenses` was previously fixed *away* from `starts_with("Expenses:")` and onto `account_type` — and its comment says the new call is *"identical to the old `starts_with("Expenses:")`"*. That is exactly the problem: the raw string was replaced by a helper that hardcodes the same string. Centralized, but no more config-aware than before.

## Scope (revised 2026-08-06 — the original was too pessimistic)

I first wrote that the FFI case "means changing the WIT contract" and needed a deliberate decision. Re-checking against main, that framing was wrong and made this look more blocked than it is.

**`get-account-type` is in the stateless `util` interface**, alongside `types()` and `is-encrypted(path)`. It has no ledger handle *by design* — that interface is the ledger-free utility bucket. Making *it* config-aware is not the fix.

**The `session` resource already holds everything required.** The WIT `ledger-options` record already exposes the renames:

```wit
name-assets: string,
name-liabilities: string,
name-equity: string,
[...]

PR #1992, opened 2026-08-09
refactor(core): one is_default_account_root, and a guard on the list it reads

robcohen:
Follow-up from reviewing #1964. That issue is about classifiers that hardcode
account roots — and the predicate asking that very question was itself
duplicated, in the two crates that must agree about it:

```rust
// crates/rustledger-wasm/src/editor/helpers.rs
matches!(s, "Assets" | "Liabilities" | "Equity" | "Income" | "Expenses")

// crates/rustledger-lsp/src/handlers/utils.rs
rustledger_core::ACCOUNT_TYPES.contains(&s)
```

They agreed only because those five words happened to appear in both, and **each
crate carried its own five-case unit test** — so a change to either would have
left both suites green while the two editor surfaces disagreed about what an
account root is.

Now one function in `rustledger-core`, next to the list it reads. Both editor
crates already depend on core, so this needed no new dependency edge.

## The list that stays duplicated, and why that is now safe

`rustledger-completion` keeps its own copy: the crate has **zero dependencies**,
which is precisely what lets both the LSP and the WASM editor take it. Core's
constant already documented that copy; completion's now names core's back — the
half of CLAUDE.md's "document the divergence at BOTH sites" that was missing.

What was *also* missing is the part that makes a deliberate copy safe: a copy is
only allowed if something asserts the agreement, and nothing did.
`account_types_match_core` now does, from `rustledger-lsp` — a crate that can see
both, which neither owner can.

A second test asks the **predicate** about every root rather than comparing two
constants, because the equality alone can't see a future `is_account_type` that
hardcodes its own words again.

## Deliberately not config-aware

This answers the editor's *lexical* question — "does this bare word look like an
account root, so should I offer hover/definition affordances?" — often asked with
no ledger loaded at all.

It is **not** the classifier for an account you hold. A bare root isn't even a
valid account name (`open Assets` is a parse error, P0012), so every real account
reference contains a `:` and belongs to `AccountTypes`, which honors the `name_*`
renames. That's said in the doc, because the distinction is what #1964 is about
and it's what made item 3 of that issue look bigger than it is.

## Verification

| sabotage | result |
|---|---|
| add a sixth root to completion's list only | `account_types_match_core` fails |
| reimplement the predicate with its own hardcoded words | `is_account_type_answers_for_every_root` fails |

`cargo test --workspace` — 134 suites, 0 failures. Clippy and fmt clean.

Behaviour is unchanged: both lists hold the same five words today, so this is a
structural fix, not a functional one.

### DIFF
--- a/crates/rustledger-completion/src/lib.rs
+++ b/crates/rustledger-completion/src/lib.rs
@@ -17,6 +17,18 @@
 //!    neutral [`CompletionCandidate`] lists for each context.
 
 /// Standard Beancount account types.
+///
+/// A DELIBERATE copy of `rustledger_core::ACCOUNT_TYPES`, which names this one
+/// in return. This crate has no dependencies at all — that is what lets both
+/// the LSP and the WASM editor take it — so it cannot read core's.
+///
+/// Kept honest by `account_types_match_core` in `rustledger-lsp`, which is a
+/// crate that can see both. Without that guard the copy is the #1964 shape:
+/// two lists that agree until one of them does not, with each side's tests
+/// passing throughout.
+///
+/// Only the DEFAULT roots. Classification of a real account belongs to
+/// `rustledger_core::AccountTypes`, which honors the `name_*` renames.
 pub const ACCOUNT_TYPES: &[&str] = &["Assets", "Liabilities", "Equity", "Income", "Expenses"];
 
 /// Standard Beancount directives.
--- a/crates/rustledger-core/src/identifiers.rs
+++ b/crates/rustledger-core/src/identifiers.rs
@@ -335,6 +335,14 @@ pub const ACCOUNT_TYPES: [&str; 5] = ["Assets", "Liabilities", "Equity", "Income
 
 /// Whether a bare word is one of the DEFAULT account-type roots.
 ///
+/// Named for what it does, not for what it might be mistaken for. It was
+/// `is_account_type`, which promises CLASSIFICATION — and on a ledger with
+/// `option "name_expenses" "Depenses"` it answers `false` for `Depenses` and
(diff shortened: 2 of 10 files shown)


robcohen:
Added the rename, per the analysis on #1964.

`is_account_type` promises **classification**. It doesn't classify — on a ledger with `option "name_expenses" "Depenses"` it answers false for `Depenses` and true for `Expenses`, both wrong as classification.

It's inert today, which is exactly why the **name** is the thing to fix rather than the behaviour. Measured, hovering both roots as bare words:

| word | predicate | hover | references |
|---|---|---|---|
| bare `Assets` | true | none | 0 |
| bare `Depenses` | false | none | none |

Gate true and gate false show the user the same thing: nothing. Every downstream lookup matches `word == open.account` or "word is a *sub*account of open.account" — all five checked — and a bare root is neither, because a valid account always contains a colon (`open Assets` is a parse error).

So nothing can act on the wrong answer. What *could* is the next caller, who reads the name and uses it somewhere it isn't inert. A rename removes that risk; #1964 item 3's proposal (thread options through five call sites) would be a change no test could demonstrate, and would make WASM answer differently per file since `ParseResult.options` is empty for every included buffer.

**The two crate-local wrappers are deleted rather than renamed twice.** Once they delegate to core they're an indirection plus a second name for one thing, which is what this branch exists to remove. Call sites now say `rustledger_core::is_default_account_root(&word)` directly, and the two duplicated five-case tests go with them — the drift guard and the per-root coverage test replace both.

Re-verified after the rename: a predicate that hardcodes its own words still fails the coverage test. 134 suites, clippy and fmt clean.

PR #1993, opened 2026-08-09
feat(vscode): one language server per workspace folder

### LINKED ISSUE #1974: [Feature]: VS Code: Support per-folder journalFile in multi-root workspaces
## Feature request

Hi Rob,

It would be very useful if the VS Code extension could support a different `rustledger.journalFile` for each folder in a multi-root workspace.

### Use case

I manage two independent accounting ledgers (for example, HK and CA) and would like to work with both in the same VS Code workspace:

```json
{
  "folders": [
    { "path": "../HK" },
    { "path": "../CA" }
  ]
}
```

Each folder has its own root journal file, e.g.:

```
HK/
  ledger/main.beancount

CA/
  ledger/main.beancount
```

At the moment, rustledger.journalFile appears to be workspace-wide, so I cannot configure one journal root for HK and another for CA in the same VS Code window.

## Desired behavior

Ideally, the extension would support folder/resource-scoped configuration, so each workspace folder could have its own setting, for example:

```
HK/.vscode/settings.json
  rustledger.journalFile = "ledger/main.beancount"

CA/.vscode/settings.json
  rustledger.journalFile = "ledger/main.beancount"
```

or otherwise automatically associate each file with the appropriate root journal for its workspace folder.

## Why this would be useful

This would make it possible to manage multiple completely separate ledgers in a single VS Code multi-root workspace, while keeping diagnostics, completion, navigation, and other LSP features tied to the correct ledger context.

From what I can tell, this may require more than just making rustledger.journalFile resource-scoped, since the extension currently appears to start a single language-server client with one journal-file configuration. Supporting one ledger context per workspace folder (or equivalent multi-root handling) would solve the use case cleanly.

Thanks for considering it!

### Alternatives Considered

_No response_

### Area

### Beancount Compatibility

robcohen:
Closes #1974.

## Two single-ledger assumptions, not one

The reporter found the first and suspected there was more:

| | |
|---|---|
| **extension** | a module-scoped `let client`, `journalFile` declared with no `scope` (VS Code defaults that to `window`), an unrestricted `documentSelector`, and `initializationOptions` sent once at startup |
| **server** | `folders.first()` in `get_workspace_root`, and a single `journal_file` with one loaded ledger in `MainLoopState` |

**The server needs no change.** Each process is now handed exactly one workspace
folder — precisely what it already assumed — so the assumption stops being a
limitation and becomes correct by construction. A multi-root-aware server would
have meant threading a ledger identity through `MainLoopState` to buy one process
instead of N, for folders that share nothing.

## What makes the reporter's proposal work

`journalFile` becomes `scope: "resource"`, and config is read against the **root**
rather than the window — that's what makes a folder-level `.vscode/settings.json`
take effect at all.

Relative paths already worked in our favor: the server resolves a relative journal
against its workspace root, so the identical `ledger/main.beancount` in two
folders' settings resolves to two different files. Their exact proposal, with no
Rust change.

## Files outside every folder

They get a client rooted at their own **directory**, not a catch-all. Document
selectors have no "everything except" form, so a broad fallback would also claim
files a folder client already owns and both servers would answer. It also
preserves what single-file users have today, which folder-scoped clients would
otherwise have taken away.

**That fallback had a bug I only found by enumerating owners per file.** An ad-hoc
root can *contain* workspace folders — open `/w/notes.beancount` while `/w/HK` and
`/w/CA` are the folders — and a recursive `/w` selector then matches everything
those two own:

| ad-hoc glob | result |
|---|---|
| `**/*` (recursive) | `/w/HK/ledger/main.beancount` → **2 owners** |
| `*` (non-recursive) | every file → exactly 1 owner |

So an ad-hoc root's glob is non-recursive while a workspace folder's stays
recursive. The cost is a client per directory for unowned files, which beats one
document answered twice.

## Fixes something already broken single-root

Changing `journalFile` did nothing until the user found the restart command,
because it reaches the server only through `initializationOptions`.
`onDidChangeConfiguration` now restarts the affected roots, and
`onDidChangeWorkspaceFolders` stops a removed folder's server.

`findBinary` is asked once per activation rather than per client — N
missing-binary popups for N folders would be worse than one — and reset on
restart, since restarting is how a user retries after installing it.

## Verification

- `tsc --noEmit` clean (it caught a real one: the protocol's `RelativePattern`
  takes `baseUri: string`, not VS Code's `Uri`, so the selector and the file
  watcher need different spellings of the same pattern);
- `npm run build` bundles;
- the routing decision checked against a stubbed workspace — two folders plus
  outside files → four clients, every file with exactly one owner.

## Known limitation

Nested workspace folders (adding both `/w` and `/w/HK`) would double-claim, since
the outer folder's recursive selector covers the inner. VS Code discourages
nesting and `getWorkspaceFolder` returns the innermost, so document routing is
right; only selector matching overlaps. Left alone rather than solved with
exclusion globs — worth revisiting if anyone actually nests.

### DIFF
--- a/packages/vscode/package.json
+++ b/packages/vscode/package.json
@@ -1,7 +1,7 @@
 {
   "name": "rustledger-vscode",
   "displayName": "rustledger",
-  "description": "Beancount language support powered by rustledger — completions, diagnostics, formatting, and more",
+  "description": "Beancount language support powered by rustledger \u2014 completions, diagnostics, formatting, and more",
   "version": "0.1.0",
   "license": "GPL-3.0-only",
   "repository": {
@@ -76,7 +76,8 @@
         "rustledger.journalFile": {
           "type": "string",
           "default": "",
-          "description": "Path to the root journal file (e.g., main.beancount). If empty, the LSP discovers it automatically."
+          "description": "Path to the root journal file. Relative paths resolve against the workspace folder, so each folder in a multi-root workspace can point at its own ledger from its .vscode/settings.json. Leave empty to auto-discover.",
+          "scope": "resource"
         },
         "rustledger.checkForUpdates": {
           "type": "boolean",
--- a/packages/vscode/src/extension.ts
+++ b/packages/vscode/src/extension.ts
@@ -2,7 +2,7 @@ import { execFileSync } from "child_process";
 import { createWriteStream } from "fs";
 import { get } from "https";
 import { tmpdir } from "os";
-import { join } from "path";
+import { dirname, join } from "path";
 import * as vscode from "vscode";
 import {
   LanguageClient,
@@ -16,7 +16,14 @@ const GITHUB_API_URL =
   "https://api.github.com/repos/rustledger/rustledger/releases/latest";
 const VSIX_ASSET_NAME = "rustledger-vscode.vsix";
 
-let client: LanguageClient | undefined;
+// One client per ledger root, keyed by that root's URI string.
+//
+// Multi-root workspaces hold independent ledgers (#1974): two folders with
+// their own `main.beancount` share nothing, so they get a server each. That
+// also means the SERVER needs no change — `rledger-lsp` reads
+// `workspace_folders.first()` and holds one ledger, which is exactly right
+// once each process is given exactly one folder.
+const clients = new Map<string, LanguageClient>();
 // Created with `{ log: true }` below, so it's a LogOutputChannel — which is
 // also what vscode-languageclient v10's LanguageClientOptions.outputChannel
 // requires (v9 accepted a plain OutputChannel).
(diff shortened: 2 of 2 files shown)


robcohen:
Both Copilot comments fixed in 4a42be46f — the second was a real race — plus two more from re-reading the diff.

## 1. Two servers for one folder (Copilot)

`startClientForRoot` checked `clients.has(key)` and then awaited `ensureBinary` before recording anything. Two documents from the **same** folder — exactly what `startClientsForOpenDocuments` produces with `Promise.all` — both passed the guard and started a server each.

Not an edge case: that's a user with `main.beancount` and `2024.beancount` open in one folder. Reproduced against a model of the structure:

```
current (guard, then await)  : servers started for one folder = 2
fixed  (reserve before await): servers started for one folder = 1
```

Fixed by reserving an in-flight promise. The reservation is sound because it happens in the same synchronous turn as the lookup — an async function body runs synchronously up to its first `await`, so nothing can interleave between the `get` and the `set`.

## 2. The binary check ignored its argument (Copilot)

It cached a single global flag while taking `command`. `server.path` is readable per-resource, so two folders can name different binaries, and the first answer was being returned for the second. Now keyed by command — and the cache is written **before** the popup is awaited, which is what preserves the "one popup, not N" property.

## 3 and 4, from my own re-read

**A failed start poisoned the root.** The client is registered before `start()` so a concurrent caller sees it — which meant a *failed* start left a dead entry, and `clients.has` then made every later attempt a no-op. One transient spawn failure disabled that folder until the window was reloaded.

**Stopping a root mid-start leaked a server.** Removing a workspace folder while its server was still coming up found nothing in `clients`, returned, and let the start finish afterwards — a server left running for a folder that no longer exists.

## Verification

All five properties checked against a model of the fixed structure:

| property | result |
|---|---|
| 3 concurrent starts, one root | 1 server |
| 2 roots, same missing binary | 1 popup |
| 2 roots, **different** commands (one present, one missing) | 1 server, 1 popup |
| failed start | no map entry left; retry succeeds |
| stop during start | 0 clients left |

`tsc --noEmit` clean, `npm run build` bundles.

Worth noting for the record: three of these four are lifecycle bugs that only appear with more than one client, which is exactly what this PR introduces. The single-client version had nowhere for them to hide.

PR #1994, opened 2026-08-09
ci(vscode): type-check the extension, and put it in the required gate

robcohen:
Nothing in PR CI compiled, type-checked or linted `packages/vscode`. The
extension was built only by `release-build.yml` — at release time.

**And the release build would not have caught a type error either.**
`npm run build` is esbuild, which strips types without checking them. Measured
on a deliberately broken tree:

| | exit | bundle |
|---|---|---|
| old `build` | **0** | **produced** — the type error ships |
| new `build` | 1 | none |

Not hypothetical: #1993 carried a real one — the LSP protocol's
`RelativePattern` takes `baseUri: string` where VS Code's takes a `Uri` — caught
only by a local `tsc --noEmit` that no gate runs.

## Three parts

- a `typecheck` script, so CI and a developer run the same command;
- `build` runs it first, so a bundle cannot be produced from code that doesn't
  type-check;
- a `VS Code Extension` job, path-gated on `packages/vscode/**` and wired into
  the required `build` aggregator.

## Where the check sits is the load-bearing detail

It goes **ahead** of the aggregator's `should_build` early-exit, because
`should_build` tracks *Rust* changes. A vscode-only PR has it false, so a check
placed after the exit would never run on the one kind of PR it exists for — a
job that cannot gate anything, which is how the extension ended up with no CI in
the first place.

Simulated across the four shapes:

| PR | result |
|---|---|
| vscode-only, typecheck passes | pass |
| vscode-only, typecheck **fails** | **blocked** |
| rust-only (job skipped) | pass |
| mixed, typecheck **fails** | **blocked** |

## Why a separate path filter

The `vscode` filter is not folded into `code`. The extension shares nothing with
the workspace crates — it is a thin LSP client — so a Rust change cannot break
it, and Rust PRs should not pay for a Node install. That is the opposite of
`mcp-server`, which runs ungated precisely because a Rust change *can* break it
through `@rustledger/wasm`.

## Verification

`ci.yml` parses; `changes` exposes the new output; the gate check precedes the
early-exit and accepts `skipped`; `npm run typecheck` and `npm run build` both
green on a clean tree and both non-zero on a broken one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       code: ${{ steps.filter.outputs.code }}
+      vscode: ${{ steps.filter.outputs.vscode }}
       should_build: ${{ steps.decide.outputs.should_build }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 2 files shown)


robcohen:
Superseded — reopened on a `chore/` branch. `ci/` is not an allowed branch-name prefix (feat, feature, fix, docs, chore, refactor, release, hotfix, perf, claude, dependabot, copilot), and renaming a branch closes its PR anyway.

Issue #1991, opened 2026-08-08
Entries from the top-level file aren't globally sorted with included entries after load_full — breaks default date-range detection


MunsuBahk:
## Summary

When the main Beancount file contains at least one `Transaction` directive of its own (in addition to `include`d files), `rustledger`'s `load_full` (used by rustfava's `load_uncached`) does not produce a globally date-sorted entry list. The main file's own directives are appended *after* all entries pulled in via `include`, without a final sort pass. If the main file's own transaction has an early date, it ends up physically last in the entry list even though it is chronologically first (or otherwise out of order).

This breaks any code that assumes `entries` is sorted and relies on scanning from the end to find the last entry — notably rustfava's default ("no time filter") date-range computation, which does:

```python
for entry in self.entries:
    if isinstance(entry, Transaction):
        self._date_first = entry.date
        break
for entry in reversed(self.entries):
    if isinstance(entry, (Transaction, Price)):
        self._date_last = entry.date + timedelta(1)
        break
```
(`src/rustfava/core/__init__.py`, `FilteredLedger.__init__`)

Because the physically-last transaction is the early, misplaced one from the main file, `_date_last` comes out far too early. In our case this collapsed rustfava's default (no `time=` query param) view down to a single month, even though the ledger spans ~2 years, while explicitly setting a `time=` filter worked correctly (since that code path never touches `self.entries` ordering).

## Environment

- `rustfava` 1.30.12, installed via `uv tool install rustfava` (isolated venv managed by uv, not the system Python)
- `beancount` 3.2.3 — a transitive dependency pulled in automatically by rustfava's own install, not installed separately
- macOS, Python 3.13
- Rust engine downloaded per `rustfava/rustledger/engine.py` from `github.com/rustledger/rustledger/releases`

## Steps to reproduce

We hit this on a real ledger, not a constructed minimal case, so the repro shape below is the pattern we confirmed triggers it (a small synthetic 2-file version did *not* reproduce it for us on a first attempt, so the trigger may depend on the number/size of `open` directives after the transaction, or on the number of `include`d files — we did not narrow it further):

- `main.bean` (top-level file) contains ~130 `open` directives *and one* inline `Transaction` directive, dated near the start of the ledger's history.
- `main.bean` then `include`s 21 monthly journal files containing ~5600 transactions total, spanning 2020-10 through 2022-06.

Load with rustfava/rustledger and inspect the resulting entry order, e.g.:

```python
[...]

PR #1996, opened 2026-08-09
test(ffi): pin that the multi-file load stream is globally date-sorted

robcohen:
Guard for #1991. **Not a fix** — that bug is already fixed; this pins the
behaviour so it stays fixed.

## What was reported, and why it happened

rustfava 1.30.12 pins rustledger **v0.11.0**, and in that pair *nothing sorted at
any layer*: the FFI had a hand-rolled parse-and-book path rather than the
loader's `sort → synth → book → regular → finalize` pipeline, and rustfava had no
`_sort_entries` yet.

The stream therefore came out in include-expansion order — every included file's
entries, then the top-level file's own in **file** order. A consumer scanning
backwards for "the last entry" reads an early date; the reporter's two-year
ledger collapsed to a single month.

Verified not to reproduce on current versions, both through the Rust helper and
end-to-end through rustfava (196 entries, sorted, `_date_last` correct).

## Why it still needs a guard

The fix was **incidental**. The FFI stopped being unsorted because it was
rerouted onto the canonical pipeline, not because anyone set out to fix ordering.
And the guard added at that time covers only the single-source surface:

```rust
// crates/rustledger-ffi-wasi/tests/load_regular_plugins.rs
/// Directives are date-sorted before they reach consumers, regardless of file
/// order — the canonical pipeline sorts; the old hand-rolled path did not.
fn load_source_sorts_directives_by_date() { ... }
```

Nothing covered `load_file` — the path *with* include resolution, which is the
one that broke.

## Three tests, on the reported shape

A top-level file including 21 monthly journals **and** declaring its own early
transaction plus 130 `open`s:

- the stream is globally date-sorted;
- the top-level file's early entry is not stranded last (the symptom, asserted
  separately because it is what a consumer actually depends on);
- the **pad-expanded** stream stays sorted — the path rustfava actually calls,
  and it sorts by a *different* key (`d.date()` rather than `canonical_sort_key`),
  so it is its own opportunity to lose the ordering.

## What they cannot distinguish, recorded in the file

The first sabotage **did not fire**, and the docstring says so rather than
claiming more than is true. Two sorts stand between this fixture and disorder —
the pipeline's first phase and `finalize`'s re-sort — and they are redundant for
it:

| sabotage | result |
|---|---|
| remove `finalize`'s sort only | all pass |
| remove the pipeline sort only | all pass |
| remove **both** (the v0.11.0 state) | **2 of 3 fail** |
| remove `expand_pads`' sort | **the pad test fails** |

Depth rather than vacuity: the state these guard against is the one that shipped,
and they catch it. But a future change leaving only one sort would not be noticed
by these, and the file says so instead of leaving the next reader to find out.

`cargo test --workspace` — 135 suites, 0 failures. Clippy and fmt clean.

### DIFF
--- a/crates/rustledger-ffi-wasi/tests/load_file_ordering.rs
+++ b/crates/rustledger-ffi-wasi/tests/load_file_ordering.rs
@@ -51,6 +51,10 @@ use std::path::{Path, PathBuf};
 use rustledger_core::Directive;
 use rustledger_ffi_wasi::helpers::{expand_pads, load_file};
 
+/// `load_file`'s second positional argument. Named because a bare `true` next
+/// to a file about `expand_pads=true` reads as the wrong flag entirely.
+const CONFINE_INCLUDES: bool = true;
+
 fn write(dir: &Path, name: &str, body: &str) -> PathBuf {
     let path = dir.join(name);
     let mut file = std::fs::File::create(&path).expect("create fixture");
@@ -111,7 +115,7 @@ fn load_file_emits_a_globally_date_sorted_stream() {
     let dir = tempfile::tempdir().expect("tempdir");
     let main = write_reported_shape(dir.path());
 
-    let loaded = load_file(&main, true).expect("fixture loads");
+    let loaded = load_file(&main, CONFINE_INCLUDES).expect("fixture loads");
 
     assert!(
         loaded.directives.len() > 150,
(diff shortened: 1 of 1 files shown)


robcohen:
Both fixed in 53e12880c — the second one had substance.

The pad test passed `directive_lines` alone as tags, but `convert.rs` — the path this test claims to mirror — tags with `(line, file)`:

```rust
let tags: Vec<(u32, String)> =
    directive_lines.into_iter().zip(directive_files).collect();
```

`expand_pads` is generic over the tag, so a line-only `u32` compiled and passed *while exercising something narrower than the docstring promised* — the multi-file surface is precisely the one where the tag carries the originating file. Now `(line, file)`, with an assertion that a real filename survives expansion.

Sabotage-verified both ways: blanking the file component fails the new assertion, and removing `expand_pads`' sort still fails the sortedness assertion.

Also named `load_file`'s second positional argument `CONFINE_INCLUDES` — a bare `true` in a file that repeatedly discusses `expand_pads=true` reads as the wrong flag.

Worth noting this is the same failure mode the PR already documents one level down: a test that passes while testing less than it claims. The docstring said "the path rustfava actually calls" and the call didn't match it.

PR #2019, opened 2026-08-10
fix: identify compat fixtures by path, not bare filename

### LINKED ISSUE #2016: [Bug]: BQL compat identifies fixtures by bare filename, so exclusions leak and coverage is overstated
### Summary

The BQL compat layer identifies fixtures by **bare filename**, not by path. Three corpus fixtures are named `example.beancount`:

```
tests/compatibility/files/pinto-reports/example.beancount
tests/compatibility/files/fava-budget-freedom/example.beancount
tests/compatibility/files/fava-envelope/example.beancount
```

`compat-check-results.jsonl` records `"file": "example.beancount"`, and `compat-bql-test.py`'s `find_file()` resolves that with `rglob(name)` and takes `matches[0]`. Which of the three is actually queried is therefore filesystem-order dependent, and the other two are never queried at all.

### Two consequences

**1. `exclusions.toml` does not reach the BQL suite.** Exclusions are path-keyed and applied in the check step, so excluding `fava-budget-freedom/example.beancount` removes that path from the check run — but the *name* `example.beancount` still enters `compat-check-results.jsonl` from the other two copies, and `rglob` can resolve it straight back to the excluded file. Observed in #2014: the check step reported `Excluded (known limitations): 9`, and the BQL step then queried the excluded fixture anyway and reported its load error.

**2. Coverage is silently smaller than it looks.** Two of the three fixtures are never exercised by the BQL suite. `Files queried: N` counts names, not files.

### Reproduction

```console
$ find tests/compatibility/files -name example.beancount | wc -l
3

$ python3 -c "
import json
n=sum(1 for l in open('compat-check-results.jsonl') if json.loads(l)['file']=='example.beancount')
print(n)"
2                      # two entries, one name, three files on disk
```

### Suggested fix

Carry the path rather than the name: record a corpus-relative path in `compat-check-results.jsonl` and have `find_file()` resolve it directly instead of globbing. That makes exclusions propagate, makes coverage counts honest, and removes the filesystem-order dependence.

Worth checking whether any other fixture names collide once paths are available — the same silent under-coverage would apply to each.

### Context

Surfaced while excluding a fixture that neither tool can be compared on (#2014), and only visible because #2013 made partial-ledger comparisons report a load error instead of passing silently.

robcohen:
## What

Fixtures in the compatibility suite are now identified by **repo-relative path** instead of bare filename.

`compat-check-results.jsonl` recorded `file` as a basename, and `compat-bql-test.py` resolved it with `rglob(name)` → `matches[0]`. Four basenames collide in the corpus:

| basename | copies |
|---|---|
| `example.beancount` | 3 (`pinto-reports`, `fava-budget-freedom`, `fava-envelope`) |
| `example_example.beancount` | 3 (`fava-dashboards`, `fava-portfolio-returns`, `fava-portfolio-summary`) |
| `data_sample.beancount` | 2 (`gnucash-to-beancount`, `henriquebastos-gnucash`) |
| `accounts.beancount` | 2 (`beanbot`, `donearm-ledger`) |

Ten files on disk resolved to four, and *which* four was filesystem-order dependent.

## Why it matters

Two consequences are from #2016; the third turned up while fixing it.

1. **`exclusions.toml` did not reach the BQL suite.** Exclusions are path-keyed and applied in the check step, but a surviving twin put the shared *name* into the results and the glob resolved it back to an excluded file. 2 of the 10 exclusions are leakable this way — and a `sorted()` glob lands on `fava-budget-freedom/example.beancount`, which is one of them. This is exactly what #2014 observed.
1. **Six files were never queried**, while `Files queried: N` counted names and reported full coverage.
1. **The check step's own regression detection was affected too** — `previous_results[prev["file"]]` is a dict, so each collision group kept one entry and the rest were compared against an unrelated file's history or dropped. Silent, and it predates the BQL suite entirely.

## How

- `FileResult.file` is now `str(file_path)`. `TEST_DIRS` are already repo-relative and the job runs from the repo root, so no extra plumbing.
- `find_file` resolves the path directly. A path that does not exist returns `None` rather than falling back to its basename — that fallback *is* the bug. A bare name still resolves, via a now-`sorted()` glob, purely so a pre-#2016 baseline artifact stays readable.
- `KNOWN_*_DIVERGENCES` keys move to paths. All five live in `tests/compatibility/files/beancount-import/` and none collide, so it is mechanical.
- New `bare_name_registry_entries` gate: a bare-name key is a mask that can never match, so it fails the run. The check is syntactic (`"/" in key`) on purpose — it needs no corpus on disk and so cannot silently skip on a machine where fixtures were never downloaded, which is how the stale/dangling check degrades to a warning.

## Verification

Against the real corpus:

```
10 colliding fixtures on disk
resolve by NAME (old): 4 distinct -> 6 never queried
resolve by PATH (new): 10 distinct -> 0 never queried
every path round-trips exactly: True
nonexistent path -> None (no basename fallback): True
legacy bare name still resolves (old baselines): True
```

A full selection run over those ten reports `Testing against 10 files`.

**The new gate is sabotage-verified in both directions.** Reverting one registry key to its bare name:

```
self-test FAIL: shipped registries must be path-keyed, got
  [('KNOWN_PYTHON_DIVERGENCES', 'testdata_source_ofx_test_fidelity_journal.beancount')]
::error::1 deliberate-divergence registry entry keyed by bare filename.
```

Both go green again on restore. The self-test asserts the flagged case, the clean case, *and* that the shipped registries are clean.

## Expected: one quiet cycle on the baseline gates

Both baselines (`previous-check-results.jsonl`, `previous-bql-results.jsonl`) are name-keyed, so after this lands no pair matches and nothing is flagged. A missing key already means "new", not "regressed", so this **cannot produce a false failure** — only reduced sensitivity for one run, until the next publish writes path-keyed baselines. Documented at both gates.

Also expect `Files queried` to rise by 6 and the BQL run count with it.

Closes #2016

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -301,6 +301,13 @@ jobs:
 
           @dataclass
           class FileResult:
+              # Repo-relative PATH, not a bare filename. Four basenames collide
+              # in the corpus (3x example.beancount, 3x example_example.beancount,
+              # 2x data_sample.beancount, 2x accounts.beancount), so a bare name
+              # is not an identity: it made `previous_results` below silently drop
+              # all but one file per collision, let `exclusions.toml` leak (an
+              # excluded path re-entered the BQL suite under a surviving twin's
+              # name), and left 6 files never queried at all. See #2016.
               file: str
               category: str
               # Check comparison
@@ -339,7 +346,9 @@ jobs:
           def test_file(file_path: Path) -> FileResult:
               """Test a single file - designed to run in parallel."""
               category = get_category(file_path)
-              result = FileResult(file=file_path.name, category=category)
+              # TEST_DIRS are repo-relative and the job runs from the repo root,
+              # so str(file_path) is already the repo-relative path.
+              result = FileResult(file=str(file_path), category=category)
 
               # === Python parsing ===
               python_accounts = set()
@@ -564,6 +573,15 @@ jobs:
           json_parse_failures = [r for r in results if r.rust_json_parse_failed]
 
           # === Regression detection ===
+          # Keyed by `file`, a repo-relative path since #2016. It used to be a
+          # bare name and `previous_results` is a dict, so each of the four
+          # colliding basenames kept only ONE entry — the other fixtures were
+          # compared against an unrelated file's history, or dropped entirely.
+          #
+          # The first run after #2016 sees a name-keyed baseline, so no key
+          # matches and nothing is flagged (a missing key means "new", not
+          # "regressed"). One cycle of reduced sensitivity, then the published
+          # baseline is path-keyed and the comparison is exact.
           previous_file = Path(".github/badges/previous-check-results.jsonl")
           regressions = []
           if previous_file.exists() and previous_file.stat().st_size > 0:
--- a/scripts/compat-bql-test.py
+++ b/scripts/compat-bql-test.py
@@ -295,7 +295,7 @@ def _stale_registry_entries(
 def bare_name_registry_entries(
     registries: list[tuple[str, set[tuple[str, str]]]] | None = None,
 ) -> list[tuple[str, str]]:
-    """Registry keys written as a bare filename instead of a repo-relative path.
+    r"""Registry keys written as a bare filename instead of a repo-relative path.
 
     Since #2016 a fixture's identity is its path. A bare name is not merely
     unidiomatic here — it silently NEVER MATCHES any run, so the mask it was
@@ -307,6 +307,12 @@ def bare_name_registry_entries(
     on disk, so it cannot silently skip on a machine where the fixtures were
     never downloaded — which is exactly how the stale/dangling check degrades.
     Typo'd *paths* are a different failure and are caught by the dangling check.
+
+    `/` is the only separator accepted, deliberately. Run keys are POSIX
+    (`str(file_path)` on a Linux runner), so a key written with `\` separators
+    could never match one either, and flagging it here is the correct outcome —
+    treating `\` as a separator would make this guard go SILENT on exactly the
+    kind of unmatchable key it exists to catch.
     """
     if registries is None:
         registries = [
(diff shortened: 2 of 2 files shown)


robcohen:
CI is green on this branch (`Run compatibility tests` ✓). One prediction in the description was wrong and is worth correcting before review.

**I wrote "expect `Files queried` to rise by 6". It did not, and it never could have.**

`compat-check-results.jsonl` is a *list*, not a dict, so all 223 eligible files always had their own entry — `Corpus files eligible: 223` is identical before and after. The collapse happened at **resolution**, not eligibility: several entries resolved to the same twin. So the count was never the symptom. **The bug was identity, not arithmetic** — 6 files were silently tested as some other file.

What CI actually shows:

| | main (31421454634) | this PR |
|---|---|---|
| Corpus files eligible | 223 | 223 |
| Files queried (`--max-files=150`) | 150 | 150 |
| Runs tested | 2533 | 2550 |

`2533 / 17 = 149`, `2550 / 17 = 150`. One selected file previously failed to resolve at all and was dropped; now every selected file resolves. That +1 is the whole *count* effect on a capped PR run.

On the uncapped nightly the count will not move either — all 223 resolve both ways. What changes there is that 6 of them stop resolving to a twin and start being the file they claim to be. That is invisible in every number the workflow prints, which is precisely why this survived.

Two other things the run confirms:

- **The registry path-migration took.** `95% (raw 91%, 101 masked)` — unchanged from main. Had the 5 migrated `KNOWN_*_DIVERGENCES` keys stopped matching, the two wildcard entries alone would have dropped ~34 masks and raw would have converged on effective. It did not.
- **The exclusion leak is closed.** No excluded path appears anywhere in the BQL step's log, and no `Error importing` / `<load>:` line — the failure mode #2014 hit.

The new bare-name gate also ran clean in CI via `Validate stale-divergence detection logic` ✓, which now carries its own sabotage assertions.

robcohen:
Thanks — one adopted, one declined, both in `e5900639b2`.

**Adopted: the traversal escape.** Verified it is real:

```
REPO_ROOT / '/etc/passwd'     = /etc/passwd
REPO_ROOT / '../outside.bean' = /home/dev/rustledger/../outside.beancount
```

pathlib lets an absolute operand replace the base outright. Not reachable from entries we generate (always relative and clean), but `--files-from` also takes a downloaded baseline artifact, and quietly querying a file outside the corpus is the same "this is not the file it claims to be" failure the path migration exists to remove. Now rejected syntactically before any filesystem access, with self-test cases.

**Declined: treating `\` as a separator.** I think this one is backwards — it would disarm the guard rather than sharpen it. Traced through:

| key | `"/" not in key` (current) | if `\` were a separator |
|---|---|---|
| `tests\x\a.beancount` | flagged → **run fails** | valid path → **guard silent** |

Run keys are POSIX (`str(file_path)` on a Linux runner), so a `\`-separated key can never match one — it is unmatchable either way. Under the current `/`-only check it has no `/`, gets flagged, and fails the run, which is the outcome we want. Accepting `\` would reclassify it as well-formed and let it through silently, which is exactly the failure mode this guard was added to prevent.

So the current behavior is correct and the only real defect was the wording: the message said "BARE NAME" for something that isn't a bare name. It now reads `NOT A PATH ... use a repo-relative path with '/' separators`. The docstring records the reasoning, and a self-test case pins it so a future portability edit can't quietly flip it.

Both new assertions are sabotage-verified — deleting the traversal guard turns the self-test red, restoring it turns it green.

Issue #2008, opened 2026-08-10
[Compat]: parser accepts seven constructs beancount rejects, and misdiagnoses two of them


robcohen:
### Summary

rledger's parser accepts at least seven constructs that beancount rejects. A ledger containing any of them loads in `rledger` and fails in `beancount`, which is the wrong direction for a compatibility-focused reimplementation: the file looks fine here and breaks for anyone who takes it back to the reference implementation.

In two cases we do not just accept it — we silently reinterpret it and then report a **downstream balance error that misdiagnoses the cause**.

### How this was found

While classifying the unexplained BQL compatibility mismatches. On a full local corpus run (752 fixtures, 11 577 comparable runs), these seven files contribute **105 mismatches** — `bean-query` returns 0 rows because the file does not load, `rledger query` returns rows because it does.

That is why this has been invisible: it surfaces as a hundred query mismatches spread across a dozen BQL queries rather than as "our parser is too lenient". All seven fixtures come from `parser-lima`, a beancount **parser conformance** corpus — files whose entire purpose is to be rejected.

### The seven cases

`rledger` column is what `rledger check` reports. In every case the actual defect is never named.

| # | Fixture | Construct | beancount | rledger |
|---|---------|-----------|-----------|---------|
| 1 | `ParseLots.CostEmptyComponents` | `10 AAPL {, 100.0 USD, , }` | `syntax error, unexpected COMMA, expecting HASH or CAPITAL or CURRENCY` | accepted; `E3001 Transaction does not balance: residual 980.10 USD` |
| 2 | `ParseLots.CostWithSlashes` | `1.1 AAPL {45.23 USD / 2015-07-16 / "blabla"}` | `syntax error, unexpected SLASH, expecting RCURL or COMMA` | accepted; `E3001` unbalanced |
| 3 | `Transactions.TagThenLink` | `2014-04-20 * #trip "Money from CC" ^610fa...` | `syntax error, unexpected STRING, expecting end of file or EOL or TAG or LINK` | accepted, no diagnostic |
| 4 | `SyntaxErrors.ErrorInTransactionLine` | `2013-05-02 * "Dinner" A:*:B` | `Invalid token: 'A:*:B'` | accepted, no diagnostic |
| 5 | `ParseLots.CostTotalJustCurrency` | `20 AAPL {USD}` and `20 AAPL { # USD}` | `Too many missing numbers for currency group 'USD'` | accepted, no diagnostic |
[...]

PR #2020, opened 2026-08-11
fix(parser): reject transaction headers beancount refuses

robcohen:
Four of the seven constructs in #2008 — the ones that live in the transaction header, so one rule covers them all.

| # | construct | beancount | before | after |
|---|---|---|---|---|
| 3 | `* #trip "Money from CC" ^610f` | `unexpected STRING, expecting … TAG or LINK` | accepted silently | rejected |
| 4 | `* "Dinner" A:*:B` | `Invalid token: 'A:*:B'` | accepted silently | rejected |
| 6 | `* "Mermaid Inn" "Nice dinner" "With Caroline"` | `Too many strings on transaction description` | accepted silently | rejected |
| 7 | `* "A" "B" "C"` | same | accepted silently | rejected |

beancount's grammar is `DATE txn txn_strings tags_links eol` — strings first, then tags/links. More than two strings parses, and the *builder* rejects it. Our CST parser is deliberately permissive (lossless, recovers rather than rejects) and nothing downstream re-imposed that shape.

**Case 4 needed no lexer work.** In `* "Dinner" A:*:B` the `A` lexes as a one-character `CURRENCY` — a perfectly valid flag *in the flag position*, and junk after a string. Scoping the flag position to "before any string/tag/link" catches it structurally.

## One rule, two walkers

The rule is written once over `(SyntaxKind, Range)` pairs in a new `cst::txn_header`; the red and green walkers differ only in how they enumerate tokens. Green/red mirrors are the drift source CLAUDE.md names by name, so the shared rule *plus* the parity corpus is the defense.

Kind and range are all the rule needs — even the one-character `CURRENCY` flag is decided by `range.len() == 1` — so neither walker touches source text to reach a verdict, only to word the message.

## Scope came from measurement, not taste

I tallied every token kind appearing in a transaction header across the 731-file corpus (~30k transactions):

```
30189 DATE      51854 STRING     4309 TAG      216 LINK     158 TXN_KW
30022 STAR          9 PENDING_KW    3 CURRENCY  18 COMMENT   86589 WHITESPACE
    2 COLON  <-- both inside the case-4 fixture
```

So the allowlist is evidence, not guesswork. Re-sweeping after the change produces **exactly four diagnostics — the four target fixtures, zero false positives.** The 3 `CURRENCY` hits are `2012-12-17 P "Payee" "Narration"` (a real single-letter ticker flag, kept working) plus the `A` and `B` of case 4.

**Deliberately not widened past that evidence.** `2014-01-01 * * "x"` has two flags and beancount rejects it, but it is not one of the seven; guessing at extra strictness is how a compatibility fix starts breaking real files. The `|` separator keeps its own `DeprecatedPipeSymbol` diagnostic rather than being re-reported here.

## The cache bump was observed, not reasoned

After the parser change, the first `rledger check` against the fixtures still reported only the old downstream `E1001`s — the parse cache was serving the permissive parse. Loader 24 → 25, wasm 10 → 11, the wasm pin with it, and both archived-layout `FIXTURE_VERSION` tripwires. Those tripwires fire with instructions naming which case applies; this is v15's case — *which* errors are emitted changed, how anything is archived did not.

## `test_lima_tests_parse` could not have caught this

Its allowlist branch is a bare `continue`, so listing a fixture only *permits* an error — it never checks one occurs. `TagThenLink`, `CostEmptyComponents` and `CostWithSlashes` sat in that list for the parser's entire life while producing no error at all. That is how #2008 stayed invisible.

A new `LIMA_2008_CASES` table replaces that with a two-sided assertion over all seven cases. The three unfixed rows assert we are **still permissive**, so landing the rest of #2008 turns this test red and forces the row to flip. A row that can only be right is not a test.

28 allowlisted fixtures are in that same one-sided position today. 22 are `GrammarExceptions*`, whose intended semantics I have not verified — left alone and noted rather than guessed at.

## Verification

- 731-file corpus sweep: exactly the 4 target diagnostics, no false positives
- Full workspace suite green; clippy `1.97.0` (CI's version) green; `cargo doc` green
- **Sabotage-verified**: deleting the green walker's call makes `parse_green_eq_red_corpus` fail; flipping a `false` row in the strictness table fails with `expected REJECTED, got accepted (0 parse errors)`
- New parity cases include multi-byte headers, since green sums child lengths while red reads rowan ranges — a char/byte slip would show as a differing span
- End-to-end: all four now fail `rledger query` too (#1912), matching bean-query — they were returning rows here and erroring there, which is the 105 mismatches the issue was filed from

## Remaining

Cases 1, 2, 5 are the cost-spec group (`{, 100.0 USD, , }`, `{45.23 USD / 2015-07-16 / "blabla"}`, `{USD}`) and want their own PR — those are also the two where we currently emit a *misdiagnosis* (`E3001 does not balance`, a residual invented by our own recovery). The strictness table tracks them.

Refs #2008


robcohen:
All four taken, in `f29e368aaa`.

The third one is a real catch and the one I'd flag: `test_lima_2008_cases_have_expected_strictness` skipped itself when the fixture directory was missing. Those 441 fixtures are committed to the repo, so a missing directory is a broken checkout, not an unavailable dependency — and a self-skipping test is precisely the failure mode CLAUDE.md warns about under *"Availability-gated tests must fail loudly somewhere"*. Worse, skipping would hide the very thing the strictness table exists to pin. I quoted that same rule in the conformance test's module doc while leaving the skip two files away. Now a hard assert with a message that says a missing directory means a broken checkout.

The other three:

- **Doc/caller mismatch on the token range.** Right — "through (not past) the terminating `NEWLINE`" reads as *include it*, and both callers exclude it (red `take_while(!= NEWLINE)`, green breaks on it). Reworded to "up to but **not including**", and it now records why `NEWLINE` stays in the trivia set anyway: so that feeding one in cannot change the verdict, not because a caller should.
- **`unwrap_or_default()` on a non-char-boundary slice.** It would have rendered `unexpected "" in transaction header`, which names nothing — the exact sin #2008 is about. Token ranges are always char boundaries so the branch is unreachable, but it now falls back to the byte range and defect kind rather than an empty quote.
- **"three constructs" vs four cases.** Cases 6 and 7 are two fixtures of one construct (>2 header strings). The doc says that now instead of leaving the reader to reconcile it.

Re-verified after the change: parser lib, conformance and `fixture_tests` suites green; clippy `1.97.0` and fmt clean. #2022 has been rebased onto this.

Issue #1939, opened 2026-08-03
[Bug]: arithmetic expression in a cost spec is truncated to its first operand (silently wrong cost basis, then a false E3001)


robcohen:
Found by the new per-posting axis on the beancount oracle (#1902), on its first corpus run.

rledger evaluates an arithmetic expression in a **price annotation** but not in a **cost spec** — it keeps only the first operand.

## Minimal reproduction

```beancount
2013-01-01 open Assets:Something
2013-01-01 open Assets:Cash
2013-05-18 * "expression in cost and in price"
  Assets:Something   2 HOOL {10.00 * 3 USD} @ 20.00 * 2 USD
  Assets:Cash      -60.00 USD
```

| | cost | price |
|---|---|---|
| beancount | **30.00 USD** (`10.00 * 3`) | 40.00 USD (`20.00 * 2`) |
| rledger | **10.00 USD** (`* 3` dropped) | 40.00 USD ✓ |

The price is evaluated correctly, so this is not "expressions are unsupported" — it is specifically the cost spec.

## Two consequences, the second worse than the first

**Wrong cost basis.** `SELECT cost_number` returns `10.00` where beancount returns `30.00`. Any lot booked this way carries a cost basis a third of the real one, which propagates into capital gains and every cost-denominated report.

**A false E3001 on a file beancount accepts.** The weight follows the truncated cost, so rledger computes `2 × 10.00 = 20.00 USD` against `-60.00 USD`:

```
$ rledger check
E3001
  x Transaction does not balance: residual -40.00 USD
```

beancount loads the same file with **zero errors**. So a valid ledger is rejected.

## Corpus evidence

Two fixtures in the existing compat corpus hit it, from the `parser-lima` test suite:

- `test-cases_Arithmetic.NumberExprDifferentPlaces.beancount` — beancount `252.021`, rledger `120.01` (`120.01 * 2.1`)
- `test-cases_Arithmetic.NumberExprDifferentPlaces2.beancount` — beancount `252.0210`, rledger `120.010`

In the first, rledger's price for the same posting is `281.442` = `134.02 * 2.1`, evaluated correctly — the inconsistency inside a single posting.

## Where to look

[...]

PR #2022, opened 2026-08-11
fix(parser): reject malformed cost-spec component lists

robcohen:
Cases 1 and 2 of #2008. **Stacked on #2020** — review that one first; this branch contains its commits.

| # | construct | beancount | before | after |
|---|---|---|---|---|
| 1 | `10 AAPL {, 100.0 USD, , }` | `unexpected COMMA, expecting HASH or CAPITAL or CURRENCY` | accepted, then `E3001 does not balance` | rejected, cause named |
| 2 | `1.1 AAPL {45.23 USD / 2015-07-16 / "blabla"}` | `unexpected SLASH, expecting RCURL or COMMA` | accepted, then `E3001` | rejected, cause named |

A cost spec is a comma-separated component list, each component a `compound_amount`, `DATE`, `STRING` or `*`. An empty component is not a component, and `CURRENCY` completes a `compound_amount` so only `,` or a closer may follow.

## The obvious reading of case 2 would have been worse than the bug

"SLASH is illegal in a cost spec" passes both fixtures — and breaks every ledger that divides. `/` is a `number_expr` operator; `{10.00 * 3 USD}` is valid and archives 30.00 (#1939/#1942). The defect is **positional**, which is exactly what beancount's message says: *expecting RCURL or COMMA*, not "that token is banned".

`*` is left as ordinary content for the same reason — it is both the "any lot" component and multiplication, so treating it as a complete component turns `{10.00 * 3 USD}` into junk. That cost me a failing test before the corpus ever saw it.

## The corpus caught the one I got wrong

A first draft recognised only `}` as a closer and fired on **42 real `{{total}}` specs** across the 731-file sweep — `fava_investor`, `healthequity`, `tests_lots`, all valid. With `{{`/`}}`/`{#` in the brace sets:

```
--- swept 731 files, 2 cost-spec diagnostics ---
  test-cases_ParseLots.CostEmptyComponents.beancount
  test-cases_ParseLots.CostWithSlashes.beancount
```

Exactly the two targets, zero false positives. `{{...}}` is now a unit test rather than something only the sweep covers.

## No green mirror, deliberately

Unlike #2020, this check rides the existing whole-tree `COST_SPEC` scan next to `extract_unclosed_cost_brace_errors`, which sits **outside** the `use_green` branch and already runs for both paths. One walker, nothing to drift — so the absence of a mirror is the design, not an omission.

## Case 5 is deliberately still accepted

`{USD}` / `{ # USD}` is rejected by beancount from the **booker** — *"Too many missing numbers for currency group"* — a judgment across a currency group, not a property of one spec's tokens. It cannot be decided from the spec alone, so it stays out of the parser. A unit test asserts it stays clean, so the split is explicit rather than a gap, and the `LIMA_2008_CASES` table (from #2020) keeps tracking it.

## What this does not fix

`rledger check` still prints the downstream `E3001 does not balance` **alongside** the new parse error, because we keep processing where beancount stops at the syntax error:

```
P0012  parse error: empty cost-spec component before ","…
E1001  Account Assets:Invest:AAPL was never opened
E3001  Transaction does not balance: residual 980.10 USD
```

The real cause is now named first, and `rledger query` refuses the file (matching bean-query), which is the compatibility half of the issue. Fully removing the invented-cost residual means dropping the directive on a parse error — a booking-layer change, not this PR. I'd rather say so than claim the misdiagnosis is gone.

## Verification

- Corpus sweep: 2 diagnostics, both targets
- Full workspace suite green; clippy `1.97.0` green; fmt clean
- `LIMA_2008_CASES` rows for cases 1 and 2 flipped `false` → `true` — the table is two-sided, so this had to be updated to stay green
- Baseline manifest: only the 2 affected rows, source hashes asserted unchanged before applying
- Cache 25 → 26 (wasm 11 → 12): a *separate* bump from #2020's, because a cache written between the two lands carries v25 and is still stale for this change

Refs #2008


robcohen:
Both taken, in `59aa241c0e` — and the second one was more than wording, so thank you for it.

**The scan really was duplicated.** I documented the check as riding the existing whole-tree `COST_SPEC` scan and then gave it its own `descendants()` walk. So the doc was false *and* the code did the thing the doc existed to avoid: that scan already sits behind a `contains('{')` guard precisely because `descendants()` allocates a red node per node — profiling once flagged it as ~50k wasted allocations on cost-spec-free ledgers. A second identical pass doubles a cost the codebase deliberately profiled down.

It is now genuinely fused into `extract_unclosed_cost_brace_errors`' loop: one walk, and the module doc describes what the code does. An unclosed spec `continue`s past the shape check — reporting a component defect on a truncated spec is just noise about the truncation.

**The message did name the wrong delimiter.** `{{total}}` closes with `}}`, and the diagnostic said "only `,` or `}` may follow". Naming the wrong closer in a message about malformed syntax is the same species of misdirection #2008 is about. It reads "the closing brace" now.

While in there, the cost-spec path also picked up the non-char-boundary fallback the header path got from the #2020 review, rather than an empty-quoted message.

Re-verified after the restructure — the corpus sweep is the part that matters, since the check moved:

```
--- swept 731 files, 6 #2008 diagnostics ---
  CostEmptyComponents · CostWithSlashes · TransactionThreeStrings
  ErrorInTransactionLine · TagThenLink · TooManyStrings
```

All six targets, no false positives. Full workspace suite, clippy `1.97.0` and fmt clean. This PR now targets `main` (#2020 merged), so its compat and corpus-baseline gates run for real — both passed on the previous push, with **Check Compatibility at 100%**.

PR #2023, opened 2026-08-11
fix: never invent a cost number the author did not write

### LINKED ISSUE #2008: [Compat]: parser accepts seven constructs beancount rejects, and misdiagnoses two of them

robcohen:
The two booking-layer halves of #2008. They looked like separate problems and turned out to be **one root cause**: the parser manufactured a number where the source had none, and every layer downstream treated the invention as fact.

## Case 5 — the rule was already there

`{USD}` / `{ # USD}` loaded clean where beancount reports *"Too many missing numbers for currency group"*. I expected to have to build per-currency-group accounting. I didn't — interpolation already enforces "at most one unknown per currency group" and already counts a cost spec with no determinable number as one unknown.

It never fired because of one line:

```rust
Some(CostNumber::Compound {
    per_unit: pre_hash.unwrap_or_default(),        // 0 when the author wrote nothing
    total: post_hash_total.unwrap_or_default(),    // 0 when the author wrote nothing
})
```

`{ # USD}` became `Compound { per_unit: 0, total: 0 }` — a perfectly determinable **zero cost**. A zero is not an unknown, so the count rule had nothing to count. Reporting `None` when *both* sides are absent makes the existing rule reject the fixture with no change to the booker at all.

## The misdiagnosis — the same defect in different clothes

`{, 100.0 USD, , }` had its `100.0` scraped out of a spec we had *just diagnosed as malformed*, and `rledger check` then said:

```
E3001  Transaction does not balance: residual 980.10 USD
```

An arithmetic complaint about a typo, whose number this parser produced rather than the author. A malformed component list now yields no cost number at all, handing the posting to interpolation as an unknown. Both fixtures lose the `E3001` and keep the parse error naming the real cause:

```
P0012  parse error: empty cost-spec component before ","…
E1001  Account Assets:Invest:AAPL was never opened      <- legitimate, kept
```

This reuses `cost_spec_shape`'s rule rather than adding a second predicate: `first_cost_spec_defect` is now generic over the payload carried with each kind, so the diagnostic path passes byte ranges and the converter passes `()` and asks only *is this malformed*. One rule, two uses.

## Scope, and what I deliberately did not touch

Only **both** sides of the `#` missing. `{100 # USD}` and `{# 500 USD}` still default the absent side to zero; beancount treats it as MISSING and would solve for it, which is a larger change I have no oracle to check against. The corpus says it isn't urgent — `{ # CCY}` appears in exactly one file (the #2008 fixture), and the one-sided forms appear only in other parser-lima conformance fixtures.

## Two verification failures worth reading

**The new load-level test could never have failed.** It asserted only that `rledger check` fails — but the fixture's accounts are never opened, so check exits non-zero on `E1001` whatever the booker decides. Reverting the fix left it green. It now asserts the rejection *contains the too-many-unknowns message*.

**And the sabotage run that should have caught that was itself invalid.** The reverted binary kept passing because the parse cache still held the fixed build's output — reverting code doesn't change `CACHE_VERSION`. Only with `BEANCOUNT_DISABLE_LOAD_CACHE=1` does removing the fix turn the test red:

```
ParseLots.CostTotalJustCurrency must be rejected FOR THE RIGHT REASON -
expected "multiple postings missing amounts" in the output.
```

That is the same cache that hid the #2020 diagnostics — twice in one issue now.

## Verification

- New `cost_spec_unknowns.rs` runs through `load` + `process`, because the parser change only matters if the *pipeline* reaches beancount's verdict. It includes both controls that matter: a **single** numberless spec must still interpolate clean (the off-by-one that would break real ledgers), and a **well-formed** spec that genuinely doesn't balance must still report `E3001` (otherwise "no E3001" would be satisfied by having broken the validator).
- Full workspace suite green; clippy `1.97.0` and fmt clean.
- Baseline manifest: 3 rows, source hashes asserted unchanged.
- Cache 26 → 27 (wasm 12 → 13). Unlike v25/v26 this changes archived **values**, not just which errors are emitted — a stale blob would serve the invented number to a build that no longer produces one.

Closes #2008


robcohen:
This one I'm declining — I believe it's a false positive, and it's checkable rather than a matter of opinion.

Iterating `&[(&str, &str)]` yields `&(&str, &str)`, but Rust's default binding modes (match ergonomics) destructure *through* the reference: the pattern `(stem, expected)` matches the referenced tuple and binds each element as `&&str`. Both are then used in `format!`/`contains`, which take them by reference anyway, so nothing else needs to change.

The evidence is that it builds and runs:

```
test test_lima_2008_load_rejections ... ok
test result: ok. 1 passed; 0 failed
```

and `clippy --workspace --all-features --all-targets -- -D warnings` on `1.97.0` (CI's pinned version) is clean, so it isn't compiling with a suppressed lint either.

Happy to switch to `for &(stem, expected) in …` if you'd rather see the reference destructured explicitly — it's equivalent and arguably clearer about the binding mode — but the current form is correct as written, so I'd rather not churn the diff on a compile error that doesn't exist.

Issue #2018, opened 2026-08-10
[Bug]: valuation plugin reimplements FIFO instead of delegating to the booking engine


robcohen:
### Summary

In the valuation plugin's FIFO sell path, the **posting reports a rounded quantity while the internal lot is decremented by the unrounded one**. The two drift, and the drift only becomes visible much later, when a full close-out posts a lot balance that is already short and leaves a phantom residual in the ledger.

`crates/rustledger-plugin/src/native/plugins/valuation.rs`, partial-sell branch:

```rust
let units_to_sell = remaining / current_price;                        // unrounded
let rounded_units = round_down(units_to_sell, MAPPED_CURRENCY_PRECISION);

postings.push(PostingData {
    units: ... format_decimal_fixed(-rounded_units, ...)              // ledger sees ROUNDED
    ...
});

lot.units         -= units_to_sell;                                   // state uses UNROUNDED
state.total_units -= units_to_sell;
```

The **buy** path in the same file already does this correctly — it rounds first and then uses that one value for both the lot and the posting:

```rust
let synthetic_units = round_up(units_number / state.last_price, MAPPED_CURRENCY_PRECISION);
state.lots.push(CostLot { units: synthetic_units, ... });
new_postings.push(... format_decimal_fixed(synthetic_units, ...) ...);
```

So the fix is to make the sell path match the buy path: decrement by `rounded_units`, i.e. by what was actually posted.

### Observed against beancount

`tests/compatibility/files/beancount-lazy-plugins/tests_data_some_fund_example.beancount`, the lot bought at cost `1 USD`:

```
                rledger                    bean-query (beancount_lazy_plugins)
2024-01-10       500                        500.0000000
2024-02-12      -374.9999999               -374.9999999
2024-03-07      -125                       -125.0000001
                ------------                ------------
sum              0.0000001                  0
```

After the partial sell, `500 − 374.9999999 = 125.0000001` remains. beancount's plugin closes the lot with exactly that and the lot reaches zero. Ours posts `-125`, because our lot balance had already been decremented by the unrounded quantity, so it believes only `125` remains. The `1e-7` difference is left behind as a position that should not exist.

Reproduce:

```console
$ F=tests/compatibility/files/beancount-lazy-plugins/tests_data_some_fund_example.beancount
$ rledger query "$F" "SELECT date, number WHERE account ~ 'SomeFund:Total' AND cost_number = 1 ORDER BY date"
```

[...]

Issue #2015, opened 2026-08-10
[Bug]: query output renders a commodity at 0 dp, printing -375 for -374.9999999


robcohen:
### Summary

Two findings on `tests/compatibility/files/beancount-lazy-plugins/tests_data_some_fund_example.beancount`, both about a `1e-7`-scale residual that should survive to the end of the ledger.

The first is an **internal inconsistency in rledger** and does not involve beancount at all:

```console
$ rledger query "$F" "SELECT account, SUM(number)   WHERE account ~ 'SomeFund:Total' GROUP BY account"
Assets:SomeFund:Total  0.0000001

$ rledger query "$F" "SELECT account, SUM(position) WHERE account ~ 'SomeFund:Total' GROUP BY account"
Assets:SomeFund:Total          <- empty inventory
```

Same account, same ledger. `SUM(number)` says `0.0000001` of `SOME_FUND_USD` remains; `SUM(position)` says the inventory is empty. One of the two is wrong.

beancount agrees with our *number*, not our *inventory*:

```
bean-query  SUM(position) -> 0.0000001 SOME_FUND_USD { 1 USD}
```

### Second finding: PnL differs by 1e-8

Same fixture, same query:

```
                        bean-query                    rledger
Income:SomeFund:PnL     -300.00000001                 -300.0000000000000000000000000
```

We produce exactly `-300`; beancount produces `-300.00000001`. This one *is* plugin arithmetic — the fixture loads `plugin "beancount_lazy_plugins.valuation"` and rledger runs its native reimplementation (`crates/rustledger-plugin/src/native/plugins/valuation.rs`), so the two implementations round differently somewhere in the value-denominated FIFO.

### What this is NOT

Worth recording, because the surrounding mismatches look alarming and are not:

- **`0.00` vs `1E-7` is rendering, not arithmetic.** On the sibling fixture `tests_data_output_cool_fund_output.beancount`, `SUM(position)` shows `0.00` for beancount and `0.0000001` for us. Asked for the raw number both report the same value, beancount as `1E-7` and rledger as `0.0000001`. The `0.00` was beancount rendering at the position's display precision. That fixture also has **no plugin directives** — it is a static ledger, and the arithmetic is exact: `1000.0000000 − 454.5454545 − 545.4545454 = 0.0000001`. Both tools are right.
[...]

PR #2030, opened 2026-08-11
fix(query): never render a number the ledger does not hold

robcohen:
> **This is a decision, not a merge-ready fix.** It corrects a real defect and costs measurable compat. The numbers are below; the call is yours.

#2015: a column of `SOME_FUND_USD` rendered `-374.9999999` as **`-375`**. Not a rounded total, not a blanked cell — a number the ledger does not contain, printed with nothing to signal it.

## I got an oracle instead of guessing

The issue flagged the deciding question as unanswered: *"beancount prints `0.0000001 SOME_FUND_USD` here, so either its context gives this commodity ≥7 dp or its renderer does not suppress below precision. That needs checking on the beancount side before picking a fix."*

I installed `beancount` + `beanquery` + `beancount_lazy_plugins` locally and asked it:

```
bean-query                                    rledger (before)
 500.0000000 SOME_FUND_USD { 1 USD}            500 SOME_FUND_USD { 1 USD}
-374.9999999 SOME_FUND_USD { 1 USD}           -375 SOME_FUND_USD { 1 USD}
 700         USD                               700 USD
```

bean-query renders **per currency, at the widest precision present in the column** — `SOME_FUND_USD` at 7 dp, `USD` at 0 dp. So the per-column context is right in principle, and `MostCommon` was the wrong *policy* for it. A column renderer aligns a column; it does not rewrite the values in it.

## The obvious fix is wrong, and the measurement says so

Flipping to raw observation + `Maximum`: raw compat **87% → 75%** on a 1020-run local sample.

The reason is worth recording: **our arithmetic carries scale beancount's does not.** We compute `15152.070000000000000000000000` where beancount computes `15152.07`. Of the 131 runs that moved, **114 were numerically identical** and differed only in trailing zeros.

Observing `normalize()`d values was worse still (**67%**) — it discards *authored* trailing zeros, which in beancount's model declare precision (`500.0000000`).

## The rule that works

Round to the ledger's precision when that **doesn't change the number**; widen only when it does. The comparison is numeric, so scale is ignored:

| value | ledger dp | quantized | equal? | column |
|---|---|---|---|---|
| `15152.070000000000000000000000` | 2 | `15152.07` | yes | stays 2 dp |
| `-374.9999999` | 0 | `-375` | **no** | widens to 7 dp |

Verified against the oracle: the `SOME_FUND_USD` column now matches bean-query value for value.

## What it costs — measured, not estimated

Net **−24 runs** on the 1020-run sample (raw 87% → 84%). 36 regress, 12 improve. Of the 36:

- **17 are cases where we genuinely hold a different value than beancount** — `0.5569686372085825412164812257` vs `0.56`, `5293.505` vs `5293.50`. The old rounding was hiding these.
- **19 are collateral** — per-currency precision means one such value widens every value of that currency in the column. beancount's model has the same property.

## The finding that matters more than the fix

**The display rounding was masking real arithmetic divergence across the corpus**, not just mis-rendering one fixture — 17 hidden divergences in a 60-file sample. That is the same mechanism by which this issue hid #2018: a `1e-7` discrepancy is invisible in a column rendered at 0 decimal places.

So the headline compat number is currently flattered by rounding. Whether to take the honest number now or sequence it after the arithmetic work is a judgment call, which is why I've stopped here rather than merging.

## Options

1. **Land it** — accept ~−2 pts raw compat, gain a renderer that never lies, and surface the 17 for follow-up.
2. **Hold it** — file the 17 divergences first, fix those, then land this with a much smaller delta.
3. **Drop it** — decide the current rounding is the behaviour we want; I'd argue against, since it prints values the ledger does not hold.

I lean 2: the branch is ready either way, and landing it after the arithmetic fixes costs almost nothing.

`cargo test -p rustledger` (including the CLI snapshot tests) is green.

Refs #2015

### DIFF
--- a/crates/rustledger/src/cmd/query/output.rs
+++ b/crates/rustledger/src/cmd/query/output.rs
@@ -79,6 +79,23 @@ fn write_text<W: Write>(
     // a commodity's own `render_commas:` declaration unreadable here (#1896).
     for col in &mut col_contexts {
         col.adopt_grouping_from(ctx);
+        // A COLUMN renderer takes the widest precision in the column, not the
+        // most common one. That is what bean-query's `DecimalRenderer` does,
+        // and it is the difference between aligning a column and rewriting the
+        // values in it.
+        //
+        // `MostCommon` is right for a LEDGER context — "how does this currency
+        // usually get written" — and wrong here. On `SOME_FUND_USD` the
+        // observed decimals are `0 0 7 2 0 0 2`; the mode is 0, so a column
+        // holding `-374.9999999` rendered it as `-375` (#2015). Not a rounded
+        // total, not a blanked cell: a number the ledger does not contain,
+        // printed with nothing to signal it. Under `Maximum` the column is 7 dp
+        // and every value in it prints as itself.
+        //
+        // Scoped to these per-column contexts. The ledger context `ctx` keeps
+        // `MostCommon`, so `print`/`format` and every other surface are
+        // untouched — this is a query-rendering policy only.
+        col.set_precision(rustledger_core::Precision::Maximum);
     }
     let mut col_inherited: Vec<bool> = vec![false; result.columns.len()];
     for row in &result.rows {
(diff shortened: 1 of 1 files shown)


robcohen:
Closing — the premise is wrong.

I verified against a local beancount install that bean-query renders positions at the **ledger's per-currency precision**, not at a per-column maximum: it holds `5293.505` and prints `5293.50 BRL`. The 7-dp `SOME_FUND_USD` column that suggested otherwise is 7 dp in beancount's *ledger* context, because its valuation plugin emits uniform precision where ours emits ragged.

So `MostCommon` is the correct policy for this surface and this PR trades real agreement (raw 87% → 84%) for a model beancount does not use.

Full correction on #2015. The one genuine divergence the sweep did find is #2031.

Issue #1104, opened 2026-05-13
compat(bql): zero-amount balance rows render an explicit '0.00 USD' instead of bean-query's blank cell


robcohen:
## Problem

When a query produces a balance row whose total nets to zero, `bean-query` renders the value cell as **empty** (account-only row), while rustledger renders it as **`0.00 USD`**. This causes **2 of 510 BQL compat tests** to fail (5% of BQL compat gap), both on the same file and same row but via two different queries.

## Affected file + queries

`testdata_source_ofx_test_non_default_capital_gains_journal.beancount`

Both queries hit the same row 3 of `Income:Vanguard:CapitalGains:VANGUARD-92202V351`:

| Query | Python | Rustledger |
|-------|--------|------------|
| `SELECT account, SUM(position) GROUP BY account ORDER BY account LIMIT 10` | `Income:Vanguard:CapitalGains:VANGUARD-92202V351` (no value) | `Income:Vanguard:CapitalGains:VANGUARD-92202V351 0.00 USD` |
| `BALANCES FROM has_account('Assets')` | same: account only | same: account + `0.00 USD` |

The account's total position is mathematically zero; bean-query suppresses the value cell entirely, rustledger renders the zero.

## Same underlying bug, two queries

These are not two distinct issues — same file, same row, same shape, same cause. The fix likely touches the `SUM(position)` aggregate's output emission or the `BALANCES`-flavored renderer's zero-suppression logic. Fixing one likely fixes both.

## Impact

5% of the BQL compat gap (smallest of the four buckets). Low priority — both renderings convey the same information; bean-query's blank-on-zero is arguably less readable.

## Suspected root cause

`bean-query` likely treats a position whose net Decimal is exactly zero as "no position" for display purposes. rustledger's `format_value` path renders `Value::Inventory` with all positions, including those that net to zero.

Either:
- The `SUM(position)` aggregator should drop zero-net positions before emitting the row, or
- The `Value::Inventory` formatter should skip zero-amount positions.

## Fix shape

Audit `crates/rustledger-query/src/executor/` aggregator output for `SUM(position)` and `BALANCES` — drop zero-net positions before they reach the formatter. Or, alternatively, apply the suppression in `format_value()` itself for `Value::Inventory` rendering when a position's units equal zero.

The aggregator-level fix is more correct (Python's semantics: zero-net positions are absent, not present-but-zero). The formatter-level fix is more cosmetic.

## Test plan

[...]

PR #2033, opened 2026-08-11
fix(query): render a position the inventory holds, even at 0.00

robcohen:
The "zero-sum cluster" from the nightly tail. It turned out not to be about the zero-sum plugin at all — the accounts agree; the rendering doesn't.

## #1104's premise was wrong

It blanked any position whose units round to zero at the currency's display precision, stating that bean-query blank-cells a residual like `-0.0003183 USD`. Checked against bean-query 3.2.3:

```
; Assets:A netted to nothing; Assets:B holds 0.0003183 USD
bean-query "SELECT account, SUM(position) ..."
  Assets:A            <- blank: the inventory is EMPTY
  Assets:B  0.00 USD  <- shown: the position exists, rounded to render
```

**beancount blanks an empty inventory, not a small number.** Rounding to `0.00` is display precision doing its job; suppressing the row instead hid positions that are really there.

## The other half of the predicate is load-bearing — I checked

An *exactly* zero position stands in for a difference in the **inventory**, not a display rule: beancount removes a lot when it reduces to exactly zero, so `0 ORNG {1 USD}` is simply absent and the cell is blank, while ours retains the closed lot.

I tried dropping the whole filter first. Measured: **492 → 459** raw matches on the reds-plugins + regressions fixtures, with diffs like

```
py: Assets:Brokerage
rs: Assets:Brokerage 0 ORNG { 1 USD}
```

So the `is_zero` test stays, now with a comment saying to delete it the day `Inventory` drops zero lots itself.

## Result

Keeping `is_zero`, dropping only the rounding test:

| set | before | after | |
|---|---|---|---|
| reds-plugins + regressions (544 runs) | 492 | **500** | 8 improved, **0 regressed** |
| broad 60-file sample (1020 runs) | 888 | 888 | no change |

Six of the eight are the zero-sum fixture, where we reported nothing for two accounts bean-query reports as `0.00 USD`.

## Tests

The three #1104 tests asserted the disproven behavior, so they are **inverted rather than deleted** — same fixtures, now pinning that the residual renders as `0.00 USD`. Renamed from `..._suppresses_sub_precision_positions` to `..._renders_sub_precision_positions` so the name matches the claim.

## Known remainder

bean-query prints `-0.00 USD` for one account where we print `0.00 USD`. That is **signed zero**, which `rust_decimal` does not carry the way Python's `Decimal` does. Left alone — it is a representation difference, not a rendering rule, and chasing it means giving `Decimal` a signed zero.

Full workspace suite green; clippy `1.97.0` and fmt clean.

Refs #2015

### DIFF
--- a/crates/rustledger/src/cmd/query/output.rs
+++ b/crates/rustledger/src/cmd/query/output.rs
@@ -347,26 +347,38 @@ fn update_column_context(col_ctx: &mut DisplayContext, value: &Value, ledger_ctx
 /// `DisplayContext::format`, whose unknown-currency fallback calls
 /// `normalize()` and *strips* trailing zeros (`0.000` → `0`), making
 /// output worse than the pre-fix state.
-/// Whether a `Position`'s units, when rounded to the currency's tracked
-/// display precision, equal zero.
+/// Whether a `Position` should be omitted from `Value::Inventory` rendering.
 ///
-/// Used to suppress sub-cent residuals (rounding artifacts of cost-spec
-/// interpolation) from `Value::Inventory` rendering. Pure mathematical
-/// zero is already filtered via `Position::is_empty`; this also catches
-/// `-0.0003183 USD` (the kind of capital-gains rounding residual that
-/// matches bean-query's blank-cell display for SUM(position) / BALANCES
-/// outputs when the underlying value is below currency display
-/// precision). Matches Python beancount's behavior of suppressing
-/// zero-valued positions in aggregate output (#1104).
-fn position_renders_as_zero(pos: &rustledger_core::Position, ctx: &DisplayContext) -> bool {
-    if pos.units.number.is_zero() {
-        return true;
-    }
-    if let Some(dp) = ctx.get_precision(pos.units.currency.as_str()) {
-        pos.units.number.round_dp(dp).is_zero()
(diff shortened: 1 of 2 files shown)


robcohen:
Both taken, with one correction to the first.

**The test rustdocs were plainly wrong** — two of them still said sub-precision positions are "suppressed"/"blanked" directly above assertions that now require `0.00 USD`. Rewritten to state the corrected rule and cite the bean-query 3.2.3 check.

**On the helper comment**, the premise needs adjusting: `Inventory::positions()` is *not* filtered — it is `self.positions.iter()`. The `!p.is_empty()` filter lives on the `Display` impl, which is a different path. That is precisely why the coalesced zero reaches this filter and has to be dropped here; if `positions()` did filter, the helper would be dead code.

Your substantive point stands, though, and the comment is better for it: "closed lot" was loose. It now says what actually happens — `Inventory::add` coalesces same-key lots, so a lot bought and fully sold nets to a `0 ORNG {1 USD}` entry that stays in the map, while beancount drops the key at that point. I added the `positions()`-is-unfiltered note so the next reader does not have to re-derive why the filter is reachable.

No behavior change; tests, clippy `1.97.0` and fmt still clean.

PR #2104, opened 2026-08-20
perf: recover the cache-hit regression, size the postings vector, benchmark the shipping config

robcohen:
Seven changes from profiling `rledger check`. Two paths matter and they had
moved in opposite directions since v0.21.0, which is what most of this
addresses.

## Results

`rledger check` on the 10k-transaction ledger `bench.yml` generates, release
+ mimalloc (the configuration release-build.yml ships):

| | v0.21.0 | main | this branch |
|---|---|---|---|
| **warm** instructions | 176.8M | 215.7M (+22%) | **177.3M** |
| **cold** instructions | 797.3M | 578.1M | **552.4M** (-31%) |
| warm wall clock | 48.1 ms | 47.9 ms | **38.7 ms** (min 34.5) |
| peak RSS | | 42.2 MB | **33.9 MB** (-18%) |
| parse cache file | | 5.62 MB | **3.62 MB** (-36%) |

The cache-hit path had regressed 22% since v0.21.0, and that is the path the
nightly benchmark actually measures: `bench.yml` runs `hyperfine --warmup 3`
with the parse cache enabled, so every timed run is a cache hit that never
parses. It is now back to parity on instructions and ~20% better in wall
clock, because the memory reduction does not show up in an instruction count.

Cold is 31% better than v0.21.0 — that is #2056-#2067 turning quadratics into
linear scans, plus the parser change here.

## The changes

**Layout**

1. **`perf(core)`: box `Posting::cost` and `Posting::price`.** `Posting` was
   280 bytes and a plain two-posting expense carried ~190 bytes of `None`.
   rkyv mirrors the in-memory layout, so all of that padding was written to
   the cache, then validated and materialized on every hit — the archive ran
   5.2x the size of the ledger. Now 160 bytes; cache -36%, RSS -11%.
   Cost-bearing ledgers are the case that could have regressed, since their
   boxes are really allocated: `investment` -0.1%, `fifo` +0.03%. Loader cache
   v29 / wasm v15.

**Cache-hit path**

2. **`perf(loader)`: intern while deserializing instead of deduplicating
   after.** rkyv's deserializer carries no interner, so `AsInternedStr` gave
   every occurrence its own `Arc<str>` — 40,015 for a ledger with a few dozen
   distinct strings — and `reintern_directives` then walked every directive
   again, 50,015 interner calls, to collapse them. A scoped `InternScope`
   interns on the way in, establishing exactly the postcondition that pass
   guarantees, so the cache-hit path no longer runs it. Scoped rather than
   process-wide: a thread-local behind `From<&str>` was the first thing I
   tried and it accumulates every payee a long-running LSP ever sees.

3. **`perf(loader)`: decode cached source text the way the loader does.**
   `into_load_result` used `from_utf8_lossy(&bytes).into_owned()` under a
   comment saying it matches `DiskFileSystem::read` — which uses
   `String::from_utf8` with a lossy fallback. The slow form validates a byte
   at a time and copies even when every byte was valid: 22.8M instructions,
   12.9% of the cache-hit run.

4. **`refactor(cli)`: delete the write-only `SourceCache`.** `new` and `add`,
   no reader, private field. `check` filled one every run and dropped it.
   `Utf8Chunks::next` was the hottest single function in the profile at 9.8%.

**Everywhere**

5. **`perf(core)`: drop a tautological ceiling test.**
   `checked_add(..).is_some_and(|sum| sum <= Decimal::MAX)` — a `Some` from
   `checked_add` is representable by construction, and `Decimal`'s comparison
   aligns scales before answering. Twice per posting. 11.0M -> 5.9M.

6. **`perf(parser)`: size the postings vector.** 10,000 per-transaction
   overshoots. Cold: `simple` 81.0 -> 71.2 ms, `investment` 120.3 -> 105.1,
   `fifo` 116.0 -> 97.0.

7. **`ci(bench)`: benchmark the allocator we ship.** bench-pr.yml built both
   sides without `--features mimalloc` while release-build.yml ships it. An
   interning change measured **-14% on glibc and 0% on mimalloc**; it was
   dropped rather than shipped, but the job as written would have endorsed it.
   Also adds the sub-phases of `throughput/10k_transactions`: lexing 2.4 ms,
   green tree 13.8 ms, CST-to-AST conversion 40.8 ms.

## Measured and rejected

- **rkyv unchecked deserialization**: worth 7.8M (-4.3%), and declined. It
  needs `#![forbid(unsafe_code)]` dropped from the loader, and a payload
  checksum does not rescue it — unkeyed, in a user-writable directory next to
  the ledger, so anyone who can write the cache can write a matching checksum.
  Today a crafted cache is safely rejected; unchecked it is UB in a process
  that also runs as an LSP and MCP server.
- **True zero-copy** (`rkyv::access`, checked, no unsafe) would remove the
  ~15M materialization, but the pipeline mutates directives — booking fills in
  units, interpolation rewrites amounts — so it cannot run on `&Archived*`.
- **Always opening the undo log** (dropping `rollback_needed`): 196.2M vs
  192.9M, **4% worse**. The guard still earns its place even though the
  snapshot it was written for (#1897) is now a cheap undo log.
- **Sizing the top-level `directives` vector**: no gain. ~13 amortized
  reallocations for a whole file, unlike 10,000 per-item overshoots.

## Verification

- `cargo test --workspace --no-fail-fast`: passes except the known
  pre-existing `corpus_loads_identically_via_native_and_component` (stale
  debug wasm).
- `report balances` byte-identical cold vs warm on `simple`, `investment`,
  `fifo` and bench.yml's ledger, including the 5,964-line `fifo` report.
- The `CostNumber` byte fixtures and `MetaValue` layout hash fired when
  `CACHE_VERSION` advanced; their layouts genuinely did not change, so only
  `FIXTURE_VERSION` moved and both still assert against their original bytes.
- New tests assert their negative half first so they cannot pass vacuously;
  the overflow/headroom tests were shown to FAIL when `fits` is broken.
- `cargo clippy --workspace --all-targets` on 1.97.0: exit 0, zero warnings.
  `actionlint` clean.


robcohen:
## Self-review, correctness focus

Four findings, all fixed in `bb206a18cc`. The first is the one that mattered.

### 1. A test that could not fail (the one real defect)

`the_intern_scope_stops_interning_once_the_guard_drops` built its values with
`InternedStr::new`. That constructor **does not consult the scope** —
interning happens inside `AsInternedStr::deserialize_with` and nowhere else.
So the test asserted that two freshly allocated `Arc`s differ, which holds
whether `InternScope` works or not. It passed for a reason unrelated to what
it claimed to check.

Found by writing a *new* test (nesting) that failed, then asking why the
existing one passed.

Both scope tests now deserialize a real archive, and all three fail when
`intern_scoped` is stubbed to return `None`.

### 2. The equivalence claim was tested more narrowly than it was made

Skipping `reintern_directives` on the cache-hit path is only sound if the
scope covers everything that pass covers — accounts, currencies, payees and
narrations. The test checked `Posting::account` alone. `account` reaches
`AsInternedStr` through the `Account` newtype and `currency` through `Amount`,
neither naming the wrapper at the field, so covering one does not imply the
others.

Verified all four empirically, then pinned all four in the test, in both
directions.

### 3. A stale comment this branch created

`Loader::load` said the cache-hit path "already runs `reintern_directives`" —
made false by this branch — and pointed at `check.rs`, where that call had not
lived for some time. Corrected to describe the three paths as they are.

### 4. Nesting was documented but untested

`InternScope::new` returns a guard whether or not it installed the table; only
the `installed` flag stops an inner guard's `Drop` from clearing it and
silently ending interning for the rest of the outer scope. Now tested.

## Checked and found sound

- **`ptr_eq` is not load-bearing.** It appears only as the fast path inside
  `InternedStr`'s `PartialEq`, which falls back to byte comparison. So even a
  scope coverage miss would have been a performance regression, not a wrong
  answer.
- **The tautology removal.** `checked_add` returns `None` exactly when the sum
  is unrepresentable, so any `Some` is a `Decimal` and therefore `<=
  Decimal::MAX`. Both operands are `abs()`-ed, so sign cannot matter either.
- **The UTF-8 fallback.** `FromUtf8Error::as_bytes()` returns the bytes that
  were attempted, so the lossy branch produces the identical string.
- **Boxing did not move any external contract.** `rledger query --format json`
  over a cost-bearing ledger is byte-identical before and after (64,670 bytes,
  200 rows, exit 0). `meta.hash` is computed by `rustledger-ops::fingerprint`
  from `CostData` / `PriceAnnotationData`, the plugin wrapper types, which are
  not boxed.
- **The deleted `failed to read` path.** Removing `SourceCache` also removed a
  `?` on reading the main file. Verified against a deleted ledger with a live
  cache: both old and new report `file not found` and exit non-zero, because
  an earlier existence check catches it.
- **`as_deref().is_some_and(..)`** operates on `Option<&CostSpec>`, which is
  `Copy`, so nothing moves out of `posting.cost`.
- **`map_or_else(CostSpec::empty, |boxed| *boxed)`** matches the previous
  `unwrap_or_else(CostSpec::empty)` on both arms, and `.take()` still clears
  the field.
- **Cache versioning.** Loader v29 and wasm v15 both moved. The `CostNumber`
  byte fixtures and `MetaValue` layout hash still assert against their
  original bytes, which is what proves those inner layouts did not change.
- **The WASM path is correctly untouched** — it deserializes without a scope
  and still calls `reintern_plain_directives`.

## Not fixed — needs a decision

**This breaks the public API.** `Posting::cost` and `Posting::price` change
type, and `report::SourceCache` is removed. `rustledger` publishes to
crates.io, so embedders reading `posting.cost` need `as_deref()`. Worth a
CHANGELOG breaking-change entry before release, which I have not written.

PR #2105, opened 2026-08-20
fix(booking): make multi-currency interpolation deterministic; build the line index lazily

robcohen:
Two changes on the cache-hit path after #2104. The second started as an
allocation cleanup and turned up a real bug, so it leads.

## 1. Multi-currency interpolation was nondeterministic

`interpolate_with_tolerance_map` accumulated residuals in a
`std::collections::HashMap`. `RandomState` is seeded per process, so
`residuals.iter()` yielded a different order every run — and the
multi-currency split picks `non_zero_residuals[0]` to fill the original
posting. The same ledger emitted its postings in a different currency
order from one run to the next:

```
2024-01-02 * "multi-currency absorb"
  Assets:A   10.00 USD
  Assets:B   20.00 EUR
  Equity:X
```

`report balances` on that gave **two distinct outputs across 30 runs**
before this change and **one** after. Amounts always agreed; only the
order moved — enough to make a diff churn and any snapshot flake, and to
change which posting index carries which currency.

The codebase already knew about this in one place: where
`missing_by_currency`'s keys are scanned for a diagnostic they are
sorted first, under a comment saying unsorted order "would produce
non-reproducible test output". That covered one of three iterations. The
consuming iteration that decides fill order, and the residual scan, were
not covered.

`Residuals` and `MissingByCurrency` are now ordered by construction —
insertion order, i.e. the order the transaction introduces each currency
— so every use site is deterministic without anyone remembering to sort.

**Why the existing test missed it.**
`test_interpolate_multi_currency_three_currencies` asserts the posting
count and that residuals are ~zero, and deliberately says nothing about
which currency landed where. That is exactly the thing that varied. The
new test asserts the order.

Both new types are `SmallVec`, which is where this started: a
transaction has one currency, occasionally two, and both maps were built
with `with_capacity`, so they allocated whether used or not — ~30,000
allocations on a 10,000-transaction ledger. Worth 169.2M -> 168.2M
instructions (-0.6%), which is real and much less important than the
ordering.

## 2. The source-map line index is built on first use

`SourceFile::new` scanned the whole source for newlines and kept a
`usize` per line. Every consumer is a diagnostic — `line_col`, `line`,
`line_start`, `num_lines` — so a ledger that reports nothing built the
table and never read it. 6.8M instructions and a 320 KB table on a
40,000-line ledger.

`line_col` also gets a fix it has always wanted: it located a line with
`rposition`, a **linear scan**, over a table sorted ascending by
construction. One diagnostic never notices; a file reporting thousands
pays O(lines) each time. `partition_point` is the same answer by binary
search — the predicate is monotone over sorted data, so the count of
entries satisfying it is one past the last that does, and entry 0 is
always 0, so it cannot underflow.

| | before | after |
|---|---|---|
| warm `check` | 176.8M | **168.2M** (-4.9%) |
| cold `check` | 552.4M | **543.3M** |
| wall clock (warm) | 39.3 ms | **36.3 ms** |

## Verification

- Determinism: 30 runs, 2 distinct outputs before, 1 after.
- Line/col: a 1,609-line ledger with 59 errors from line 6 to line 1606,
  including the final transaction — `check` output byte-identical,
  spans, carets and rendered excerpts included. The comparison is not
  vacuous: that output carries 59 `file:line:col` references.
- Workspace tests pass except the known pre-existing
  `corpus_loads_identically_via_native_and_component` (stale debug wasm).
- `cargo clippy --workspace --all-targets` on 1.97.0: exit 0, zero
  warnings. It caught a real slip on the way — an inserted test had
  landed between an existing `#[test]` and its function, silently
  disabling `test_interpolate_multi_currency_three_currencies`.

## Noted while here

`cache::tests::test_invalidate_cache` and
`..._removes_legacy_sidecar` failed once under the full workspace run
and passed on re-run, on this branch and on main, and pass 3/3 in
isolation. Pre-existing test-isolation flakiness under cross-crate
parallelism, not from this change. Worth its own issue.

See also the correction comment below on the mimalloc finding from the
earlier revision of this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3425,6 +3425,7 @@ dependencies = [
  "rustc-hash 2.1.3",
  "rustledger-core",
  "serde_json",
+ "smallvec",
  "thiserror 2.0.20",
 ]
 
--- a/crates/rustledger-booking/Cargo.toml
+++ b/crates/rustledger-booking/Cargo.toml
@@ -13,6 +13,7 @@ authors.workspace = true
 readme = "README.md"
 
 [dependencies]
+smallvec.workspace = true
 rustledger-core.workspace = true
 rust_decimal.workspace = true
 rust_decimal_macros = "1.40"
--- a/crates/rustledger-booking/src/book.rs
+++ b/crates/rustledger-booking/src/book.rs
@@ -1345,6 +1345,52 @@ impl BookingEngine {
         Ok(self.book_and_interpolate_with_gains(txn)?.0)
     }
 
+    /// Book, interpolate and apply `txn` in one step, mutating it in place.
+    ///
+    /// The three are one step because they share a failure rule: a transaction
+    /// that fails any of them is set aside by `run_booking` and merged back
+    /// into the ledger by `finalize`, so a user sees it, and it must read as
+    /// the author wrote it. Splitting them would hand the caller a
+    /// half-interpolated transaction to put back.
+    ///
+    /// This exists so the common path stops cloning. `book_and_interpolate_
+    /// with_gains` hands back an owned transaction the caller assigns over the
+    /// original, so every transaction paid a full clone and the original's
+    /// drop — 6.9M and 6.0M instructions on a 10,000-transaction ledger, 7.7%
+    /// of a warm `check` — to keep a rare failure undoable. Interpolation
+    /// touches one or two postings, so `Rollback` saves those instead.
+    ///
+    /// # Errors
+    ///
+    /// Any booking, interpolation or apply failure. On every one of them
+    /// `txn` is left exactly as it was passed in.
+    pub fn book_interpolate_apply(
+        &mut self,
+        txn: &mut Transaction,
+    ) -> Result<Vec<CapitalGain>, BookingError> {
+        // Fast path, as in `book_and_interpolate_with_gains`: with no cost
+        // specs `book` is an identity, so interpolate the transaction itself.
+        if !txn.postings.iter().any(|p| p.cost.is_some()) {
+            let tolerances = crate::transaction_tolerances(txn, &self.tolerance.options());
+            let undo = crate::interpolate::interpolate_in_place_undoable(txn, &tolerances)?;
+            if let Err(e) = self.apply(txn) {
+                undo.restore(txn);
+                return Err(e);
+            }
+            return Ok(Vec::new());
+        }
+        // Slow path: `book` already produces an owned transaction, so work on
+        // that and commit only once apply has succeeded. Nothing needs undoing
+        // — `txn` is not touched until the end.
+        let tolerances = crate::transaction_tolerances(txn, &self.tolerance.options());
+        let booked = self.book(txn)?;
+        let mut candidate = booked.transaction;
+        let _ = crate::interpolate::interpolate_in_place_undoable(&mut candidate, &tolerances)?;
+        self.apply(&candidate)?;
+        *txn = candidate;
+        Ok(booked.gains)
+    }
+
     /// Book and interpolate a transaction, also returning the per-lot
     /// [`CapitalGain`]s `book` computed for it.
     ///
--- a/crates/rustledger-booking/src/interpolate.rs
+++ b/crates/rustledger-booking/src/interpolate.rs
@@ -840,7 +840,10 @@ pub struct Rollback {
     /// prune. A removal also shifts every later index, so the saved indices
     /// below would restore into the wrong postings; the failure would be
     /// silent and would corrupt a ledger.
-    #[cfg(debug_assertions)]
+    ///
+    /// Not `#[cfg(debug_assertions)]`: `debug_assert!` type-checks its
+    /// expression in every profile, so a gated field breaks the release
+    /// build — which `cargo test` does not compile.
     removed: bool,
 }
 
@@ -857,10 +860,7 @@ impl Rollback {
 
     /// Records that a posting was removed. See the `removed` field.
     const fn note_removal(&mut self) {
-        #[cfg(debug_assertions)]
-        {
-            self.removed = true;
-        }
+        self.removed = true;
     }
 
     pub fn restore(self, txn: &mut Transaction) {
--- a/crates/rustledger-booking/src/interpolate.rs
+++ b/crates/rustledger-booking/src/interpolate.rs
@@ -811,12 +811,116 @@ pub fn interpolate_with_tolerances(
 ///
 /// # Errors
 /// Same as [`interpolate`].
+/// Restores the postings interpolation touched, so a failure leaves the
+/// transaction exactly as the author wrote it.
+///
+/// A failed transaction is not discarded — `run_booking` partitions it into
+/// `failed` and `finalize` merges it back into `Ledger.directives`, so a user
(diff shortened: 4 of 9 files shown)


robcohen:
## Correction: the mimalloc dispatch must not be done in Rust

I suggested above that "the clean fix is upstream in the `mimalloc`
crate" and offered to write it. That is wrong, and worth recording
before anyone acts on it.

**The optimization I measured is a known unsoundness advisory.**
[RUSTSEC-2022-0094](https://rustsec.org/advisories/RUSTSEC-2022-0094):
the `mimalloc` crate did exactly this before 0.1.39. Per the advisory,
it "depended on a promise regarding alignments made by the author of the
mimalloc allocator to avoid using aligned allocation functions where
possible for performance reasons"; mimalloc's internals changed, the
promise no longer held, and the crate returned memory with incorrect
alignment. "The flaw was fixed by always using the aligned allocation
functions."

So the unconditional `mi_malloc_aligned` I flagged as an inefficiency is
the **fix** for that advisory. Proposing to remove it upstream would be
proposing to reintroduce the vulnerability. My instinct not to ship it
was right; my stated reason was incomplete.

**The jemalloc binding does dispatch, and the difference is instructive.**
`jemallocator`:

```rust
fn layout_to_flags(align: usize, size: usize) -> c_int {
    if align <= ALIGNOF_MAX_ALIGN_T && align <= size { 0 }
    else { ffi::MALLOCX_ALIGN(align) }
}
```

Two things my experiment got wrong. It tests `align <= size` as well as
`align <= max_align_t` — the condition I had to derive for myself from
`alloc.c:52` and had omitted. And it rests on the **C standard's**
malloc guarantee, not on one allocator author's private assurance.

**That predicate is still unsound on mimalloc**, which is the sharp
finding. mimalloc's own check is stricter:

```c
static bool mi_malloc_is_naturally_aligned(size_t size, size_t alignment) {
  if (alignment > size) return false;
  const size_t bsize = mi_good_size(size);
  return (bsize <= MI_PAGE_MAX_START_BLOCK_ALIGN2 && _mi_is_power_of_two(bsize)) ||
         (alignment == MI_PAGE_OSPAGE_BLOCK_ALIGN2 && (bsize % MI_PAGE_OSPAGE_BLOCK_ALIGN2) == 0);
}
```

It also demands the *bucket* size be a power of two. A 48-byte
allocation at align 16 fails that, so mimalloc does **not** currently
guarantee 16-byte alignment for it — it reaches the guarantee by
over-allocating. Port jemallocator's rule to mimalloc and you get the
2022 bug back.

That same clause explains the profile: an ordinary 48-byte, align-8
allocation is not a power-of-two bucket, so it takes the overalloc path
even though every block is already suitably aligned. 52,710 of 54,221
zeroed-aligned allocations in this run landed there.

**Where the fix belongs.** In mimalloc's C, where the invariant lives —
a short-circuit for `alignment <= MI_MAX_ALIGN_SIZE && alignment <= size`
backed by a *documented* guarantee. That fixes every binding at once and
puts the soundness argument with the people who own it. Worth filing
upstream with these numbers; not worth patching locally in either Rust
or vendored C.

**What is actually ours to fix.** 154,318 allocations for 10,000
transactions is ~15 per transaction. Allocating less is entirely in our
control and needs no assumptions about anyone's allocator. That is the
direction I would take the remaining warm-path work.

Sources: [RUSTSEC-2022-0094](https://rustsec.org/advisories/RUSTSEC-2022-0094),
[jemallocator](https://github.com/tikv/jemallocator),
[microsoft/mimalloc `alloc-aligned.c`](https://github.com/microsoft/mimalloc/blob/main/src/alloc-aligned.c)

robcohen:
## Self-review, correctness focus

Three findings, all fixed in `c5465f1d28` and `cca28a2e1f`. The first is
a real hole in the in-place change.

### 1. `Rollback` did not cover posting REMOVAL

`interpolate_inner` routes all 11 writes through `Rollback`, but it also
**removes** postings — the zero-fill prune at its tail — and that was not
recorded.

It is correct today, and only by position: nothing fallible runs after
the prune, so a removal is never followed by a rollback. I verified that
by checking there is no `return Err` or `?` between the prune and the
`Ok`.

That is an ordering invariant, and it is invisible exactly where someone
would break it — by adding a fallible step after the prune. The failure
would be quiet and bad: `restore` truncates appended postings and writes
saved ones back **by index**, and a removal shifts every later index, so
it would restore the wrong postings and one would simply be gone from a
ledger.

`Rollback` records the removal now and `restore` asserts it never
happened. Verified by simulating precisely that mistake — inserting a
fallible step after the prune — which trips the assertion in
`test_zero_interpolated_posting_pruned_on_opened_account` instead of
silently corrupting the transaction. Debug-only, matching the witness
clone `begin_undo` takes for the inventory undo log.

### 2. The tripwire broke the release build

`debug_assert!` type-checks its expression in every profile, so gating
the field on `#[cfg(debug_assertions)]` left `restore` naming a field
that does not exist without them. `cargo test` builds debug and never
saw it.

Worth recording how it surfaced: a `cargo build --profile profiling` in
the same command printed "build failed" while the output comparison
beside it still reported "identical" — it was diffing a **stale binary
from the previous build**. A green comparison against an artifact that
did not come from the code under test is worth nothing. I have added a
release build to the gate.

### 3. `add_headroom_for`'s early return was undocumented

Folding the two maps replaced a fall-through with an early `return true`
when the currency has no stats entry. The two agree, but only because a
recorded slot always had a stats entry beside it: `add` writes the totals
first, `rebuild_caches` writes both in one pass, and only the slot is ever
cleared. The just-deserialized inventory, where that does not hold, is
refused above. Now said in the code.

## Checked and found sound

- **Nothing but postings is mutated.** The only other `transaction.`
  reference in `interpolate_inner` is a read of `transaction.date`.
- **`units_cache.insert` cannot clobber `simple_slot`.** In `add` the
  totals are written before the slot, and the `insert` path runs only when
  no entry exists, so there is no slot to lose. `rebuild_caches` uses
  `entry().or_default()` for the totals, which never replaces the struct,
  and runs before the slot write in the same iteration.
- **`removing_a_lot_repairs_the_index_of_a_later_cost_less_lot`** — the
  test guarding the removal path, where a stale entry pointed `add`'s
  merge at a tombstone — passes unchanged against the folded field. I
  converted both of its `retain` call sites, so I would have been
  suspicious if it had not.
- **Ordering.** `Residuals::remove` uses `retain`, which preserves the
  order of survivors, and `slot()` appends, so insertion order holds
  across both.
- **`OnceLock` is fine under rayon**: `get_or_init` is thread-safe, and
  the loader reads sibling includes in parallel.
- **`book_and_interpolate` is not dead** — wasm, the fuzz target and
  tests still use the cloning wrapper.

## Gaps, not fixed

- **The new path is not fuzzed.** `fuzz_booking` drives
  `book_and_interpolate`, the cloning wrapper, so `book_interpolate_apply`
  and `Rollback` get no fuzz coverage. Pointing the target at the in-place
  path would exercise the rollback against arbitrary transactions, which
  is where I would want it.
- **`SourceFile::num_lines` is no longer `const fn`** — it reads a
  `OnceLock` now. Breaking for a caller using it in a const context, which
  is unlikely but real.

Issue #2093, opened 2026-08-19
[Compat]: Same-date buy before sell reordering breaks empty-cost-spec lot matching


olivierbellone:
### Description

`sort_directives` orders same-date transactions with `has_cost_reduction` as the third sort key (https://github.com/rustledger/rustledger/blob/v0.21.0/crates/rustledger-core/src/directive.rs#L641), so augmentations (buys) always process before reductions (sells) regardless of file order. This was introduced  in #841.
                                                                         
When a sell uses an empty cost spec `{}` and a buy of the same commodity exists on the same date, the reordering adds lots to the inventory before the sell runs. The sell then hits multiple lots with different costs → `AmbiguousMatch` error. Python beancount processes same-date transactions in file order (`lineno` tiebreaker), so the sell sees only the earlier lot and succeeds.                                                                
                                                                                                                 
The STRICT booking total-match exception (https://github.com/rustledger/rustledger/blob/v0.21.0/crates/rustledger-core/src/inventory/booking.rs#L176: `total_units == units.number.abs()`) doesn't help because the sell amount no longer equals the total inventory after the reordering injected the extra lots.

### Beancount File

```text
option "title" "Compat Test"
option "operating_currency" "USD"

2020-01-01 open Assets:Invest:Stock AAPL
2020-01-01 open Income:Gains USD
2020-01-01 open Assets:Cash USD

2020-01-01 * "Deposit"
  Assets:Cash        5000 USD
  Income:Gains      -5000 USD

2020-06-01 * "Buy 10 AAPL"
  Assets:Invest:Stock  10 AAPL {100 USD}
  Assets:Cash       -1000 USD

; Same-date: sell first (file order), then buy.
; Beancount: sell sees 1 lot → unambiguous.
; Rustledger: buy processed first → sell sees 2 lots → AmbiguousMatch.
2020-12-01 * "Sell 10 AAPL"
  Assets:Invest:Stock -10 AAPL {} @ 120 USD
  Assets:Cash        1200 USD
  Income:Gains       -200 USD

2020-12-01 * "Buy 20 AAPL"
  Assets:Invest:Stock  20 AAPL {130 USD}
  Assets:Cash       -2600 USD

2020-12-02 balance Assets:Invest:Stock  20 AAPL
2020-12-02 balance Assets:Cash         2600 USD
```

### Python Beancount Output

```text
No errors.

Transactions in processing order:
  2020-01-01 Deposit
  2020-06-01 Buy 10 AAPL
[...]

PR #2114, opened 2026-08-23
test(compat): add a differential booking fuzzer

robcohen:
Booking decides **which lot** a reduction consumes. Get it wrong and the postings still balance, the count is still right, and the error list is still empty — the ledger is simply wrong about what you own and what you gained.

Example-based tests are worst at exactly that, and the fix log agrees: **five booking fixes in twenty commits** (#2058, #2069, #2081, #2093, #2099), each an ordering or lot-selection case nobody had written a test for.

## What it does

Generates ledgers and compares **booked lots** against Python beancount — an independent implementation rather than a restatement of our own logic. For every account it compares full lot identity, `(currency, per-unit cost, cost currency, lot date) -> units`, so consuming the wrong lot *at the same total value* is still a failure.

It also compares **acceptance**: a ledger one engine books and the other rejects is a divergence even when no numbers differ. That's the shape of #2099 — the bug there was answering at all.

## It found a real divergence immediately

First campaign: **58/60 agreed**. Both failures were LIFO with same-date lots. Filed separately with a 12-line reproduction — beancount and rustledger disagree about which same-date lot LIFO treats as "last in", giving different cost bases and different realised gains.

## Two things learned building it

**`cost(position)` is the wrong column.** For a costless posting it returns the *units themselves*, so a cash leg looks like it holds a cost of 1. `cost_number` is per-unit and null when there is no cost, which makes it the right discriminator. The first draft compared cash legs against nothing and reported four false divergences per run — caught by the self-test, not by reading.

**AVERAGE and NONE are deliberately not generated.** beancount rejects both outright, so every such run would be a known-unsupported skip rather than a test.

## Why you can believe a green run

`--self-test` plants six known outcomes — wrong lot, wrong units, either side rejecting, and two agreements — and requires each to be classified correctly. Then it requires the real engines to agree on a hand-checked ledger.

A harness that has only ever printed "no divergences" cannot be distinguished from one incapable of printing anything else. This one is checked in both directions before it's believed, which is also how the `cost(position)` mistake surfaced within minutes rather than living as a permanent source of false alarms.

```
scripts/compat-booking-fuzz.py --runs 200
scripts/compat-booking-fuzz.py --seed 35     # reproduce one case
scripts/compat-booking-fuzz.py --self-test
```

### DIFF
--- a/scripts/compat-booking-fuzz.py
+++ b/scripts/compat-booking-fuzz.py
@@ -1,0 +1,292 @@
+#!/usr/bin/env python3
+"""Booking fuzzer — differential checks on generated ledgers against beancount.
+
+Booking decides WHICH lot a reduction consumes. Get it wrong and the postings
+still balance, the count is still right, and the error list is still empty —
+the ledger is simply wrong about what you own and what you gained. That is the
+failure mode example-based tests are worst at catching, and the recent fix log
+shows why it matters: five booking fixes in twenty commits (#2058, #2069,
+#2081, #2093, #2099), each an ordering or lot-selection case nobody had
+written a test for.
+
+So this generates ledgers instead and compares the BOOKED LOTS against Python
+beancount, which is an independent implementation rather than a restatement of
+our own logic. For every account it compares the full lot identity —
+(currency, per-unit cost, cost currency, lot date) -> units — so consuming the
+wrong lot at the same total value is still a failure.
+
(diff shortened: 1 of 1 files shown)


robcohen:
Superseded by the same commit on an allowed branch prefix — `test/` is rejected by the branch-name check.

Issue #2115, opened 2026-08-23
LIFO picks a different same-date lot than beancount


robcohen:
Found by the differential booking fuzzer in #2114 — 58/60 generated ledgers agreed, and both failures were this.

When two lots share an acquisition date, LIFO disagrees with beancount about which one is "last in". The postings balance either way and no error is raised; only the cost basis and the realised gain differ.

## Reproduction

```beancount
option "booking_method" "LIFO"
2020-01-01 open Assets:Stock  HOOL "LIFO"
2020-01-01 open Assets:Cash   USD
2020-01-01 open Income:Gains  USD

2020-01-03 * "buy A (first in file)"
  Assets:Stock  1 HOOL {10.00 USD}
  Assets:Cash  -10.00 USD

2020-01-03 * "buy B (same date, later in file)"
  Assets:Stock  1 HOOL {12.00 USD}
  Assets:Cash  -12.00 USD

2020-02-01 * "sell one, LIFO chooses"
  Assets:Stock  -1 HOOL {}
  Assets:Cash   13.00 USD
  Income:Gains
```

| | consumes | keeps |
|---|---|---|
| **beancount 3.2.3** | A, the **first**-in-file lot | 1 @ 12.00 |
| **rledger** | B, the **last**-in-file lot | 1 @ 10.00 |

Both accept the ledger; neither reports an error.

## Why it matters

On the generated case (seed 35, four lots, two sharing 2020-01-03) the divergence compounds:

| | remaining | cost basis | Income:Gains |
|---|---|---|---|
| beancount | 5 @ 12.00 + 9 @ 12.00 | 168.00 | -92.00 |
| rledger | 5 @ 12.00 + 9 @ 10.00 | 150.00 | -74.00 |

An 18.00 USD difference in both basis and realised gain, silently.

## Which is right is a judgement call

rledger's behaviour is arguably the more literal reading of LIFO: the lot appearing later in the file *is* the one acquired last, and #2093 deliberately made same-date directives book in file order.

beancount appears to sort by date only; with equal dates its stable sort preserves insertion order, so "last in" resolves to the first-inserted lot. That may be incidental to its implementation rather than intended.

So this is reported as a **divergence**, not a defect — I don't think it should be changed without deciding which semantics are wanted. Worth noting the same question applies to HIFO with equal costs and to FIFO with equal dates, which the fuzzer doesn't currently distinguish because it only varies one axis at a time.

Reproduce with:

```
[...]

PR #2116, opened 2026-08-23
test(compat): add a differential booking fuzzer

robcohen:
Supersedes #2114, which used a `test/` branch prefix — rejected by the branch-name check (only `feat|fix|chore|docs|refactor|perf|release|hotfix`). Same commit, allowed prefix.

---

Booking decides **which lot** a reduction consumes. Get it wrong and the postings still balance, the count is still right, and the error list is still empty — the ledger is simply wrong about what you own and what you gained.

Example-based tests are worst at exactly that, and the fix log agrees: **five booking fixes in twenty commits** (#2058, #2069, #2081, #2093, #2099), each an ordering or lot-selection case nobody had written a test for.

## What it does

Generates ledgers and compares **booked lots** against Python beancount — an independent implementation rather than a restatement of our own logic. For every account it compares full lot identity, `(currency, per-unit cost, cost currency, lot date) -> units`, so consuming the wrong lot *at the same total value* is still a failure.

It also compares **acceptance**: a ledger one engine books and the other rejects is a divergence even when no numbers differ. That's the shape of #2099 — the bug there was answering at all.

## It found a real divergence immediately

First campaign: **58/60 agreed**. Both failures were LIFO with same-date lots, now filed as #2115 with a 12-line reproduction — beancount and rustledger disagree about which same-date lot LIFO treats as "last in", giving different cost bases and different realised gains (18.00 USD apart on the generated case).

## Two things learned building it

**`cost(position)` is the wrong column.** For a costless posting it returns the *units themselves*, so a cash leg looks like it holds a cost of 1. `cost_number` is per-unit and null when there is no cost, which makes it the right discriminator. The first draft compared cash legs against nothing and reported four false divergences per run — caught by the self-test, not by reading.

**AVERAGE and NONE are deliberately not generated.** beancount rejects both outright, so every such run would be a known-unsupported skip rather than a test.

## Why you can believe a green run

`--self-test` plants six known outcomes — wrong lot, wrong units, either side rejecting, and two agreements — and requires each to be classified correctly. Then it requires the real engines to agree on a hand-checked ledger.

A harness that has only ever printed "no divergences" cannot be distinguished from one incapable of printing anything else. This one is checked in both directions before it's believed, which is also how the `cost(position)` mistake surfaced within minutes rather than living on as a permanent source of false alarms.

```
scripts/compat-booking-fuzz.py --runs 200
scripts/compat-booking-fuzz.py --seed 35     # reproduce the LIFO divergence
scripts/compat-booking-fuzz.py --self-test
```

### DIFF
--- a/scripts/compat-booking-fuzz.py
+++ b/scripts/compat-booking-fuzz.py
@@ -1,0 +1,292 @@
+#!/usr/bin/env python3
+"""Booking fuzzer — differential checks on generated ledgers against beancount.
+
+Booking decides WHICH lot a reduction consumes. Get it wrong and the postings
+still balance, the count is still right, and the error list is still empty —
+the ledger is simply wrong about what you own and what you gained. That is the
+failure mode example-based tests are worst at catching, and the recent fix log
+shows why it matters: five booking fixes in twenty commits (#2058, #2069,
+#2081, #2093, #2099), each an ordering or lot-selection case nobody had
+written a test for.
+
+So this generates ledgers instead and compares the BOOKED LOTS against Python
+beancount, which is an independent implementation rather than a restatement of
+our own logic. For every account it compares the full lot identity —
+(currency, per-unit cost, cost currency, lot date) -> units — so consuming the
+wrong lot at the same total value is still a failure.
+
+It also compares ACCEPTANCE. A ledger one engine books and the other rejects
+is a divergence even when no numbers differ, and #2099 (report an ambiguous
+STRICT match instead of guessing FIFO) is exactly that shape: the bug was
+answering at all.
+
+Deliberately NOT generated:
+
+  - AVERAGE and NONE booking. beancount rejects both ("AVERAGE method is not
+    supported", "Too many missing numbers"), so every run would be a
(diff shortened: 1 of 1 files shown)


robcohen:
All four addressed in `823379c577`. Two were real defects rather than style, and worth naming:

**The `--python` default was `/home/dev/.compat-venv/bin/python`** — my own machine's path, committed. Now `sys.executable`, which also matches how the sibling compat scripts are invoked (`python3 scripts/compat-values.py`, under an interpreter that already has beancount), with the flag still available for a dedicated venv.

**`json.loads` was unguarded on the rledger side** while the beancount side already caught decode failures. A banner or stray log line on stdout would have aborted the whole campaign instead of counting as one booking failure. The asymmetry is what made it wrong — the comparison treats the two paths as equivalent, so they have to fail the same way. Verified: non-JSON stdout now returns `ERROR` rather than raising.

Also extracted the query out of adjacent string literals, and stopped materializing the seed sequence a second time after iterating it.

Self-test still passes, and seed 35 still reproduces the LIFO divergence (#2115).

PR #2117, opened 2026-08-23
test(compat): vary lot-identity axes together in the booking fuzzer

robcohen:
The first campaign answered one question well and never asked the others. Every divergence it found had same-date lots — **12/12, against 0/189** for seeds without them — because random generation collided on date, cost and label only by accident. Two lots at the same *highest* cost under HIFO, or two on the same date under STRICT, essentially never appeared.

Lot identity has three components a method can tie on, so the generator now constructs each configuration deliberately (`none`, `date`, `cost`, `both`, `label`) and reports the tie mode with every divergence.

## Coverage is now checkable

```
             none   date   cost   both  label
  STRICT       26     28     30     17     28
  FIFO         23     27     18     31     26
  LIFO         22     24     26     29     24
  HIFO         19     20     23     31     28
```

All 20 cells populated at `--runs 500`. That's what makes the *empty* results meaningful: **HIFO with tied costs and STRICT with tied dates now have real coverage and agree with beancount.** Previously "no HIFO divergences" just meant the case had never been generated.

## Two corrections, both from measuring rather than reading

**`none` was not a control.** Costs were drawn *with replacement* from a four-element pool, so **79%** of "no tie" runs quietly contained a cost tie — useless for attributing a divergence to the date axis. Now sampled without replacement: 0 date ties, 0 cost ties.

**Tying exactly two lots lost a known bug.** The FIFO coalescing divergence needs three lots sharing a date with two sharing a cost — a *non-contiguous* repeat — which a uniform draw produced about once in seven date-tied runs. I confirmed this by running the known-divergent ledger directly: the bug was still there, the generator had simply stopped reaching it. Ties now span a random-sized group, and date-tied groups draw costs from a narrow pool so repeats land non-contiguously. FIFO divergences reappeared immediately.

## What it finds now

At `--runs 500`: **490/500 agreed**, 10 divergences — 8 LIFO × date, 1 FIFO × date, and one new class:

**LIFO × label** — two lots differing only by label, where **rledger rejects a ledger beancount books**. That's the mirror of the acceptance divergence already in #2115, so the disagreement cuts both ways: depending on the ledger, either engine may be the one that refuses.

All are the same root cause and tracked in #2115.

### DIFF
--- a/scripts/compat-booking-fuzz.py
+++ b/scripts/compat-booking-fuzz.py
@@ -28,6 +28,15 @@ Deliberately NOT generated:
   - Prices on reductions, which interact with cost inference in ways that are
     a separate surface from lot selection.
 
+Lot identity has three components a method can tie on — acquisition date,
+per-unit cost, and label — and the first campaign only ever collided on them
+by accident. Every divergence it found had same-date lots (12/12, against
+0/189 for seeds without them), which meant the generator was answering one
+question well and never asking the others: two lots at the same HIGHEST cost
+under HIFO, or two lots on the same date under STRICT, essentially never
+appeared. `TIES` now constructs each configuration deliberately, and the tie
+mode is reported with every divergence so the classes stay separable.
+
 Usage:
     scripts/compat-booking-fuzz.py --runs 200
     scripts/compat-booking-fuzz.py --seed 12345        # reproduce one case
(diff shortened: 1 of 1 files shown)


robcohen:
Both correct, both fixed in `9de3960635` — and together they were hiding two thirds of the divergences in this space.

**The comparison key didn't include the label.** Exactly as you say: the generator gained a label axis but the key didn't, so two lots differing only by label collapsed into one entry *on both sides*. A label divergence was visible only when it changed acceptance; a divergence in *which* labelled lot got consumed was invisible by construction.

Adding `cost_label` took the campaign from **490/500 to 465/500** — 15 LIFO × label divergences where there had been 1. Seed 30 is representative:

```
Assets:Stock/HOOL/12/USD/2020-01-02/lot-a: rledger=5 beancount=3
Assets:Stock/HOOL/12/USD/2020-01-02/lot-b: rledger=1 beancount=3
```

Same total, same cost, same date — different distribution across labels.

**The comment claimed a shape the code didn't produce.** Also right: an independent draw per lot meant `group==2` never repeated and `group>=3` mostly came out contiguous or all-equal. The shape is now constructed outright (`[a, b, a, …]`) rather than hoped for — **47 of 47** date-tied groups of three or more have it, against roughly one in seven before. FIFO divergences went from 1 to 7.

I'd been treating that comment as documentation when it was really an assertion I hadn't checked, which is the more useful half of the catch.

Issue #2118, opened 2026-08-24
[Compat]: rustledger keeps a slot per acquisition where beancount pools identical lots — FIFO/LIFO consume in different orders


robcohen:
## Summary

beancount's `Inventory` is keyed by `(currency, cost)`, so two acquisitions with identical `(cost, date, label)` become **one pooled position**. rustledger keeps **one slot per acquisition**. Both engines then walk their own positions in order, so once a ledger buys the same lot identity more than once on the same date, FIFO and LIFO consume in different orders and report a different cost basis and different realized gains.

Found by the differential booking fuzzer (`scripts/compat-booking-fuzz.py`, #2116/#2117). After the LIFO tie-break fix for #2115, this is the **only** remaining divergence class in a 200-seed run: 5 of 5 residual failures share this root cause (3 FIFO, 2 LIFO).

## Reproducer

Three buys, all on one date, where the first and third share a lot identity:

```beancount
option "booking_method" "FIFO"
2020-01-01 open Assets:Stock  ACME "FIFO"
2020-01-01 open Assets:Cash   USD
2020-01-01 open Income:Gains  USD

2020-01-02 * "buy A (cost 10)"
  Assets:Stock  10 ACME {10.00 USD}
  Assets:Cash  -100.00 USD

2020-01-02 * "buy B (cost 20)"
  Assets:Stock  10 ACME {20.00 USD}
  Assets:Cash  -200.00 USD

2020-01-02 * "buy C (cost 10, same lot identity as A)"
  Assets:Stock  10 ACME {10.00 USD}
  Assets:Cash  -100.00 USD

2020-03-01 * "sell 15"
  Assets:Stock  -15 ACME {}
  Assets:Cash   300.00 USD
  Income:Gains
```

Remaining holdings after the sale:

| engine | @10.00 | @20.00 | remaining basis |
|---|---|---|---|
| rledger | 10 | 5 | 200.00 |
| beancount | 5 | 10 | 250.00 |

Note this is **FIFO**, and every buy is on the same date, so no tie-break *direction* is involved — both engines walk forward. The only difference is what they are walking.

## Why they differ

- **rledger** holds three slots — `[10@10 (A), 10@20 (B), 10@10 (C)]` — and consumes A(10) then B(5).
- **beancount** pools A and C into `20@10.00`, and FIFO drains that position first: it consumes A(10) then **C(5)**, never touching B.

## Which is right

[...]

Issue #2091, opened 2026-08-19
HIFO and STRICT_WITH_SIZE still scan every lot per reduction (HIFO: 18s at 20k, 101x for 10x input)


robcohen:
## Summary

#2083 fixed ordered lot selection for FIFO and LIFO. Two booking methods carry their own scans and were untouched: both are still quadratic, and HIFO badly so.

Measured on `main` @ 105cc6ca5b, one FIFO-shaped ledger regenerated per method (`gen-profile-workload.py fifo N`, with the account's booking method swapped), cold cache:

| method | 2,000 | 20,000 | growth for 10x input |
|---|---|---|---|
| FIFO | 0.03s | 0.27s | 9x — linear |
| LIFO | 0.02s | 0.27s | 14x |
| AVERAGE | 0.02s | 0.32s | 16x |
| **STRICT_WITH_SIZE** | 0.06s | **3.34s** | **56x** |
| **HIFO** | 0.18s | **18.23s** | **101x** |

HIFO takes **18 seconds** where FIFO takes 0.27s on the same ledger — 68x slower for a choice of booking method that should not change the shape of the work.

## Cause

`reduce_hifo` and `reduce_strict_with_size` do their own lot selection rather than going through `plan_ordered`, so #2083's ordered index never reaches them. Each walks every slot, applies the cost-spec predicate to each, and sorts the survivors — once per reduction. With lots accumulating as the ledger grows, that is O(lots) per reduction across O(N) reductions.

`reduce_hifo`:

```rust
let mut matching: Vec<(usize, Decimal)> = self
    .positions
    .iter_slots()
    .filter(|(_, p)| {
        p.units.currency == units.currency
            && !p.is_empty()
            && p.units.number.signum() != units.number.signum()
            && p.matches_cost_spec(spec)
    })
    .map(|(i, p)| (i, p.cost.as_ref().map_or(Decimal::ZERO, |c| c.number)))
    .collect();
```

Same shape as the code #2083 replaced, one function over.

## Fix

The same treatment, with a different key. #2083 added `ordered_index`, holding each currency's lots in `(date, slot)` order so FIFO is a prefix walk and LIFO a suffix walk, with the walk stopping as soon as the reduction is covered. HIFO wants `(cost, slot)` order; STRICT_WITH_SIZE needs looking at on its own terms, since its selection is size-aware rather than purely ordered.

[...]

PR #2119, opened 2026-08-24
fix(core): break LIFO ties by insertion order like FIFO and HIFO

### LINKED ISSUE #2115: LIFO picks a different same-date lot than beancount

robcohen:
Closes #2115.

## The bug

LIFO reduced through `plan_ordered(.., reverse: true)`, which walked the candidate list with `.iter().rev()`. That list is sorted `(date ASC, slot ASC)`, so reversing it reversed the **slot tiebreak** along with the sort key: LIFO took the *last* of two same-date lots while FIFO and HIFO take the first.

beancount reverses only the primary key — `sorted(key=.., reverse=True)` is stable, so ties keep insertion order in either direction. The mismatch was structural, in one `bool`.

## Why this is an incomplete migration, not a judgement call about LIFO

`order_key` already documented the rule this broke. HIFO negates its cost key:

> `// Negated rather than reversed so the tuple still sorts ascending`
> `// and the slot tiebreak keeps its direction.`

#2091 moved HIFO onto that rule and left LIFO on the older reverse-the-walk one. So FIFO and HIFO agreed with beancount on ties while LIFO did not, and the split was incidental rather than intended.

## The fix

`LotOrder::DateDescending` reverses the **key** (via `Reverse`, since `jiff::civil::Date` cannot be negated the way a `Decimal` cost can), and the `reverse` flag is **deleted** — `plan_ordered`/`reduce_ordered` take a `LotOrder` and the walk is unconditionally forward. A reversed tiebreak is no longer expressible rather than merely corrected.

`Reverse(None)` sorts *last*, which is exactly where the reversed walk left date-less lots, so only the tie moves.

Also splits `BookingMethod::Fifo | Lifo => LotOrder::Date` into separate arms in **both** places that map method to order — `reduce`, and the parallel copy in the `try_reduce` preview path. Those must move together or preview and commit disagree.

## Nothing caught this, so two tests changed

`test_hifo_with_tie_breaking` is named for this behavior and asserted only units and total cost basis. Every lot in its fixture shares one cost, so those totals are identical under any tiebreak — it passed with the tiebreak reversed at all three sort sites. Verified by mutation, not by reading: with the tiebreak flipped, all 485 `rustledger-core` tests stayed green while the BQL output demonstrably changed (`2020-01-03` → `2020-01-04`). It now asserts which lots were consumed.

`ordered_methods_break_ties_on_insertion_order` (new) pins the rule for all three methods, each tied on the key its *own* method sorts by so the primary key cannot decide the outcome. Grafted onto the pre-fix tree it fails with:

```
assertion `left == right` failed: Lifo must consume the FIRST of two same-date lots, not the last
  left: 12
 right: 10
```

and the FIFO case passes before it — which is the asymmetry itself.

`the_ordered_index_selects_what_the_scan_selects` now iterates three `LotOrder`s instead of `(order, reverse)` pairs: strictly better coverage, since `DateDescending` is a real ordering rather than a flag.

## Verification

- 486 `rustledger-core` lib tests pass; `cargo +1.97.0 clippy --workspace --all-targets` clean.
- Differential fuzz against beancount over 200 generated ledgers (`scripts/compat-booking-fuzz.py`): **16 → 5** divergences. LIFO/label 8 → **0**, LIFO/date 5 → 2, FIFO/date 3 → 3.
  <br>(An earlier revision of this description said 15 → 5 with LIFO/date 4 → 2. That came from a log file truncated by a disk-full kill mid-run; the re-measured pre-fix figure is 16 divergences, 184/200.)
- The 5 that remain are one *unrelated* root cause, now filed as #2118: rustledger keeps a slot per acquisition while beancount pools lots with identical `(cost, date, label)`. Confirmed on all five by hand. Not addressed here.
- `corpus_parity`'s 28 divergences are pre-existing — identical count on the stashed pre-fix tree, and they are decimal formatting (`-60` vs `-60.00`) and parse-error-vs-never-opened, which no booking change can produce.

## Reviewer note: this is user-visible

Cost basis, realized gains, and therefore `meta.hash` move for any LIFO ledger holding same-date lots. `Cost` has **day resolution**, so among same-date lots there is no fact about which was "last in" — this makes LIFO consistent with FIFO, HIFO, and beancount, but it is not recovering a uniquely correct answer. Probably worth a release note.


robcohen:
Good catch, and correct — fixed in 36c2fb1b88.

You identified a claim I had asserted in a comment and in the commit message and pinned with nothing. `Reverse<Option<NaiveDate>>` puts date-less lots last; `Option<Reverse<NaiveDate>>` puts them first, and your reading of why nothing separated the two is exactly right — `the_ordered_index_selects_what_the_scan_selects` calls `order_key` on both sides so it moves with any change to the key, and every other LIFO test uses dated lots only.

Verified rather than assumed: making precisely that swap fails `lifo_takes_the_date_less_lot_last` and **only** that test, out of 487.

The fixture puts the date-less lot **first in the file**, so slot order alone cannot produce the expected answer — otherwise the test would pass for the wrong reason.

PR #2120, opened 2026-08-24
test(compat): classify the lot-pooling divergence and gate CI on the booking fuzzer (#2118)

robcohen:
Follow-up to #2118, which we decided **not** to fix: beancount pools acquisitions sharing `(cost, date, label)` into one inventory position while rustledger keeps a slot per acquisition, so the two consume in different orders. rustledger's answer respects acquisition order and beancount's cannot, so we keep ours.

That leaves the fuzzer reporting 5 permanent failures per 200 seeds, and a permanently red run is one nobody reads. This teaches it the difference between "known and accepted" and "new".

## The classifier is deliberately narrow

`pooling_shape` matches only a **non-contiguous repeat of one lot identity within a single date**:

| lots on one date | matches? | why |
|---|---|---|
| `[a, b, a]` | yes | pooling moves `a`'s later units ahead of `b` |
| `[a, a, b]` | no | pools into the order it already had |
| `[a, b, a]` on 3 dates | no | different dates cannot pool |
| `[a("x"), b, a("y")]` | no | labels make the repeat a different identity |

Every case it matches is a case this harness stops guarding, so "some identity repeats" would have been too broad.

## What does not change

Divergences are still compared and still **printed in full** — only the verdict changes. Nothing is skipped.

An **acceptance** divergence (one engine rejecting a ledger the other books) is never waived whatever shape the lots take: pooling redistributes units between lots, it cannot change accept/reject.

All three counts print unconditionally, including a zero. A count that appears only when non-zero reads as "nothing was waived" when the line is simply absent — and waived cases are exactly the ones worth keeping in view.

The exit code now tracks **unexplained** divergences alone.

## Evidence it can still say no

A waiver is only safe if you can show what it refuses to waive.

1. `--self-test` gained five `pooling_shape` cases, four of them near-misses it must reject. Mutating the function to `return True` fails four of the five.
2. **Against a pre-fix `rledger`** (before #2119 landed the LIFO tie-break fix) the sweep is `exit=1` with **11 unexplained** — all 8 LIFO/label and 3 of the LIFO/date divergences, which is the whole of #2115 — while waiving only the 5 pooling cases. The classifier does not swallow the bug it would be most tempting to swallow.
3. Against a fixed `rledger`: `195/200 agreed, 5 expected, 0 unexplained`, `exit=0`.


robcohen:
Self-review notes.

**This PR is what makes CI integration possible, and that follow-up is worth taking.**

`compat-booking-fuzz.py` is not referenced by any workflow. Its sibling `compat-budget-fuzz.py` is (`compat.yml:835`), as is `compat-bql-test.py` and `compat-values.py` — so a booking fuzzer of the same kind is the odd one out.

It could not have been wired in before now: with five permanent divergences it would have been red on every run. Separating "expected" from "unexplained" and moving the exit code onto the latter is exactly the precondition. `compat.yml` already has the shape to copy — `--self-test` first, then the sweep, failure captured into a step output, and a separate `Gate on …` step that turns it into an error with a reproduce command, plus a larger `--runs` on the nightly schedule.

I have not done it here to keep this PR to one change. Filing it as a follow-up unless a reviewer would rather it landed together.

**Two things verified rather than assumed while reviewing:**

- Both engines encode cost-less positions identically — `""` in position 2 of a six-element key — which is what makes `same_totals`' lot-bearing filter compare like with like instead of silently grouping differently on each side.
- `pooling_shape` reads the generator's final `days`/`costs`/`labels`; every mutation happens before the emission loop, so the classifier cannot drift from the ledger it describes.

**Known limitation, stated deliberately.** The shape is computed from the generator, not from the divergence, so a false NEGATIVE is possible: a pooling divergence the matcher misses is reported as a real failure. That is the safe direction. The dangerous direction — waiving a genuine bug — is what the acceptance guard, the units-total guard, and the pre-fix-binary check cover.

robcohen:
Wired it in after all — 239d0364db. The PR is now classify + gate rather than classify alone, which reads better as one change: the classification exists *so that* CI can run this, and separating them would have left the enabling commit with no visible effect.

`compat.yml` gains a `Booking fuzz` step and a `Gate on booking fuzz`, following the budget fuzz exactly — `--self-test` first (a harness that cannot detect a bug should fail loudly, not pass vacuously), then the sweep, `continue-on-error` with the verdict in a step output, and a separate gate carrying a reproduce command. Same placement, for the documented reason: the orphan branch checked out later has no `scripts/` and no `target/`.

`--start-seed` is new and worth a look. Seeds were always `range(runs)`, so every invocation re-covered identical ground and a nightly run explored nothing a PR run had not. CI now fixes the start for PR/push so a red X does not move when an unrelated commit lands, and uses the run id nightly — the same policy as the budget fuzzer and the synthetic corpus. Verified it is actually honored rather than silently ignored: `--runs 1 --start-seed 109` reports the known #2118 case, `--start-seed 0` reports none.

250 seeds on PR, 2000 nightly. A case is ~0.1s, so the PR cost is ~25s.

actionlint reports the same three findings on this branch as on `main` (lines 124, 737, 1163), none in the new steps — checked against `git show main:` rather than assumed.

robcohen:
**"Run compatibility tests" is red here, and it is the new gate working correctly — not a defect in it.**

This branch is cut from `main`, which does not yet have #2119. So CI built a pre-fix `rledger` and the fuzzer found the LIFO tie-break bug that #2119 fixes. The sample in the log is `lot-a` / `lot-b` on one date — a label tie, the exact #2115 signature.

Reproduced locally against a pre-fix binary over the same seed window, and the numbers match CI exactly:

```
230/250 agreed
6 expected divergence(s) (#2118 lot pooling)
14 unexplained divergence(s)
```

Against a binary with #2119 applied, the same window is `0 unexplained`, exit 0.

So the gate is doing precisely what it was written to do: waive the six known pooling cases, and fail on fourteen unexplained ones that are a genuine bug on this branch.

**Consequence: this PR has to merge after #2119.** I branched it off `main` on the reasoning that the classifier is independent of the fix — true of the code, false of the gate, which necessarily fails wherever the bug still exists. I will update this branch from `main` once #2119 lands and it should go green without further changes.

robcohen:
**Found a defect that no individual PR's CI could show: merged together, this gate would have been RED on `main`.**

I merged all six open PRs locally and ran the gate the way CI runs it on a PR — 250 seeds from 0. Result: one unexplained divergence, **seed 211**. It is #2118 surfacing as an acceptance difference, and this classifier was refusing to waive it.

That refusal rested on a claim I had already found to be false and only half-corrected. Pooling changes which lots *survive* a reduction, so a later reduction naming an explicit cost finds its lot drained on one engine and present on the other. Seeds 117 and 394 show both directions; **211 lands inside the PR window**. Since #2118 provably produces these, refusing to waive them means the gate can never be green — which is precisely what this PR exists to prevent.

Fixed in fef1b8baa1. Acceptance divergences with the pooling shape are now waived, but they are the *more serious* face of #2118 — a ledger that fails to LOAD rather than one that reports different figures — so they get their own line instead of disappearing into the total:

```
7 expected divergence(s) (#2118 lot pooling)
  of which 1 are ACCEPTANCE differences
0 unexplained divergence(s)
```

Waived, never silent: they still print in full like any other divergence.

**Both properties verified against real binaries, not argued:**

- with #2119 applied — `exit 0`, 0 unexplained: the gate is green post-merge
- without it — `exit 1`, 12 unexplained: the gate still catches #2115, so the wider waiver did not swallow the bug it was written to find

**This reverses a position I argued for in this PR**, and the reversal is worth naming: I said an acceptance divergence was "too severe to wave through on a heuristic". The severity is real, and it is why these get their own line and why #2118 was escalated and #2121 now warns about it. But severity is an argument for documenting and for revisiting #2118 — not for a permanently red gate that everyone learns to ignore, which masks far more than it protects.

Also confirmed while checking: all six PRs merge into `main` in sequence with **no conflicts**, including the `CHANGELOG.md` overlap (#2119 × #2123) and the `booking.rs` overlap (#2119 × #2124), and the merged tree passes 490 core tests plus the journal test.

PR #2121, opened 2026-08-24
docs(migration): document the same-day repeated-lot divergence (#2118)

robcohen:
Closes the documentation half of #2118.

Python beancount's `Inventory` is keyed by `(currency, cost)`, so acquisitions sharing a cost, date and label pool into a single position. rustledger keeps one lot per acquisition. When a lot at a different cost was acquired *between* two that pool, the engines consume in different orders — and a holding bought more than once in a day reports a different cost basis and different realized gains.

We decided to keep rustledger's behavior: it respects acquisition order and Python's cannot, because `Cost` carries a date and no time, so the two purchases are genuinely one lot under its model. That decision makes documenting it the deliverable — anyone reconciling figures against Python needs to know why they differ, and nothing warned them.

Goes in `Known Differences` in the migration guide, which is where someone comparing the two engines will actually be looking.

**Both the example and the workaround were run, not reasoned about.**

The worked example reproduces exactly as written: rustledger leaves `10 @10.00` + `5 @20.00`, Python leaves `5 @10.00` + `10 @20.00` — Python consuming the third lot while the second, acquired earlier, goes untouched.

The suggested fix is advice users will follow, so I checked it rather than assuming: giving the repeated purchases distinct labels makes them separate lots on both sides and the two engines then agree (`10 @10.00` + `5 @20.00` each).

Independent of #2119 and #2120 — this documents behavior that is not changing.

### DIFF
--- a/docs/migration/from-beancount.md
+++ b/docs/migration/from-beancount.md
@@ -64,6 +64,12 @@ See [Plugins Reference](../reference/plugins.md) for full list.
    them, the two engines then consume in different orders.
 
    ```beancount
+   option "booking_method" "FIFO"
+
+   2020-01-01 open Assets:Stock  ACME
+   2020-01-01 open Assets:Cash   USD
+   2020-01-01 open Income:Gains  USD
+
    2020-01-02 * "buy A"
      Assets:Stock  10 ACME {10.00 USD}
      Assets:Cash  -100.00 USD
--- a/docs/migration/from-beancount.md
+++ b/docs/migration/from-beancount.md
@@ -98,8 +98,23 @@ See [Plugins Reference](../reference/plugins.md) for full list.
    a date, not a time, so A and C are genuinely the same lot under its model
    and pooling is the only thing its inventory can represent.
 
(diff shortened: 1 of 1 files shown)


robcohen:
Correct on both counts — fixed in 35cfc11cbe.

Reproduced before fixing: the block as written fails `rledger check` with `E1001 Account Income:Gains was never opened`, so it never reaches the booking behavior it exists to demonstrate. And you are right that the default is STRICT, so even with the accounts opened it would not have shown the FIFO case the text describes.

The cause is worth naming: I verified this example against a ledger that *did* carry the setup, then trimmed it for the doc. That is exactly how a verified example becomes an unverified one.

The block now carries `option "booking_method" "FIFO"` and the three `open` directives, and both engines were re-run against it **as it now appears in the page**, extracted from the markdown rather than retyped:

- `rledger check` — clean
- rustledger — `10 @10.00`, `5 @20.00`
- beancount — `5 @10.00`, `10 @20.00`

which are the figures the surrounding prose quotes.

PR #2123, opened 2026-08-24
fix(report): keep journal order when --limit is given

### LINKED ISSUE #2122: [Bug]: `report journal --limit N` reverses the output order, including same-date ties
## Summary

`report journal --limit N` prints its rows in **reverse chronological order**, while the same command without `--limit` prints them chronologically. Adding a limit silently flips the sort direction, and it reverses **same-date ties** along with it.

Found by variant analysis from #2115: that bug was `.iter().rev()` over a list sorted `(date, slot)`, which reversed the slot tiebreak along with the sort key. This is the same shape in the reporting layer.

## Reproducer

```beancount
2020-01-01 open Expenses:Food USD

2020-01-02 * "first"
  Expenses:Food   1.00 USD
  Assets:Cash

2020-01-03 * "second"
  Expenses:Food   2.00 USD
  Assets:Cash

2020-01-04 * "third"
  Expenses:Food   3.00 USD
  Assets:Cash

2020-01-05 * "fourth"
  Expenses:Food   4.00 USD
  Assets:Cash
```

```
$ rledger report ledger.beancount journal
first
second
third
fourth

$ rledger report ledger.beancount journal --limit 2
fourth
third        <-- reversed
```

Same-date transactions reverse too. With `same-day A` written before `same-day B` on one date, the unlimited journal prints A then B; `--limit 2` prints **B then A**.

## Cause

`crates/rustledger/src/cmd/report_cmd/journal.rs:33`:

```rust
entries.sort_by_key(|t| t.date);

let entries_to_show = if let Some(n) = limit {
    entries.into_iter().rev().take(n).collect::<Vec<_>>()
} else {
    entries
};
```

`entries_to_show` is written out directly, so the `.rev()` used to reach the *last* N leaks into the output order. `sort_by_key` is stable, so the same `.rev()` also undoes the file order that same-date transactions were left in.

## What is actually wrong

Taking the last N is a reasonable reading of `--limit` — "the most recent N" is usually what someone wants, and it is what `tail` does. **The selection is fine; the presentation is not.** `tail` does not print lines backwards.

Suggested fix: reverse back after taking, or select with `entries.len().saturating_sub(n)..` and skip the reversal entirely.

```rust
let entries_to_show = match limit {
    Some(n) => entries.split_off(entries.len().saturating_sub(n)),
    None => entries,
};
```

That keeps the last N, keeps them chronological, and keeps same-date entries in file order — which #2093 established as the order booking uses, so the journal agreeing with it matters beyond cosmetics.

## Scope

Text output is shown above; `--format csv` and `--format json` iterate the same `entries_to_show`, so all three are affected. Worth a test asserting limited and unlimited output agree on the order of the rows they share — the property that would have caught this.

robcohen:
Closes #2122.

`--limit N` reached the most recent N with `.rev().take(n)` and wrote the result straight out, so the rows were reversed as well as selected:

```
$ rledger report ledger.beancount journal
first, second, third, fourth

$ rledger report ledger.beancount journal --limit 2
fourth, third        # reversed
```

Taking the last N is a fair reading of `--limit` — "the most recent N" is usually what you want. Printing them backwards is not, and `tail` does not either. **The selection was right; the presentation was wrong.**

`sort_by_key` is stable, so the same reversal also undid the file order of same-date transactions. That is the order booking uses (#2093), so a limited journal disagreed with the ledger about which of two same-day entries came first — which is why this is more than cosmetic.

`split_off(len - n)` takes the tail without reversing it. All three output formats share the one selection point, so all three are fixed together; I checked text, CSV and JSON.

## The test asserts a property, not a transcript

It requires the limited output to be the **tail of the unlimited output, unchanged**, for every limit from 1 to the entry count. A test pinning literal expected output would pass just as happily on a reversed list with the expectation written backwards — that is roughly how this survived.

Verified it fails against the old code, on the tie rather than the obvious case:

```
assertion `left == right` failed: --limit 2 must be the tail of the unlimited report, in the same order
  left: ["same-day B", "same-day A"]
 right: ["same-day A", "same-day B"]
```

## Provenance

Found by variant analysis from #2115 — the same shape one layer up: `.rev()` over a stably sorted list, reversing the tiebreak along with the sort key. I swept the other `.rev()` sites at the same time; `report_cmd/mod.rs` (string truncation, rev → take → rev back), the LSP edit appliers and the undo-log replay are all the correct back-to-front idiom.

Independent of #2119, #2120 and #2121 — branched from `main` and touches only the report path.


robcohen:
Self-review: checked the boundaries, since `split_off` panics if the index exceeds the length and the old `.take(n)` never could.

| `--limit` | result |
|---|---|
| `0` | empty, no panic (old behavior too) |
| `1` | `fourth` — the last entry |
| `4` (= count) | all four, in order |
| `99` (> count) | all four, in order — `saturating_sub` clamps to 0 |

Only `99` changes meaning versus the old code, and that is the fix: it previously returned all four **reversed**.

Issue #2097, opened 2026-08-19
[Compat]: STRICT accepts an ambiguous lot match that beancount rejects, and the comment justifying it is wrong about both beancount and observability


robcohen:
## Summary

Under STRICT, when a reduction's cost spec matches several lots that share a cost number and cost currency, `plan_strict` treats them as interchangeable and falls back to FIFO instead of reporting an ambiguous match. Beancount 3.2.3 reports an error.

Minimal case:

```beancount
2022-05-10 * "buy A"
  Assets:Broker:GLOB   4 GLOB {74.09 USD}
  Equity:Open       -296.36 USD

2024-02-09 * "buy B"
  Assets:Broker:GLOB  16 GLOB {74.09 USD}
  Equity:Open      -1185.44 USD

2024-02-10 * "Sell GLOB"
  Assets:Broker:GLOB     -16 GLOB {74.09 USD} @ 67.70 USD
  Assets:Broker:Cash  1083.20 USD
  Income:Gains
```

```
beancount: Ambiguous matches for "-16 GLOB {74.09 USD}":
             4 GLOB {74.09 USD, 2022-05-10}, 16 GLOB {74.09 USD, 2024-02-09}
rledger:   ✓ No errors found
```

## The rationale in the code is wrong on both counts

`crates/rustledger-core/src/inventory/booking.rs`, in `plan_strict`'s `n =>` arm:

> Two lots count as identical if they have the same cost number + cost currency — the user-visible monetary identity. Date and label differences don't make a reduction ambiguous because **the user could not have observed a different outcome** based on the cost spec they wrote. **Beancount falls back to FIFO in that case.**

1. **Beancount does not fall back to FIFO** — it raises `Ambiguous matches`, as above. Verified on 3.2.3.
2. **The outcome is observable.** The lots differ in acquisition date. FIFO leaves 4 GLOB dated 2024-02-09; taking the sale entirely from the newer lot would leave 4 GLOB dated 2022-05-10. Same cost basis either way, but a different holding period — which this codebase itself acts on, in the calendar short/long-term split of `report capgains` (#1860) and in the per-lot IRR eligibility predicate (#1861). So the choice reaches user-facing tax output.

Cost basis being equal is what makes this quiet: nothing in a balance or a basis total will ever disagree.

## Why it surfaced

[...]

PR #2124, opened 2026-08-24
test(core): pin STRICT_WITH_SIZE's two tie rules

robcohen:
Variant sweep from #2115 and #2122, looking for the same shape: an ordering whose secondary key nothing checks.

## `plan_strict_with_size` is correct — no bug here

It uses `min_by_key` on `(date_key, slot)` ascending, never reverses a walk, and *encodes* `None` dates to sort last rather than relying on `Option`'s natural order:

```rust
.map_or((1, NaiveDate::MAX), |d| (0, d)),
i,
```

That is the same house rule `order_key` follows — put the direction in the key, keep the tiebreak ascending. Reporting that as the result: the sibling defect I went looking for is not there.

## Both tie rules were unpinned

The comment above that key states each as a promise, and nothing held it:

- reversing the slot tiebreak to `Reverse(i)` left **all 485 core tests passing**, plus `rustledger`, `rustledger-validate` and `rustledger-query`
- flipping the `None` encoding so date-less lots sort **first** did the same

`strict_with_size_takes_the_oldest_of_several_exact_size_lots` covers the primary date order from #2097 and stops there. Two size matches that *share* a date, and a date-less size match, both fell through.

Each new test kills exactly its mutant and nothing else fails. Both fixtures put the lot that must **not** win first, so insertion order alone cannot produce the expected answer — otherwise they would pass for the wrong reason.

## Why it matters beyond determinism-for-its-own-sake

`report capgains` splits short from long on the surviving lot's acquisition date, and per-lot IRR keys eligibility off it. A reduction choosing arbitrarily between two same-date lots moves tax figures, silently — the postings still balance and no error is reported.

Tests only; no behavior change. Independent of #2119, #2120, #2121 and #2123.

### DIFF
--- a/crates/rustledger-core/src/inventory/booking.rs
+++ b/crates/rustledger-core/src/inventory/booking.rs
@@ -1826,6 +1826,58 @@ mod reduction_tests {
         );
     }
 
+    /// Two size matches sharing a date resolve by insertion order.
+    ///
+    /// The date comparison cannot separate these, so only the `i` in
+    /// `min_by_key`'s key decides — and the comment above it promises a
+    /// deterministic result. Nothing pinned that: reversing the slot tiebreak
+    /// to `Reverse(i)` leaves all 485 core tests passing.
+    ///
+    /// Determinism here is not decoration. `report capgains` splits short from
+    /// long on the surviving lot's acquisition date, and per-lot IRR keys
+    /// eligibility off it, so a reduction that picks arbitrarily between two
+    /// same-date lots moves tax figures between runs.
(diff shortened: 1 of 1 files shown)


robcohen:
Good point on the name — taken in cdae37f673. `a_date_less_one` reads as "date less one" and, more practically, does not match a search for "undated". Renamed to `strict_with_size_prefers_a_dated_lot_over_an_undated_lot` and aligned the surrounding prose so the doc comment and the assertion message use the same word.

PR #2125, opened 2026-08-24
test(compat): fuzz prices on reductions, and correct a wrong premise

robcohen:
Stacked on #2120 — base it there, not `main`. Merge after it.

The generator deliberately excluded prices on reductions as "a separate surface from lot selection". This adds them: a quarter of reductions take `@ <price>`, a tenth `@@ <total>`, and the price is **not** pinned to the rate the cash leg implies.

Checked before generating anything: a posting carrying a cost takes its weight from the COST, so a price that disagrees with the proceeds is still a balanced transaction on both engines. Had that not held, every generated ledger would have been unbalanced and the harness would have rejected them for a reason that has nothing to do with booking.

## The negative result, stated plainly

**Over 500 seeds the price axis found no price-specific divergence.** Prices reach the booked postings intact on both engines; every divergence in the run traces to #2118. Recording that so the next person does not re-run the same experiment expecting a different answer.

## What it did surface, and a correction

Perturbing the seed stream exposed three **acceptance** divergences, all caused by pooling, and they show #2118 is worse than that issue recorded — it can make a ledger **fail to load**, in both directions:

- **seed 117** — rledger rejects, beancount books. rledger takes `17@8.75 + 13@10.00`, leaving `6 @10.00`, so a later `-14 {10.00}` fails `Not enough units`. beancount pools the 8.75 lots, leaves `17 @10.00`, books it.
- **seed 394** — the reverse. Pooled `11.00` is fully consumed, so a later `-3 {11.00}` finds no lot; rledger keeps `7 @11.00`.

That **disproves a premise I wrote in #2120**, which justified never waiving an acceptance divergence by claiming pooling "can never" cause one. The conclusion was right and the reason was wrong. The guard keeps its behavior — a ledger one engine refuses to load is too severe to wave through on a heuristic — and the comment now says that instead of a claim this harness refutes.

The units-total guard rests on a claim that does hold: pooling moves units between lot identities and cannot create or destroy them.

Escalated on #2118, and #2121 now warns that a ledger may fail to load rather than only that figures may differ.


robcohen:
Self-review: verified a property this work has been *relying* on without ever testing it. The CI gate added in #2120 tells the reader to "Reproduce one with `--seed <n>`", so that had better be deterministic and had better reproduce what the sweep saw.

Three runs of `--seed 117` are byte-identical, and the single-seed run reproduces the sweep's finding exactly — same header, same classification:

```
seed=117 method=FIFO tie=date price=yes
rledger rejected the ledger; beancount booked it
...
0/1 agreed
0 expected divergence(s) (#2118 lot pooling)
1 unexplained divergence(s)
```

So the instruction the gate prints actually works.

PR #2127, opened 2026-08-24
test(compat): add a capital-gains fuzzer, gated in CI

robcohen:
`report capgains` produces numbers people put on a tax return and had **no generative coverage at all** — the word does not appear anywhere in `scripts/` or `tests/` tooling. It also classifies each closed lot from the *surviving* lot's acquisition date, so it consumes whatever booking decides, and booking produced two real bugs in the last few weeks (#2097, #2115). A wrong term, or a lot attributed to the wrong acquisition, still balances, still reports no error, and still looks entirely plausible.

beancount has no capgains report, so there is no differential partner. This checks **properties** and the binary **against itself**, the way `compat-budget-fuzz.py` checks the budget report:

- `gain == proceeds - cost_basis`, per row
- the term buckets partition the disposals exactly, by count and by sum — a row that reaches neither bucket vanishes from the totals a reader looks at while every individual row still checks out
- `held_days == sold - acquired`
- `--year Y` returns the disposals sold in Y and nothing else
- `--long-term-days N` is exactly "held > N days", which is checkable; the **default** calendar rule is a different function, so the two are compared against each other and may differ **only** where a leap day falls in the span

**Deliberately not asserted:** a hard answer for a 29 February acquisition. The one-year anniversary of 2020-02-29 is genuinely ambiguous, conventions differ, and asserting a guess would encode this script's opinion as the specification. Those spans are generated and reported under `--leap-report`.

## Result: 1000 seeds, zero findings

Recording that as a **negative result** rather than quietly moving on. The arithmetic, the partition, `held_days`, both term rules and the year filter all hold over ~2500 disposals.

A clean sweep is worthless without evidence the harness can fail, so: `--self-test` plants six malformed reports and requires each to be caught, and blinding `check_report` to `return []` fails five of them. Coverage was measured rather than assumed — 766 disposals over 300 seeds, 481 short / 285 long, **426 within 360–372 days of the boundary**, 109 leap-day acquisitions, all four booking methods represented.

## Two things came out of building it

**The generator oversold.** It drew sale quantities from the ledger total rather than from what was held *on the sale date*, so 9% of runs died on `Not enough units` — ledgers rledger rightly refuses, which tests nothing and hides how much coverage was lost. Capping by units available at the sale date took errors 27 → 0, disposals 682 → 766, near-boundary cases 368 → 426.

**A dependency behaviour was unpinned.** `is_long_term` rests on `jiff` resolving a Feb-29 acquisition to Feb-28 — the module comment says so, and nothing checked it. `long_term_calendar_rule_handles_leap_year` uses a span that *crosses* a leap day, which is a different thing from a lot bought *on* one. A `jiff` upgrade resolving to Mar-1 instead would move the boundary a day and silently change the term of every affected disposal, on a figure that goes to a tax return.

Now pinned. The mutation matters here: a general one-day shift of the anniversary is caught by the *existing* test and proves nothing about this case, so I used a **Feb-29-specific** mutation (resolve to Mar-1) — that fails the new test and only the new test.

## CI

Wired into `compat.yml` following the booking and budget fuzzers exactly — `--self-test` first, then the sweep, `continue-on-error` with the verdict in a step output, and a separate gate carrying a reproduce command. 400 seeds on a PR (~15s), 2000 nightly from the run id. actionlint reports the same three findings as `main`, none in the new steps.

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -882,6 +882,40 @@ jobs:
             echo "failed=1" >> "$GITHUB_OUTPUT"
           fi
 
+      # `report capgains` produces numbers that go on a tax return and has no
+      # Python beancount counterpart to diff against, so it is checked by
+      # PROPERTIES and against itself: gain == proceeds - basis per row, the
+      # term buckets partitioning the disposals exactly, held_days matching
+      # the dates, `--year` filtering to what it says, and the calendar
+      # long-term rule agreeing with an explicit `--long-term-days` count
+      # everywhere a leap day does not legitimately separate them.
+      #
+      # It classifies each closed lot from the SURVIVING lot's acquisition
+      # date, so it consumes whatever booking decided — and booking produced
+      # two real bugs in the last few weeks (#2097, #2115). A wrong term or a
+      # lot attributed to the wrong acquisition still balances and still reports no error.
+      #
+      # Same placement and seeding policy as the fuzzers above.
+      - name: Capital gains fuzz
+        id: capgains_fuzz
+        continue-on-error: true
+        run: |
+          if [ "${{ github.event_name }}" = "schedule" ]; then
+            CG_START="${{ github.run_id }}"; CG_RUNS=2000
+          else
+            CG_START=0; CG_RUNS=400
+          fi
+          printf 'start=%s\nruns=%s\n' "$CG_START" "$CG_RUNS" >> "$GITHUB_OUTPUT"
+          if python3 scripts/compat-capgains-fuzz.py --self-test \
+                  --rledger ./target/release/rledger \
+             && python3 scripts/compat-capgains-fuzz.py \
+                  --rledger ./target/release/rledger \
+                  --start-seed "$CG_START" --runs "$CG_RUNS"; then
+            echo "failed=0" >> "$GITHUB_OUTPUT"
+          else
+            echo "failed=1" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Generate YAML summary
         run: |
           cat > compat-results.yaml << EOF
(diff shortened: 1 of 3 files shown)


robcohen:
All three correct, all three fixed in 9072fed02b. Taking them in turn.

**1. The partition check was not a partition check.** You are right that the buckets hold one summary object per *currency* with a `disposals` count, so `len(bucket)` counted currencies, the only count assertion fired when every bucket was empty, and the totals loop skipped `unknown_term` altogether. A dropped row, a wrong `disposals` count, or a missing unknown-term entry all passed.

Worth saying plainly: the PR body claimed this checked the partition "by count and by sum". That was true of the description and not of the code.

It now compares `disposals`, `proceeds`, `cost_basis` and `gain` per `(term, currency)` in **both** directions, so a bucket with no rows behind it and rows reaching no bucket are each a failure.

**2. `other is None` treated a dropped row as agreement.** Correct, and it is the worse of the three, because the failure it hides is exactly the kind the script exists to find. The row sets are now compared before any term comparison, and rows present on only one side are reported with an example key.

**3. `--leap-report` failed the run it was meant to inspect.** Also correct. The notes went into `problems`, so the documented diagnostic mode exited 1 on ledgers the generator produces on purpose. They are now collected separately, printed under their own heading, and never affect the verdict.

## Evidence for each, rather than assertion

* `--self-test` grows from 6 planted reports to 9, adding a wrong `disposals` count, a bucket with no rows behind it, and unknown-term rows reaching no bucket. Blinding `check_report` to `return []` fails 8 of the 9.
* For (2), making `run_capgains` drop one row when `--long-term-days` is passed now reports the drop on every affected seed. Before the change the same mutation was completely silent, which is the hole you described.
* For (3), `--leap-report` over 60 seeds exits 0, reports 21 leap-day acquisitions under their own heading, and leaves the run 60/60 clean. Previously each of those 21 was a failure.

400 seeds at CI's PR setting: still clean.

PR #2138, opened 2026-08-24
fix(core): consume interchangeable lots independently of write order

### LINKED ISSUE #2118: [Compat]: rustledger keeps a slot per acquisition where beancount pools identical lots — FIFO/LIFO consume in different orders

robcohen:
Closes #2118. Implements the rules proposed in rustledger/pta-standards#45, which is in review.

## The defect

Two lots agreeing on commodity, cost per unit, cost currency, acquisition date and label are indistinguishable in every attribute the model records. Booking consumed them in acquisition order anyway, so an unrelated lot written **between** them changed which units were sold.

Three purchases on one date at two costs, written in the six possible orders, measuring remaining cost basis:

| order | before | after | Python beancount |
| --- | --- | --- | --- |
| A B C | 200.00 | 250.00 | 250.00 |
| A C B | 250.00 | 250.00 | 250.00 |
| C A B | 250.00 | 250.00 | 250.00 |
| C B A | 200.00 | 250.00 | 250.00 |
| B A C | 150.00 | 150.00 | 150.00 |
| B C A | 150.00 | 150.00 | 150.00 |

Three outcomes became two. **Sorting a journal or switching importers could move a reported cost basis and a realized gain without changing a fact.**

## Why this direction

No external standard supplies the finer distinction. Treas. Reg. 1.1012-1(j)(3)(i) orders units "in order of time from the earliest **date**" and is silent within one. IAS 2.23 and ASC 330 treat FIFO as a cost-flow assumption over **interchangeable** items, reserving specific identification for items that are not ordinarily interchangeable. Lots identical in commodity, basis and date are interchangeable by definition.

So the ordering below date granularity was ours to invent, and deriving it from where a line was typed was the wrong invention. The full argument is in the RFC.

## Result

The last divergence class the differential booking fuzzer reported against Python beancount is gone:

```
250/250 agreed
0 expected divergence(s) (#2118 lot pooling)
  of which 0 are ACCEPTANCE differences
0 unexplained divergence(s)
```

`main` reports six on the same sweep, two of them journals one engine refuses to load.

## Performance: parity, and that took work

`cost_index` already buckets slots by `(units currency, cost number, cost currency)` in slot order, so the earliest interchangeable slot is the first bucket member agreeing on date and label. Buckets hold one or two entries and the lookup hashes three cheap values.

| | ex-FIFO-2000 | ex-FIFO-20000 |
| --- | --- | --- |
| main | 357ms | 22021ms |
| this branch | 357ms | 21782ms |

LIFO, HIFO and STRICT are at parity too.

Worth recording, because the obvious implementations are the slow ones and I wrote two of them before measuring: a scan inside the sort comparator is **9.7x** (213s at 20k), rebuilding a map per insert **5.7x**, and a dedicated identity map 1.15x to 1.21x. Profiling showed the map was never the cost; rebuilding a lot identity, cloning a currency and a label `String` on every `add`, was.

## Test note

The regression test covers **both** ordering paths deliberately. `try_reduce` on a fresh inventory sorts a scanned list; a reduction after the index exists places later acquisitions through `ordered_index_insert`. An earlier version of the test covered only the scan and **passed with the insert path's lookup disabled**. Each path was verified by mutating it separately.

## This is user-visible

Cost basis, realized gains, surviving lot acquisition dates and `meta.hash` change for any account holding a repeated lot identity within a date, and a reduction naming an explicit cost may begin or cease to resolve. Journals that never repeat a lot identity within a date are unaffected. Changelog entry included.

## Merge order

The RFC is in review and this implements its outcome. If reviewers would rather wait for the RFC to be accepted, hold this; the two are consistent by construction and I would rather they land in that order than have the implementation set the precedent.


robcohen:
Self-review. **The fix is incomplete, and the PR body overstates it.** Found by widening the sweep past the 250 seeds CI runs.

## What I claimed, and what is true

The PR says the divergence class is gone, citing `250/250 agreed, 0 divergences`. That holds at 250 seeds. It does not hold further out:

| sweep | main | this branch |
| --- | --- | --- |
| seeds 0-249 | 6 divergences | **0** |
| seeds 0-999 | 21 divergences (9 FIFO, 12 LIFO) | **2, both LIFO** |
| seeds 5000-5499 | not measured | **1** |

So this removes roughly 90% of the class rather than all of it, and the remainder is real. Claiming elimination on the strength of the seed window CI happens to run is exactly the mistake the expected/unexplained split in #2120 was meant to prevent.

## The residual is LIFO-specific

Reduced from seed 578. The same ledger, changing only the booking method:

* under **FIFO**, rledger and Beancount agree (`2 x 8.75` remaining)
* under **LIFO**, they differ: rledger leaves `2 x 13.25`, Beancount leaves `2 x 12.00`

That points at `LotOrder::DateDescending`. It reverses the primary key while the canonical tie-break stays ascending, which is correct for FIFO and evidently not sufficient for a descending walk: an interchangeable group needs to sit at the position its earliest member occupies, and reversing the sort moves the group somewhere else.

I have **not** confirmed the mechanism, so I am stating it as the leading hypothesis rather than the cause. One hypothesis I did test and disprove along the way: that draining the earliest member of a group shifts the group's position. A minimal ledger doing exactly that agrees with Beancount.

## Also verified while reviewing

* `add` pushes to `cost_index` before calling `ordered_index_insert`, so the canonical lookup always sees a live bucket.
* Drain-then-reinsert is handled: a lot tombstoned and removed from its bucket lets a later interchangeable acquisition become its own canonical, and the result matches Beancount.
* No inline review threads or external reviews on either PR yet.

## What I am doing

Not merging. I will diagnose the LIFO case, fix it, and re-measure at 1000 seeds and on a fresh seed range rather than the range I have been developing against. The PR body will be corrected to state the measured rate rather than the headline.

robcohen:
LIFO case fixed in b522c2bc54. **Correctness is now complete; there is a performance regression I have not recovered.**

## The cause

The first cut derived a group's ordering position from its earliest **surviving** member. That is not the group's position, and the difference is observable: drain the earliest member and the survivors jump to wherever the next one sits, so a group that sat before another lot lands behind it part-way through a ledger.

That is why it was LIFO-specific. The same ledger agrees under FIFO and differs under LIFO, because LIFO reverses the primary key, so a group whose anchor slides ends up on the wrong side of its neighbor.

The anchor is now a monotonic ordinal, assigned when an identity is first acquired and released only when no lot carries it. That is exactly what Python beancount does: a pooled position holds its slot in the inventory dict until its units reach zero, and a re-acquired identity is re-inserted at the end.

**Three places had to agree, and each was a separate bug in the first cut:**

1. `add` assigns the ordinal before the ordered insert reads it.
2. `cost_index_remove` releases it only when the cost bucket holds no other lot with that identity, and asks the **bucket** rather than every slot: scanning `slot_ordinal` was O(lots) per drained lot, worth +23%.
3. `rebuild_index` repopulates the per-slot view after compaction renumbers slots, while keeping `ident_ordinal`.

Missing (3) converted two LIFO divergences into one **unexplained** HIFO divergence, which is a worse failure. The 250-seed sweep would not have shown that either.

## Correctness

| sweep | main | previous commit | now |
| --- | --- | --- | --- |
| seeds 0-999 | 21 | 2 | **0** |
| seeds 5000-5499 | not measured | 1 | **0** |

`1000/1000 agreed, 0 expected, 0 unexplained`. Capgains fuzzer 200/200, 491 core tests, clippy clean on 1.97.

## The regression, stated plainly

| | main | this branch |
| --- | --- | --- |
| ex-FIFO-2000 | 361ms | 389ms (+8%) |
| ex-FIFO-20000 | 21795ms | 24846ms (+14%) |

The cost is the ordinal bookkeeping in `add`. The earlier cost-index approach was at parity and **wrong**, so this is not a trade I can undo by reverting.

Two things I already removed: an O(lots) scan per drained lot (+23%), and a duplicate `cost_key` computation per `add`. What remains is building and hashing a `(CostKey, date, label)` per acquisition, which is inherent to keying the anchor by identity.

I do not think 8-14% is acceptable to merge as-is, and I would rather say so than bury it. Options, roughly in order of appeal:

1. Only assign ordinals when a repeated identity actually exists. A ledger that never repeats a lot identity within a date, which is most of them, would pay nothing.
2. Intern the identity so the key is a `u32` rather than a tuple with a `String`.
3. Accept the cost on the grounds that the pre-existing behavior was wrong.

I am **not** marking this ready to merge. The RFC is also still in review, and if it is rejected this branch should be closed rather than optimized.

robcohen:
Tried the more elegant route and it is genuinely better on every axis I can measure, but it changes something wider than this PR is asking for. Both are pushed so the comparison is on the record: `feat/merge-interchangeable-lots` alongside this branch.

## The realisation

`add` **already merged** cost-less positions, through `simple_slot`. The merge machinery was fully general and only the gate was narrow:

```rust
let merge_idx = position.cost.is_none().then(|| ...)
```

Widening that gate to lots agreeing on cost, cost currency, date and label makes the group ONE position, which is what Rule 1 says it is. The whole anchoring apparatus then evaporates: no `ident_ordinal`, no `slot_ordinal`, no `next_ordinal`, no per-slot anchor, and the sort key returns to `(order_key, slot)` exactly as on `main`. The group holds its slot until its units reach zero and a re-acquisition lands at the end, both for free, because that is what a single position does.

That is also why the anchoring version needed three separate fixes: it was reconstructing, at three call sites, a property that merging gives you by construction.

## It is better on every axis

| | anchoring (this PR) | merging |
| --- | --- | --- |
| fuzzer, seeds 0-999 | 1000/1000, 0 divergences | 1000/1000, 0 divergences |
| ex-FIFO-2000 | +8% | **+2%** |
| ex-FIFO-20000 | +14% | **+5%** |
| code vs the other | baseline | **109 lines fewer** |
| state to keep in sync | 3 maps, 3 call sites | none |

## Why I am not simply switching

It changes what `add` MEANS, not just how lots are ordered. `add(+10 @cost)` followed by `add(-10 @cost)` now nets to one zero-unit position where the codebase deliberately keeps two lots. Six tests fail, and they are a documented decision rather than incidental coverage:

```
test_add_costed_positions_kept_separate
test_merge_keeps_lots_separate
test_add_costed_positions_net_units
test_add_no_cancel_same_sign
iter_slots_yields_indices_that_address_their_own_position
```

with the comment "costed positions are kept as separate lots for O(1) add performance; aggregation happens at display time".

The RFC asks for interchangeable lots to **consume** as one position. It does not ask for `add` to become a netting operation. That is a bigger change, it touches short-position handling, and it deserves its own decision rather than arriving as an optimisation inside a lot-ordering fix.

## What I would do

Merging is the right end state, and the residual regression on this branch is the price of emulating it. But I would rather land the narrow fix and open the merge separately with those six tests examined one at a time, than fold a change to `add`'s contract into a PR whose stated subject is consumption ordering.

If a reviewer prefers the opposite, the branch is ready to evaluate and the six tests are the entire review surface.

robcohen:
Superseded by #2139, which takes the merging approach: same semantics, none of the anchoring machinery, and at parity instead of 8-14% slower.

Leaving this open only so the two can be compared. If #2139 is accepted this should be closed unmerged; if a reviewer would rather not change what `add` means, this branch is the narrower option and its 8-14% regression is the price of emulating merging rather than doing it.

robcohen:
Closing in favour of #2139, which takes the merging approach: same semantics, none of the anchoring machinery, and at parity instead of 8-14% slower.

Keeping the record of why, since the intermediate results are the useful part. This branch emulated "interchangeable lots are one position" by ordering per-acquisition slots around a stable anchor. That needed an identity-to-ordinal map, a per-slot view of it, and a monotonic counter, all kept in step across `add`, `cost_index_remove` and `rebuild_index`. Each of those three was separately wrong at some point during development, because each was reconstructing a property that #2139 gets by construction.

Measured cost of emulating it, against a 22.0s baseline at 20,000 transactions:

  scan inside the sort comparator            213s    9.7x
  map rebuilt per sort and per insert        124s    5.7x
  identity map maintained in the index       26.4s   1.21x
  same map keyed by slot                     25.2s   1.15x
  derived from an existing cost index        21.8s   parity, but incorrect
  merging in `add` (#2139)                   21.8s   parity, correct

The parity row above #2139 is the trap: it was fast and wrong, and only showed up when the sweep was widened past the 250 seeds CI runs.

PR #2139, opened 2026-08-24
fix(core): store interchangeable lots as one position

robcohen:
Supersedes #2138, which should be closed if this is accepted. Same semantics, far less machinery, and at parity rather than 8-14% slower. Implements Rule 1 of rustledger/pta-standards#45.

## The realisation

`add` **already merged** cost-less positions, through `simple_slot`. The merge machinery was fully general and only the gate was narrow:

```rust
let merge_idx = position.cost.is_none().then(|| ...)
```

Lots agreeing on commodity, cost per unit, cost currency, acquisition date and label are indistinguishable in exactly the same way: no attribute the model records separates them. Keeping them apart only let the order they were *written* decide which units a reduction consumed (#2118).

Widening that gate makes the group one position, which is what the RFC says it is and what Python beancount stores. Everything #2138 needed then disappears: no identity-to-ordinal map, no per-slot view, no monotonic counter, no anchor in the sort key. Ordering returns to `(order_key, slot)` exactly as before this work began. The group holds its position until its units reach zero and a re-acquisition after a full drain lands at the end, both because that is what one position does.

That is also why #2138 needed three separate fixes across `add`, `cost_index_remove` and `rebuild_index`: it was reconstructing at three call sites a property merging gives by construction.

| | #2138 anchoring | this |
| --- | --- | --- |
| fuzzer, seeds 0-999 | 1000/1000, 0 divergences | 1000/1000, 0 divergences |
| ex-FIFO-2000 | +8% | **363ms vs main's 384ms** |
| ex-FIFO-20000 | +14% | **21759ms vs main's 21996ms** |
| state to keep in sync | 3 maps, 3 call sites | none |

Faster, for having fewer positions to sort and walk.

## This changes what `add` means, so here are the six tests, one at a time

It is not only an ordering change, and the tests encoding the old contract were examined individually rather than swept:

| test | disposition |
| --- | --- |
| `test_add_no_cancel_same_sign` | Asserted two same-cost same-date buys stay separate: the exact case #2118 is about. Rewritten as `test_add_merges_same_identity`, **plus a new boundary test** that labelled lots still stay apart, since that is what labels are for. |
| `test_add_costed_positions_kept_separate` | A negative of the same identity now nets into the lot. Renamed to say so. Booking a sale is unaffected: that goes through `reduce`, not `add`. |
| `test_add_costed_positions_net_units` | Same case, same disposition. |
| `test_merge_keeps_lots_separate` | `merge` adds each position, so the same identity lands in one. Renamed. |
| `iter_slots_yields_indices_that_address_their_own_position` | **Subject preserved.** It is about slot addressing, not merging; it needed two positions equal by value as a trap for index bugs. They now differ by label, so they stay two lots while still comparing equal on units and cost. |
| `interchangeable_lots_consume_independently_of_write_order` | Mine, from #2138. Rewritten: the property is now stronger, since the lots are literally one position. |

The `add(+10)` / `add(-10)` netting is the part worth a reviewer's attention. It is a real change to a documented contract, and it is why I did not fold this into #2138 as an optimisation.

## Verification

* booking fuzzer seeds 0-999: **1000/1000, zero divergences of any kind**
* capgains fuzzer: 200/200
* 492 core tests, clippy clean on 1.97
* `corpus_parity` fails identically here and on `main` (28 files, stale prebuilt wasm), unaffected


robcohen:
Good catch, and correct on all four sites — fixed in b4d807eba6.

Those comments were written for the anchoring approach on #2138, which ordered per-acquisition slots around a stable anchor. This PR removed that machinery but I left its documentation behind, so the file described a secondary tiebreak the code no longer has.

Corrected:

- `OrderedIndex.by_currency` now says sorted by `order` then slot ascending, and says WHY no further tiebreak is needed: `add` stores interchangeable lots as one position, so a group has one slot rather than several needing to be kept adjacent.
- `ordered_index_insert` no longer claims `add` records an anchor.
- `order_key` says ties fall through to the slot number.
- `rebuild_index` no longer mentions rebuilding an identity map.

I also swept the file for other references to the removed machinery (`anchor`, `ident_ordinal`, `slot_ordinal`, `canonical`) and found none beyond unrelated uses of "canonical" in error-wording docs.

Worth noting the class of defect: this is documentation describing a design that was replaced, which is exactly the kind of thing that survives review and then misleads the next reader. The four sites you listed were all of them.

Issue #1884, opened 2026-07-28
parser: a malformed final line without a trailing newline produces no error


robcohen:
Found while adding CI coverage for `packages/mcp-server`.

## Symptom

`validateSource` returns `valid: true, errors: []` for input the native CLI rejects. Reproduced against a wasm built from current `main` (not just the published 0.21.0):

```
"2024-01-01 invalid directive"  -> valid=true  errors=0
"@@@ not beancount @@@"         -> valid=true  errors=0
```

The same inputs through `rledger check`:

```
exit=1   P0012 (parse error)
```

Same source tree, two different answers. Any wasm consumer silently reports a broken ledger as clean — that includes the MCP server's `validate`/`validate_file` tools and **rustfava**.

## Root cause

`crates/rustledger-wasm/src/helpers.rs`, `load_and_book`:

```rust
let raw = match loader.load(Path::new("input.beancount")) {
    Ok(raw) => raw,
    Err(e) => { /* only a hard Err is surfaced */ }
};
...
match process(raw, &load_options) {
    Ok(ledger) => {
        let errors: Vec<Error> = ledger.errors.into_iter().map(Error::from).collect();
        //  ^ only process() errors; raw.errors is never read
```

`LoadResult.errors` (`crates/rustledger-loader/src/lib.rs:221`) carries the parse-phase diagnostics, and `loader.load()` returns `Ok` with them populated — a malformed file is not a hard `Err`. `grep -c 'raw.errors' helpers.rs` is **0**.

The native path does read it — `crates/rustledger/src/cmd/check.rs:239`: `for load_error in &load_result.errors`.

This is the Canonical-Function Discipline case from CLAUDE.md: a second consumer re-derives the pipeline and drifts. The wasm's early `parse_beancount` guard looks like it covers this, but it doesn't: `rustledger_parser::parse` returns `errors=0, directives=0` for these inputs — it drops unparseable content silently rather than reporting — so the guard never fires and the real diagnostics only ever exist in `raw.errors`.

## Suggested fix

Carry `raw.errors` into the returned `ProcessedLedger.errors` alongside `process()`'s, mapping `LoadError` the way `check.rs` does. Both the `Ok` and `Err` arms of `process(...)` need it.

Worth checking the other `load_and_book` consumers at the same time (`query`, `balances`, `parse`, `format`) — they share this helper, so they inherit the same blind spot.

## Test coverage

`packages/mcp-server` already has three tests that encode the correct behaviour and currently fail:

[...]

PR #2141, opened 2026-08-24
chore(deps): bump @rustledger/wasm to 0.22.0 in mcp-server

robcohen:
Supersedes #2129, which can be closed.

## Why #2129 failed, and why that is good news

The bump does exactly what four `it.fails` tripwires were placed to detect: it makes them pass.

Those markers are not disabled tests. Each pins an assertion that is **correct** but that the bundled wasm could not yet satisfy, and each carries the same instruction in its comment:

> When a release lands and this turns red, delete the `.fails` — that is the signal, not a regression.

- **three** pinned to #1884, where `validateSource` dropped the loader's parse errors and answered `valid: true` for input `rledger check` rejects with P0012
- **one** pinned to the loader reporting a file reached twice

The v0.22.0 wasm carries both fixes, so all four unpinned at once. Cleared, along with their now-historical rationale. This is the mechanism working: whoever bumped the dependency was told what changed, rather than finding a stale skip years later.

## The fifth failure is a real behaviour change

The diamond-include test asserted `valid: true`. The loader now reports `Duplicate filename parsed` for a file reached twice, matching Python beancount, so the verdict flips.

Its **subject is unchanged**: the shared file still contributes once, so the balance holds at 100.00 rather than doubling to 200.00. I kept that intact and added a message check, so a doubled balance would still surface as a balance-assertion failure rather than passing quietly under the new verdict.

## Verification

* 121/121 tests pass
* `npm run build` clean
* `package-lock.json` resolves `@rustledger/wasm` at 0.22.0

### DIFF
--- a/packages/mcp-server/package-lock.json
+++ b/packages/mcp-server/package-lock.json
@@ -10,7 +10,7 @@
       "license": "GPL-3.0-only",
       "dependencies": {
         "@modelcontextprotocol/server": "^2.0.0",
-        "@rustledger/wasm": "^0.21.0"
+        "@rustledger/wasm": "^0.22.0"
       },
       "bin": {
         "rustledger-mcp": "dist/index.js"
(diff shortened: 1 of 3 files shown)


robcohen:
Good catch, taken.

`JSON.stringify(result).toContain('Duplicate')` matched anywhere in the payload, so any other `Duplicate ...` error, or a formatting change that happened to contain the word, would have satisfied it. It now checks `result.errors[..].message` for `Duplicate filename parsed`, which is the convention the sibling test already uses.

Verified the assertion is actually stronger rather than just longer: swapping the expected text to `Duplicate account name` fails the test, where the stringified form passed.

PR #2143, opened 2026-08-25
test(compat): add a returns fuzzer, gated in CI

robcohen:
`report returns` answers "what did this portfolio earn". It is hard to eyeball, easy to get subtly wrong, and had no generative coverage. It also has two independently wrong-able halves — **extraction** (which postings are cash flows, and what they are worth) and the **solve** (the rate that zeroes their NPV) — so this checks both, from three directions.

## beangrow is the partner for extraction, not the oracle for the rate

beangrow is an independent implementation from the beancount project over the same ledgers, so agreement on the flow list is real evidence rather than a restatement of our own logic.

But it is **not** trustworthy for the rate, and this does not treat it as such. Its `compute_irr` runs `fsolve` from a fixed seed against an objective with two spurious roots: `npv(-(2+t)) == -npv(t)` mirrors every true root, and every term underflows as `r -> -1`. It reports `"converged"` on both. On a two-transaction profitable position it answers **-100%** where the true IRR is **+102.28%** — filed upstream as [beancount/beangrow#51](https://github.com/beancount/beangrow/issues/51) from this work.

So the rate is adjudicated by an independent bisection solve over the *same* flows, bracketed away from the asymptote, and a divergence names which engine the referee agrees with.

Verdicts stay separate — `EXTRACTION`, `SOLVER`, `AGREE` — so a flow-list difference is never mistaken for a solver difference.

## Two traps, both of which produced false divergences first

* **beangrow's `configure.py` leaves `dividend_accounts` empty** when a dividend never touches the asset account, while rledger is told them via `-n`. Without configuring it, the comparison pits a configured engine against an unconfigured one: **13 false positives in 60 seeds**.
* **`ad.cash_flows` omits the terminal valuation.** Only `truncate_and_merge_cash_flows` appends it. Reading the raw list turned every open position into a spurious divergence and reported -41% against a true +21.87%.

The generated ledger declares `commodity ACME` deliberately: beangrow's `find_accounts` only considers an account whose *leaf* names a declared commodity, so without it the config infers zero investments and the whole comparison is vacuous rather than clean.

## Result

**1000 seeds, zero divergences of either kind.**

A clean sweep proves nothing unless the harness can report dirty, so each check is shown to fail:

* making the referee return the `-100%` asymptote → self-test fails
* dropping beangrow's terminal flow → extraction check fires on every affected seed

I also rejected one mutation as inconclusive rather than counting it: blinding the extraction comparison left 40/40 agreeing, which only means those seeds had no difference to detect. Planting an actual difference is what proves the check.

## CI

Wired into `compat.yml` on the same pattern as the booking and capgains fuzzers — `--self-test` first, then the sweep, `continue-on-error` with the verdict in a step output, and a separate gate carrying a reproduce command. 400 seeds on a PR (~16s), 2000 nightly from the run id. `beangrow` added to the compat job's pip install. actionlint reports the same three findings as `main`.

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -74,6 +74,8 @@ jobs:
       - name: Install dependencies
         run: |
           pip install beancount beanquery matplotlib pyyaml
+          # beangrow: the differential partner for `report returns`.
+          pip install beangrow
           # Third-party plugins used by test files. Compat tests rely on
           # them — a silent install failure means tests using these
           # plugins get skipped and we get false-green CI. Fail loudly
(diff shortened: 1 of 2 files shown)


robcohen:
All seven correct, all seven fixed in 887181fb33. Several are the exact failure modes this harness exists to catch, which is the uncomfortable part of the review.

**1. The nightly gate could never have run.** The step sat *after* `Commit results to compatibility branch`, which checks out an orphan branch with no `scripts/` and no `target/`. That hazard is documented at `compat.yml:819` for the budget fuzz, I read it while writing the step, and I still anchored on the **gate** rather than on the other fuzzers. Moved to sit beside Booking and Capital gains, before the checkout.

**2. A run that compared nothing exited 0.** Both engines turn load and execution failures into `skip`, so an incompatible beangrow API would print `0/0 agreed` and pass. Now an executed count of zero fails, as does skipping more than half the seeds. Verified: forcing every seed to skip exits 1 with `the gate proved nothing`.

**3. The referee gave up on steep returns.** Correct about the magnitude — a 20-day 10.00 → 16.50 sale annualizes past 900,000%, far past the fixed 10,000% bracket, so `bisect_irr` returned `None` and the solver check silently skipped exactly the cases most likely to break a Newton solver. The bracket now grows geometrically.

**4. The `closer:` label was unreachable.** Reaching that branch already means rledger differs from the referee by more than tolerance, so the ternary could only print `referee` while reading as though it had adjudicated between two engines. Replaced with both rates and their delta.

**5. `want` computed and never compared.** The self-test accepted any non-`None` result, including a wrong bisection. CodeQL flagged the unused variable; the variable was the *symptom* and the missing assertion was the defect. It now asserts to 1e-6.

**6. The self-test never exercised the dirty verdicts.** Right — it only ran a clean seed, so deleting either comparison would have left `--self-test` green. It now drives both branches with planted values. Writing that caught a further error of mine: my first planted fixture put a terminal flow in the distributions column, and the self-test told me so.

**7. Bisection assumed a single sign change.** A series with several has several real roots, and picking one arbitrarily would let the referee adjudicate toward a root rledger never claimed. It now scans the bracket and declines when the count is not exactly one — `None` means "no verdict", which is honest where a confident wrong answer is not.

## And one overclaim corrected in the text rather than the code

The PR described a "flow-list differential". `report returns --format json` exposes aggregates and a count, **not** dated flows, so a change preserving all three sums *and* the count while moving a flow's DATE is invisible to it. The docstring now states that limit and names what would close it (a diagnostic flow dump from rledger). The count is now compared, which it was not before.

## Verification

1000 seeds still agree, self-test passes, actionlint unchanged from `main` (3 findings, all pre-existing).

PR #2145, opened 2026-08-25
fix(format): keep currency, cost and price on a numberless posting

### LINKED ISSUE #2142: [Bug]: `rledger format` deletes the currency and price/cost annotation when a posting's units number is missing
## Summary

`rledger format` silently deletes the units currency and any price or cost annotation from a posting whose units NUMBER is missing. The tokens are not reformatted or flagged: they are dropped from the output.

Same class as #1332 (`format` deletes some comments), different token.

## Reproducer

```beancount
2013-05-18 open Assets:Investments:MSFT MSFT
2013-05-18 open Assets:Investments:Cash USD

2013-05-18 * "buy"
  Assets:Investments:MSFT            MSFT @@ 2000.00 USD
  Assets:Investments:Cash   -2000.00 USD
```

```
$ rledger format --no-ledger t.beancount
2013-05-18 open Assets:Investments:MSFT MSFT
2013-05-18 open Assets:Investments:Cash USD

2013-05-18 * "buy"
  Assets:Investments:MSFT
  Assets:Investments:Cash  -2000.00 USD
```

`MSFT @@ 2000.00 USD` is gone. What the author wrote cannot be recovered from the output.

## Why this matters even though the input is invalid

Python beancount rejects this input too (`Total price on a posting without units`), so the file was already wrong. That is exactly the case where deleting content is worst:

* `format` is documented as running **on editor save** ("`format` finds the nearest root journal ... matching what the editor does on save"). A user who mistypes a units number and saves loses the rest of the posting.
* The remaining diagnostic gets *less* informative. Before formatting the file reports `BOOK`; afterwards it reports `E1001` twice, because the evidence of what the posting meant is gone.
* A formatter's contract is to change layout, not content. Refusing to format an unparseable posting, or passing it through verbatim, are both fine. Silently discarding it is not.

Valid input is unaffected: `10 MSFT @@ 2000.00 USD` round-trips correctly, so the trigger is specifically the missing units number.

## Scope

Found by checking two formatter properties across the 735-file compatibility corpus:

* `format(format(x)) == format(x)` — **no violations** in 693 checked files
* `check(format(x))` reports the same diagnostic codes as `check(x)` — **3 violations**, all this bug

The three corpus files:

```
tests/compatibility/files/parser-lima/test-cases_TotalsAndSigns.TotalPriceWithMissing2.beancount
tests/compatibility/files/parser-lima/test-cases_IncompleteInputs.UnitsMissingNumberWithPrice.beancount
tests/compatibility/files/parser-lima/test-cases_IncompleteInputs.UnitsMissingWithCost.beancount
```

The last one shows the same loss for a `{cost}` annotation rather than a price, so it is not specific to `@@`.

## Suggested fix

Emit the posting's original text when its units number is absent, rather than rendering from a partially-populated AST. The lossless CST (#1262) should already retain the tokens needed.

Worth a regression test asserting the property directly: formatting must not change the set of diagnostics a file produces. That is what found this, and it generalizes past this one token.

robcohen:
Closes #2142.

`emit_posting` rendered the amount only when `amount_number_text` yielded a number, and the currency, cost spec and price annotation all sat inside that branch. A posting without a number lost every one of them — not reformatted, **deleted**:

```
  Assets:Investments:MSFT            MSFT @@ 2000.00 USD
```
```
  Assets:Investments:MSFT
```

## Worse than the issue first recorded

While fixing it I found the bug reaches **valid** input:

```beancount
2024-01-15 * "x"
  Assets:Bank  -5.00 USD
  Assets:Other USD          ; valid: constrains interpolation to USD
```

Both engines accept this and interpolate `Assets:Other 5.00 USD`. Formatting deleted the `USD`. So this was data loss on correct input, not only on malformed input — I filed the issue believing otherwise, and the fix and its tests cover both.

The malformed shapes still matter, because `format` runs on editor save: mistyping a units number and saving deleted the rest of the line, and the surviving diagnostic got *worse* for it (`BOOK` became `E1001` twice, once the evidence of intent was gone).

## The fix

Numberless postings emit their parts with two spaces rather than the aligned number column. There is no number to align, and `compute_alignment` deliberately excludes these postings from the column width so a long currency-only account cannot widen the file — the existing `transaction_currency_only_posting_does_not_widen_amount_column` test still passes.

## Testing

The regression test covers all four shapes — currency only, total price, cost, per-unit price — and **asserts idempotence on each**, so a fix that re-emitted tokens in a shape it could not reparse would fail here rather than silently.

Verified it fails against the old code: reverting gives `formatting deleted "@@ 2000.00 USD"`.

The corpus property sweep that found this now reports **3 fewer drifts** (5 → 2). The 2 remaining are artifacts of formatting to a temp file, which breaks relative `include` and plugin paths — not formatter bugs.

Parser tests green, workspace green except the pre-existing `corpus_parity` failure, clippy clean on 1.97.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ Shipped work, by area. Forward-looking plans live in [docs/roadmap/](docs/roadma
 
 ### Fixed
 
+- **`rledger format` no longer deletes a posting's currency, cost or price when it has no units number** — `emit_posting` rendered the amount only when a number was present, and the currency, cost spec and price annotation all sat inside that branch, so a posting without a number lost every one of them. The tokens were not reformatted, they were dropped, and what the author wrote could not be recovered from the output. `Assets:Other USD` is valid beancount that constrains interpolation to USD, so this was data loss on CORRECT input; the malformed shapes matter too, since `format` runs on editor save and mistyping a units number then saving deleted the rest of the line. Found by checking two properties across the 735-file compatibility corpus: formatting is idempotent, and it does not change the set of diagnostics a file produces (issue #2142).
+
+
 - **`report journal --limit N` no longer reverses the report** — the limit was reached with `.rev().take(n)`, which left the ROWS reversed as well as selecting the last N: the unlimited journal read oldest-first and adding a limit silently flipped it newest-first. `sort_by_key` is stable, so the same reversal also undid the file order of same-date transactions — the order booking itself uses (#2093) — so a limited journal disagreed with the ledger about which of two same-day entries came first. All three output formats were affected; they share the selection. Found by variant analysis from #2115, which was the same `.rev()`-over-a-sorted-list shape in the booking path (issue #2122).
 
 ### Performance
--- a/crates/rustledger-parser/src/cst/format.rs
+++ b/crates/rustledger-parser/src/cst/format.rs
@@ -2056,7 +2056,34 @@ fn emit_posting(
                 out.push(' ');
                 out.push_str(&format_price_annotation(&pa, group));
             }
+        } else {
+            // No NUMBER, but the posting may still carry a currency, a cost
+            // spec or a price. `Assets:Other USD` is valid beancount that
+            // constrains interpolation to USD, and a units-less posting with
+            // a cost or price is malformed but is still what the author
+            // wrote. Emitting nothing here deleted all of it (#2142).
+            //
+            // Two spaces rather than the aligned number column: there is no
+            // number to align, and `compute_alignment` deliberately excludes
+            // these postings from the column width so a long currency-only
+            // account cannot widen the file.
+            for part in [
+                amt.currency().map(|c| c.text().to_string()),
+                p.cost_spec().map(|cs| format_cost_spec(&cs, group)),
+                p.price_annotation()
+                    .map(|pa| format_price_annotation(&pa, group)),
+            ]
+            .into_iter()
+            .flatten()
+            {
+                out.push_str("  ");
+                out.push_str(&part);
+            }
         }
+    } else if let Some(cs) = p.cost_spec() {
+        // No amount node at all, yet a cost spec parsed. Same rule: keep it.
+        out.push_str("  ");
+        out.push_str(&format_cost_spec(&cs, group));
     }
     out.push('\n');
     // Splice the trailing comment in BEFORE the posting-line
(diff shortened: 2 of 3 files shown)


robcohen:
Both correct, both fixed in c2647805cb.

**The fix was incomplete.** The no-AMOUNT-node branch handled `cost_spec` and not `price_annotation`, so `Assets:MSFT @@ 2000.00 USD` still lost its price while `Assets:MSFT {12.00 USD}` was kept. Confirmed before fixing:

```
  Assets:MSFT              <- price deleted
  Assets:MSFT  {12.00 USD} <- cost kept
```

Both are now emitted from one loop rather than separate branches, which is what allowed one to be forgotten. Two regression cases added for that shape, which nothing covered.

**The test could pass without testing anything.** You are right that `must_contain = "USD"` is satisfied by the other posting's `-5.00 USD`, so the currency-only case would have stayed green while the currency was still being deleted. Every expectation now names the whole posting line.

Verified the difference rather than assuming it: mutating the formatter to drop **only** the currency, keeping cost and price, now fails with `formatting deleted "Assets:Other  USD"`. The old assertion passed that same mutation.

That is the second time today I have accepted this exact correction — a test written to prove content survives must not be satisfiable by unrelated content. The first was on #2141, in someone else's shape; this one was mine.

Corpus sweep unchanged at 2 drifts, both known temp-file artifacts. Parser tests green, clippy clean on 1.97.

robcohen:
Self-review after the last fix, since the incomplete first cut suggested the shape space was worth enumerating rather than sampling.

Every combination of currency / cost / price on a numberless posting, and the two controls:

| input | output |
| --- | --- |
| `Assets:A USD {12.00 USD}` | `Assets:A  USD  {12.00 USD}` |
| `Assets:A USD @ 3.00 EUR` | `Assets:A  USD  @ 3.00 EUR` |
| `Assets:A {12.00 USD} @ 3.00 EUR` | `Assets:A  {12.00 USD}  @ 3.00 EUR` |
| `Assets:A` | `Assets:A` (auto-balance, correctly unchanged) |
| `Assets:A 5 USD` | `Assets:A      5 USD` (normal alignment preserved) |

No double-emission from the two branches, nothing dropped, and the bare-account and normal-amount controls behave as before.

All four numberless shapes are idempotent, which was the risk worth checking: emitting with two spaces rather than the aligned column could in principle reparse into something that formats differently again. It does not.

robcohen:
Extended the fix with a corpus-level gate in 4b2c34bcde, after noticing why this class escaped in the first place.

**CI already ran `format-idempotence.sh` over the corpus.** That is exactly why #2142 got through: deleting the same tokens twice is perfectly idempotent. A unit test alone would leave the same blind spot for the next token.

The script now also checks that formatting does not change the set of diagnostic CODES a file produces. Formatting is a layout change; if `check` says something different afterwards, content moved or vanished. Codes only and sorted, since wording and spans legitimately move when layout does.

**Both "harness artifacts" I reported earlier were my own doing.** Writing the formatted copy to `/tmp` breaks relative `include` and plugin paths. Writing it beside the original resolves both, so the corpus is genuinely at **0 drifts**, not 2, and this is a clean gate rather than one with permanent exceptions.

**Verified in both directions**, which is the only way it is worth having:

- against a binary built **before** the #2142 fix: reports the 3 affected files, exits 1
- against the fixed binary: reports none, exits 0

That second run earned its keep immediately. My first version grepped `"code":"` while the JSON is pretty-printed as `"code": "E3004"` — it matched nothing, compared empty against empty, and passed every file including the ones with the bug. It looked exactly like a clean run, and only running it against the known-bad binary exposed it.

Issue #2149, opened 2026-08-25
Same-date directives from different included files book in a different order than beancount, changing realized gains


robcohen:
`booking_sort_key` documents an equivalence to beancount that does not hold once a ledger uses `include`. The result is a silent difference in which lot a FIFO sale consumes, and therefore in realized capital gains.

## The claim

`crates/rustledger-core/src/directive.rs`:

```rust
/// This is a stable sort, so directives sharing a date and type keep their
/// file order - which is the order they are booked in, matching Python's
/// `(date, type_priority, lineno)`.
```

That holds inside a single file, where parse order and `lineno` rise together. Across includes it does not. Python compares `lineno` values taken from different files, so a directive on line 1 of a file included second sorts ahead of one on line 5 of a file included first. rustledger uses stable sort over `(date, priority)`, so it keeps include order instead.

The two rules agree only when they happen to coincide.

## Reproduction

`first.beancount`, included first, transaction on line 5:

```beancount
; this file is included FIRST, but its transaction sits on line 5
;
;
;
2024-06-01 * "buy-at-10"
  Assets:S      1 HOOL {10.00 USD}
  Assets:Cash            -10.00 USD
```

`second.beancount`, included second, transaction on line 1:

```beancount
2024-06-01 * "buy-at-20"
  Assets:S      1 HOOL {20.00 USD}
  Assets:Cash            -20.00 USD
```

`ledger.beancount`:

```beancount
option "booking_method" "FIFO"
2024-01-01 open Assets:S HOOL
2024-01-01 open Assets:Cash USD
2024-01-01 open Income:Gains USD

include "first.beancount"
include "second.beancount"

2024-09-01 * "sell one unit"
  Assets:S     -1 HOOL {}
  Assets:Cash   30.00 USD
  Income:Gains
```

Both buys share a date, so FIFO's lot-date comparison ties and the tiebreak falls through to insertion order. The two implementations insert in different orders:

| | lot consumed | `Income:Gains` |
|---|---|---|
| beancount 3.x | 20.00 | -10.00 USD |
| rustledger | 10.00 | -20.00 USD |

A 2x difference in reported gain, decided by which file a buy lives in and what line it happens to sit on. No error or warning on either side.

## Control

Swapping the two `include` lines pins down each rule. beancount is invariant, always booking the `lineno` 1 buy first, so it reports -10.00 either way. rustledger tracks include order, so it reports -20.00 in one arrangement and -10.00 in the other.

## Which behavior is right

[...]

PR #2187, opened 2026-08-30
fix(booking): stop padding a balance on the pad's own date

### LINKED ISSUE #2150: Same-date pad and balance inserts a phantom padding transaction that its own validator calls unused
A `pad` and the `balance` it targets on the same date make rustledger insert a padding transaction that beancount does not. The account ends up 60.00 USD richer than in beancount, and rustledger's own validator reports the pad as unused in the same run that books it.

## Reproduction

```beancount
2024-01-01 open Assets:A USD
2024-01-01 open Equity:Opening USD

2024-01-05 * "some activity"
  Assets:A    40.00 USD
  Equity:Opening

2024-06-15 pad Assets:A Equity:Opening
2024-06-15 balance Assets:A  100.00 USD
```

| | pad applied | final `Assets:A` |
|---|---|---|
| beancount 3.x | no, "Unused Pad entry" | 40.00 USD |
| rustledger | yes, 60.00 USD | 100.00 USD |

Both report the balance failure. Only rustledger mutates the ledger:

```
$ rledger query t.beancount --format csv "SELECT date, narration, account, number WHERE account='Assets:A'"
2024-01-05,some activity,Assets:A,40.00
2024-06-15,(Padding inserted for Balance of 100.00 USD for difference 60.00 USD),Assets:A,60.00
```

`check` exits non-zero, but the padded figure is what the read paths return. Verified on two:

```
$ rledger query t.beancount --format csv "SELECT sum(number) WHERE account='Assets:A'"
100.00

$ rledger report t.beancount balances
Assets:A
           100.00 USD
```

So a consumer that reads rather than checks sees a 60.00 phantom.

I have not verified the FFI path and am not claiming it behaves the same. It is worth checking separately, because `expand_pads` in `crates/rustledger-ffi-wasi/src/helpers.rs` runs its own date-only sort (line 305) before placing synths, so it could diverge from the CLI in either direction here.

## Three things disagree internally

`rledger check` on that file emits both of these at once:

```
E2001 - Balance failed for Assets:A: expected 100.00 USD, got 40.00 USD
E2003 - Unused Pad entry
```

E2003 is "Pad without a subsequent balance assertion". The validator concludes the pad has nothing to satisfy. Booking pads anyway. One of the two is wrong.

The doc on `merge_with_padding` (`crates/rustledger-booking/src/pad.rs`) states the intent:

```
/// On a same-date pad+balance pair (legal in beancount), the synth must
/// appear BEFORE the balance so any consumer that checks balance assertions
/// mid-stream sees the correct inventory.
```

Both halves are wrong.

**"legal in beancount"** is not: beancount emits `PadError: Unused Pad entry` for exactly this input and inserts nothing.

**The mechanism is inert.** Balance assertions are start-of-day, not mid-stream, in both implementations. Verified directly, with the transaction written before the assertion in source:

```beancount
2024-06-15 * "same-date txn BEFORE the assertion in source"
  Assets:A    40.00 USD
  Equity:O   -40.00 USD

2024-06-15 balance Assets:A  40.00 USD
```

beancount and rustledger both fail it, rustledger with `expected 40.00 USD, got 0 USD`. A same-date transaction never counts toward that day's assertion, so ordering a same-date synth ahead of the balance cannot make the assertion see it. The ordering work in `merge_with_padding` buys nothing, which is why E2001 still fails at 40.00 while the synth is nonetheless in the ledger.

## Likely root cause

`process_pads` orders its working copy by date alone:

```rust
// crates/rustledger-booking/src/pad.rs:147
sorted.sort_by_key(|d| d.date());
```

No priority component, so a same-date pad and balance resolve by source position, and a pad written above its balance is treated as preceding it. Python's `entry_sortkey` is `(date, SORT_ORDER, lineno)` with `Balance = -1` and `Pad = 0`, so the balance sorts first and the pad finds no subsequent assertion.

Note this is not the same ordering as `booking_sort_key`, which does carry a priority and ranks `Pad` before `Balance`. Neither matches Python here, and the two disagree with each other.

## Control

Normal usage is unaffected. With the pad one day earlier:

```beancount
2024-06-14 pad Assets:A Equity:Opening
2024-06-15 balance Assets:A  100.00 USD
```

both implementations report no errors and a final balance of 100.00 USD. The divergence is specific to the same-date pair.

## Suggested resolution

Decide whether a same-date pad is unused (beancount's answer) or applies. Whichever is chosen, E2003 and the booking phase have to agree, since today one of them is always wrong. If it is unused, `process_pads` should not synthesize, and the `merge_with_padding` doc should drop both the "legal in beancount" claim and the mid-stream rationale. If it applies, E2003 must stop firing and the balance semantics need stating, because start-of-day assertions mean the pad cannot satisfy the assertion that motivated it.

Related but distinct: #1300 covered multiple pads preceding one balance.

robcohen:
## What

A `pad` and the `balance` it targets **on the same date** synthesized a padding
transaction beancount does not:

| | pad applied | final `Assets:A` |
|---|---|---|
| beancount 3.x | no — "Unused Pad entry" | **40.00 USD** |
| rustledger (before) | yes, 60.00 USD | **100.00 USD** |

`check` exited non-zero, but every read path — `query`, `report balances` —
returned the padded figure. And it reported the pad as **unused in the same run
that booked it**.

## Cause

beancount checks a balance at the **start** of its day: `entry_sortkey` gives
Balance `-1` and Pad the default `0`, so a same-date balance is evaluated
before the pad exists and consumes nothing.

Two places disagreed with that:

- `DirectivePriority` had `Pad = 2` before `Balance = 3`, with a comment
  asserting *"padding before balance assertions"* — the wrong belief, written
  down.
- `process_pads` sorted by **date alone**, so same-date directives kept source
  order: a pad written above its balance was processed first and the balance
  consumed it.

Both fixed — Balance now precedes Pad, and `process_pads` sorts by the
canonical `booking_sort_key`.

## Two tests pinned the old behavior

Both are inverted, with the evidence in their comments:

- `test_merge_with_padding_same_date_pad_balance_synth_comes_first` asserted a
  same-date synth exists. The *ordering* guarantee it was really protecting —
  that a synth precedes its balance — is preserved in a new test on the
  day-earlier case, which is where a synth actually exists.
- `test_sort_directives_pad_before_balance`, wrong down to its name.

## Verification

- The fixture returns **40.00**, matching beancount. A pad one day earlier
  still pads to **100.00**, also matching — so the fix is specific to the
  same-date case rather than disabling padding.
- `check` still emits E2003 **and** E2001, which is exactly what `bean-check`
  emits for this file. The difference is that we no longer book the pad we call
  unused.
- **Booked values unchanged across all 23 pad-bearing corpus files**, compared
  against a `main`-built binary — with a negative control confirming the
  comparison detects the issue's own fixture changing.
- **The validator and the booker now agree.** On the reported file we emit
  E2003 *and* synthesize nothing; on the three padding cases we emit no E2003
  and do pad. All four match `bean-check`'s judgement. That contradiction —
  calling a pad unused in the same run that booked it — was the issue's third
  complaint.
- **Blast radius confined to pad-bearing files**: 59 pad-free files produce
  byte-identical `#entries` output.
- The four new `corpus_parity` divergences are the prebuilt component still
  using the old order — `native=(balance) component=(pad)` — which is what a
  correct engine change looks like against a stale component.

## Two things I want to be explicit about

**The booked-value gate reports clean, and that is not evidence here.** Its
axis is prices; it does not compare balances, so it cannot see this change. I
verified values by diffing per-account sums against a `main` binary instead.

**Synth placement within a date is still not beancount's, and this PR moved it.**
For a pad sharing a date with an *unrelated* balance, `#entries` gives
`transaction balance pad` where bean-query gives `balance pad transaction`;
before this PR it was `pad transaction balance`. All three produce the same
account total, so it is placement rather than arithmetic — filed as #2188,
along with the now-stale `pad_insertion_index` special case that favours a
same-date balance.

**Same-date ordering is still not fully beancount's.** beancount interleaves
pads and transactions by `lineno` (both sort at `0`); we group by type. This PR
fixes Balance-before-Pad only. The interleaving is #2149, unchanged here.

Closes #2150

### DIFF
--- a/crates/rustledger-booking/src/pad.rs
+++ b/crates/rustledger-booking/src/pad.rs
@@ -449,6 +449,13 @@ where
 {
     let mut end_of_group = 0;
     for (index, directive) in sorted.into_iter().enumerate() {
+        // Leftover from before #2150: this exists to put a synth AHEAD of a
+        // balance on the pad's own date. Such a pad no longer synthesizes at
+        // all -- the balance is checked first and the pad is unused -- so this
+        // arm now only fires for an UNRELATED same-date balance, where placing
+        // the synth ahead of it contradicts the Balance-before-Pad order.
+        // Removing it changes synth placement on the FFI path, which is its
+        // own decision; tracked in #2188 rather than ridden in here.
         if directive.date() == date && matches!(directive, Directive::Balance(_)) {
             return index;
         }
--- a/crates/rustledger-booking/src/pad.rs
+++ b/crates/rustledger-booking/src/pad.rs
@@ -899,16 +899,12 @@ mod tests {
 
         let merged = merge_with_padding(&directives);
 
-        let synths: Vec<_> = merged
+        let synths = merged
             .iter()
-            .filter_map(|d| match d {
-                Directive::Transaction(t) if is_synthesized_pad(t) => Some(t),
-                _ => None,
-            })
-            .collect();
+            .filter(|d| matches!(d, Directive::Transaction(t) if is_synthesized_pad(t)))
+            .count();
         assert_eq!(
-            synths.len(),
-            1,
+            synths, 1,
             "the pad must satisfy the LATER balance, producing exactly one synth"
         );
     }
--- a/crates/rustledger-booking/src/pad.rs
+++ b/crates/rustledger-booking/src/pad.rs
@@ -142,7 +142,6 @@ pub fn process_pads(directives: &[Directive]) -> PadResult {
     let mut padding_transactions = Vec::with_capacity(num_directives.min(16));
     let mut errors = Vec::with_capacity(4);
 
-    // Sort directives by date for processing
     let mut sorted: Vec<&Directive> = directives.iter().collect();
     // Sort by the canonical booking key, not by date alone. Date-only left
     // same-date directives in source order, so a `pad` written above the
--- a/crates/rustledger-booking/src/pad.rs
+++ b/crates/rustledger-booking/src/pad.rs
@@ -144,7 +144,13 @@ pub fn process_pads(directives: &[Directive]) -> PadResult {
 
     // Sort directives by date for processing
     let mut sorted: Vec<&Directive> = directives.iter().collect();
-    sorted.sort_by_key(|d| d.date());
+    // Sort by the canonical booking key, not by date alone. Date-only left
+    // same-date directives in source order, so a `pad` written above the
+    // `balance` it targets was processed first and the balance consumed it --
+    // synthesizing a padding transaction beancount does not (#2150). The key
+    // puts Balance before Pad, so that balance is seen first and the pad ends
+    // up with nothing to satisfy.
+    sorted.sort_by_key(|d| rustledger_core::booking_sort_key(d));
 
     for directive in sorted {
         match directive {
@@ -339,12 +345,20 @@ fn create_padding_transaction(
 ///
 /// # Sort ordering on date ties
 ///
-/// Synth transactions carry the pad's date, not the balance's date.
-/// On a same-date pad+balance pair (legal in beancount), the synth must
-/// appear BEFORE the balance so any consumer that checks balance assertions
-/// mid-stream sees the correct inventory. This is achieved by prepending
-/// the synth list to the original directives before the stable sort:
-/// synths land at the front of their date-group, originals follow.
+/// Synth transactions carry the pad's date, not the balance's date. When a
+/// pad precedes its balance by at least a day, the synth must appear BEFORE
+/// that balance so any consumer checking assertions mid-stream sees the
+/// padded inventory. This is achieved by prepending the synth list to the
+/// original directives before the stable sort: synths land at the front of
+/// their date-group, originals follow.
+///
+/// A pad and the balance it targets on the SAME date produce no synth at
+/// all. beancount checks a balance at the start of the day, so it is ordered
+/// ahead of a same-date pad and consumes nothing; the pad is reported unused.
+/// This doc previously called that pairing "legal in beancount" and required
+/// the synth to precede the balance -- which is what produced a padding
+/// transaction beancount does not, leaving the account richer by the
+/// difference while our own validator called the pad unused (#2150).
 ///
 /// # Errors are discarded
 ///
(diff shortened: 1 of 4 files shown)


robcohen:
Rebased on main and added a wasm `CACHE_VERSION` bump (15 → 16) this PR was missing.

`LedgerPayload` archives the output of the shared `process()` pipeline, which sorts by `canonical_sort_key` — and that key reads `DirectivePriority`. Swapping `Pad` and `Balance` changes the **list** this cache holds without changing a single directive in it: same bytes per directive, different order, and the pad amounts that went with the old order. A v15 blob read by a fixed build serves exactly the arrangement this PR removes.

Nothing in the suite catches that class today — the version pin only forces a question, and it is answerable "loader-internal" when it is not. #2189 adds `processed_ledger_archived_form_is_pinned`, which fails on this exact swap; that is how the gap was found.

robcohen:
Correction to the verification notes on this PR.

I reported `corpus_parity` divergences as "20, unchanged from main (pre-existing)". That measurement was wrong. The test loads a prebuilt `target/wasm32-wasip2/debug/rustledger_ffi_component.wasm`, and **no test command rebuilds it** — the copy on my disk was three weeks old, so it was comparing today's native code against a stale component.

After `cargo build -p rustledger-ffi-component --target wasm32-wasip2`:

```
cargo test -p rustledger-ffi-component-tests   ->  ok, 0 divergences
cargo test --workspace                          ->  green
```

So there is no known-failing test here at all; the actual state is better than I described. The tell I missed was the number's *stability* — it stayed 20 because the component never changed, not because the failure was inherent. Every other verification figure I quoted (clippy, fmt, typos, the 693-file format scripts, the baselines) was measured directly and stands.

robcohen:
Fixed in the rebase (d533cd43d2). You were right that the comment contradicted the code — it said "Sort directives by date for processing" directly above the block explaining that date alone was the bug. I removed the stale line rather than reword it; the comment underneath already states the rule and why it changed.

Also folded into that commit: the `cache.rs` variant census now identifies variants by `Directive::type_name` instead of `Debug` on `mem::Discriminant`, whose format is not a documented guarantee.

Issue #2160, opened 2026-08-27
note silently discards its tags and links: parsed, accepted, and never stored


robcohen:
`note` accepts `#tag` and `^link`, parses clean, and silently discards both. There is no way to get them back: the `Note` directive has no fields to hold them.

## Reproduction

```beancount
2024-01-01 open Assets:A USD
2024-01-05 note Assets:A "note text" #ntag ^nlink
```

```
bean-query   Note: tags=['ntag'] links=['nlink']
rledger      check exits 0, and every route to them is empty
```

```
$ rledger query t.beancount "SELECT type, tags, links FROM #entries WHERE type='note'"
note,,

$ rledger query t.beancount "SELECT ENTRY_META('tags') FROM #entries WHERE type='note'"
(empty)
```

Not stored under another name, either. `crates/rustledger-core/src/directive.rs`:

```rust
pub struct Note {
    pub date: NaiveDate,
    pub account: Account,
    pub comment: String,
    pub meta: Metadata,
}
```

Compare `Document`, which does carry them:

```rust
pub struct Document {
    pub date, pub account, pub path,
    pub tags: ...,
    pub links: ...,
    pub meta: Metadata,
}
```

## The parser already knows notes take them

`reject_tags_and_links` in `crates/rustledger-parser/src/cst/convert.rs` refuses to run for `note` and `document`, and says why:

> Not called from `note` or `document`: both take tags and links in beancount v3 **and we already agree with it there**.

That is half right and the half that is wrong is the one that matters. We agree in the sense that we do not *reject* them, which is what the surrounding code is about. We do not agree in the sense of keeping them: the tokens are parsed, not objected to, and then dropped on the floor because `Note` has nowhere to put them.

So the accepting side landed and the storing side did not, and the comment reads as though both had.

## Why this is worse than a missing column

The related gaps in #2154 are query-surface problems: the data exists and a table does not expose it. This is data loss at parse. A user tagging notes for later retrieval gets a clean exit code and no indication that the tags went nowhere, and no query can recover them because they were never stored.

It also fails silently in the direction that hides it. Rejecting the syntax would at least be loud.

## Scope

- `Note` needs `tags` and `links`, populated by `convert_note`.
- Once stored, `#notes` can expose the `tags` and `links` columns bean-query has and we lack (part of #2154).
[...]

PR #2189, opened 2026-08-30
fix(parser): a blank line no longer eats a directive's tags

robcohen:
## The bug

`convert_note` and `convert_document` collected a directive's `#tag` / `^link` tokens by walking the node's direct tokens and breaking at the first `NEWLINE`. A directive node begins with the trivia that precedes it, so when a **blank line sits above the directive**, its first token is a `NEWLINE` — the loop broke immediately and every tag and link was dropped.

That is the common case, not a corner. A blank line between directives is how ledgers are normally written. Of the three tagged `document` directives in the compatibility corpus, **two lost their tags** and one kept them — the one written directly under another directive.

`document` has carried this since long before `note` had any tags to lose (#2160). Nothing caught it because every fixture placed the directive immediately after another one.

## Fix

Both call sites now share `header_tags_and_links`, which starts collecting only once the header's `STRING` has been seen and breaks on the newline *after* it.

## Cache versions

Parser output changed, so both cache versions move: **loader 31 → 32**, **wasm 15 → 16**.

The wasm bump is overdue. #2160 moved the loader version and the pin but left `CACHE_VERSION` at 15, so **v15 is now ambiguous** — a blob written before #2160 holds the two-field-shorter `ArchivedNote`, one written by a dev build after it holds the new shape, and both claim v15. Rejecting every v15 blob is the only reading that is safe for either.

## The guard that should have caught it

`LOADER_CACHE_VERSION_PIN` can only force a *question*; it cannot distinguish a loader-internal change from one that moves our archived directives. On #2160 the question got the wrong answer and every test still passed.

`directive_archived_form_is_pinned` can tell. It parses one of every directive kind and hashes the archived bytes, so a **layout** change (new/reordered field) and an **output** change (different value) both fail it, with a message naming the bump to make.

### Verification

Both negative controls fire:

| mutation | kind | result |
|---|---|---|
| revert this parser fix | output | **FAILS** |
| swap `Note`'s `tags`/`links` declaration order | layout, values identical | **FAILS** |

The first control initially **passed** — the fixture's note and document followed other directives directly, so it never exercised the blank-line path. The blank lines in that fixture are now load-bearing and documented as such. The completion test made the identical mistake and is fixed the same way.

- Baseline drift is exactly the 2 corpus files containing tagged documents. Probed before/after: old code kept tags on 1 of 3 tagged documents, new code keeps 3 of 3.
- `corpus_parity` divergences: 20, unchanged from main (pre-existing).
- clippy 1.97 workspace clean; fmt, typos, `cargo doc` clean.


robcohen:
Pushed a second commit: `processed_ledger_archived_form_is_pinned`.

The parse-side pin covers `ParsedLedgerPayload`. `LedgerPayload` holds something else — the output of the shared `process()` pipeline, sorted by `canonical_sort_key` and booked. A `DirectivePriority` change reorders that list without touching a single parsed directive, so the parse-side digest does not move while every cached blob goes stale.

Not hypothetical: #2187 swaps `Pad` and `Balance` priorities. Applying that swap fails the new test and passes the parse-side one, which is the control for it. #2187 needs a wasm `CACHE_VERSION` bump it does not currently have; I will add it there.

Issue #2191, opened 2026-08-30
rledger format turns a balance with arithmetic + tolerance into a file that no longer parses


robcohen:
`rledger format` can produce output that fails to load. Formatting is supposed to be meaning-preserving; here it corrupts the file.

```
$ cat b.bean
2024-01-15 balance Assets:Cash 0.25 + 0.75 USD ~ 0.01 USD

$ rledger format b.bean
2024-01-15 balance Assets:Cash 0.25 + 0.75 ~ 0.01 USD

$ rledger format b.bean > b2.bean && rledger check b2.bean
P0012
  x parse error: malformed amount: expected one number, optionally signed, or
    an arithmetic expression.
   ,-[b2.bean:1:32]
```

The fixture is already in the formatter's own `IDEMPOTENCE_MATRIX`
(`balance_with_arithmetic_and_tolerance`). Idempotence holds — the broken output
is stable — so the existing test cannot see it.

## Root cause: the parser, not the formatter

The tolerance grammar accepts a currency-less left side only when it is a plain
number, not when it is an arithmetic expression:

| source | parses |
|---|---|
| `100.00 USD` | yes |
| `100.00 ~ 0.01 USD` | yes |
| `100.00 USD ~ 0.01 USD` | yes |
| `0.25 + 0.75 USD` | yes |
| `0.25 + 0.75 USD ~ 0.01 USD` | yes |
| **`0.25 + 0.75 ~ 0.01 USD`** | **no** |
| **`(0.25 + 0.75) ~ 0.01 USD`** | **no** |

`emit_balance` deliberately emits the single-trailing-currency form
(`ACCOUNT AMOUNT [~ TOL] CURRENCY`), which is the beancount form and matches
what the parser accepts for a plain number. So the formatter is right and the
parser is inconsistent: `~ TOL CURRENCY` should be accepted after an arithmetic
expression exactly as it is after a number.

Fixing the parser fixes the round-trip; no formatter change needed.

Found by a format-fidelity harness (parse -> format -> parse, compare models)
written while fixing #2184. That harness is blocked on this, so it will land
with the fix.

PR #2192, opened 2026-08-30
fix(format): stop deleting a note's tags and links

### LINKED ISSUE #2184: rledger format silently drops a note's tags and links; bean-format keeps them
Found while fixing #2160. Pre-existing — no formatter logic changed there —
but that fix is what makes this fixable, so filing now.

`rledger format` drops the tags and links from a `note` header. It keeps a
`document`'s:

```beancount
2024-01-05 note Assets:A "note text" #ntag ^nlink
2024-01-06 document Assets:A "/tmp/x.pdf" #dtag ^dlink
```

```
$ rledger format f.bean
2024-01-05 note Assets:A "note text"                      <-- tags gone
2024-01-06 document Assets:A "/tmp/x.pdf" #dtag ^dlink    <-- kept

$ bean-format f.bean
2024-01-05 note Assets:A "note text" #ntag ^nlink         <-- both kept
2024-01-06 document Assets:A "/tmp/x.pdf" #dtag ^dlink
```

Formatting a file should not change what it means, and this silently deletes
user data.

## Why it could not be fixed before

`Note` had no fields to hold tags or links — the parser accepted them and threw
them away (#2160) — so no formatter could have emitted them. Now that the data
survives parsing, this is a matter of rendering it.

## Where

`format_note_lines` in `crates/rustledger-core/src/format/directives.rs`
emits only the date, account and comment. `format_document_lines` next to it
does the same, yet a document's tags DO survive `rledger format` — so the CLI
appears not to use the model formatter for that path. Worth establishing which
formatter actually runs before changing either, rather than fixing the one that
is easier to find.

The round-trip scripts (`format-idempotence.sh`,
`format-preserves-comments.sh`) do not cover this: the output is stable and
comments are preserved. It is only the tags that vanish, and they vanish on the
first pass.

robcohen:
Closes #2184.

## What was happening

```
$ cat f.bean
2024-01-05 note Assets:A "note text" #ntag ^nlink
2024-01-06 document Assets:A "/tmp/x.pdf" #dtag ^dlink

$ rledger format f.bean
2024-01-05 note Assets:A "note text"                      <- tags deleted
2024-01-06 document Assets:A "/tmp/x.pdf" #dtag ^dlink    <- kept
```

Formatting is supposed to preserve meaning. This silently deleted user data on every run.

## Which formatter

The issue points at `format_note_lines` in `rustledger-core`. That is not the code that runs — `rledger format` calls `rustledger_parser::format::format_source`, the **CST** formatter. That is also why a document's tags survived: `emit_document` there carried a direct-child token walk re-emitting the header's TAG / LINK tokens, and `emit_note` carried nothing.

## Fix

Both now share `emit_header_tags_and_links` — the document walk unchanged, including its two hard-won details: skip **leading trivia** (a blank line before a non-first directive attaches its NEWLINE inside the node) and skip **leading comment lines** for the same reason. The converter had the identical bug from the identical cause (#2189).

## Why nothing caught it

The formatter's own `IDEMPOTENCE_MATRIX` asserts that formatting twice equals formatting once. Dropping a note's tags is **perfectly idempotent** — the property cannot see data loss. Stability is not fidelity.

The new test asserts fidelity: parse → format → parse → compare models.

**That alone is still not enough**, and not hypothetically. Run it on a tree whose *converter* also drops these tags (#2189, unmerged) and both sides come back tagless and **equal**, while every tag in the file has been deleted — the assertion passes while the bug is live. It therefore also asserts on the formatter's **output text**, which is the thing under change and cannot be satisfied by a matching pair of empty models.

## Verification

- Reverting the one `emit_note` line fails the new test (control).
- Fixture covers a note after a blank line, after a comment line, with metadata, and a document — the trivia cases that broke the equivalent walk before.
- `format-idempotence.sh`: `checked=693 skipped=42 non_idempotent=0 diagnostics_changed=0`
- `format-preserves-comments.sh`: `checked=693 dropped_comments_in=0`
- `corpus_parity` divergences: 20, unchanged from main (pre-existing).
- clippy 1.97 workspace clean; fmt, typos, `cargo doc` clean.

## Found on the way: #2191

The same fidelity check over the **whole** matrix fails on `balance_with_arithmetic_and_tolerance` — the formatter emits `0.25 + 0.75 ~ 0.01 USD`, which no longer parses. That is a parser inconsistency (it accepts `~ TOL CURRENCY` after a number but not after an expression), filed as #2191. The matrix-wide harness lands with that fix rather than behind an allow-list here.

### DIFF
--- a/crates/rustledger-core/src/format/directives.rs
+++ b/crates/rustledger-core/src/format/directives.rs
@@ -148,26 +148,48 @@ pub fn format_query_lines(query: &Query, config: &FormatConfig) -> Vec<FormatLin
     lines
 }
 
+/// Append a header's `#tag` / `^link` list, in the order the transaction
+/// emitter uses.
+///
+/// `note` and `document` take these exactly as a transaction does. Both
+/// emitters here omitted them, so any consumer of the typed-directive
+/// emitter -- `rledger add`, `rledger extract`, the FFI `format.entry`
+/// endpoints, all of which reach it through
+/// `rustledger_parser::format::canonicalize_directives` -- deleted them
+/// from the text it produced (#2184).
+fn write_tags_and_links(header: &mut String, tags: &[crate::Tag], links: &[crate::Link]) {
+    for tag in tags {
+        write!(header, " #{tag}").expect("write to String is infallible");
+    }
+    for link in links {
+        write!(header, " ^{link}").expect("write to String is infallible");
+    }
+}
+
 /// Render a note directive into format lines.
 pub fn format_note_lines(note: &Note, config: &FormatConfig) -> Vec<FormatLine> {
-    let mut lines = vec![FormatLine::Plain(format!(
+    let mut header = format!(
         "{} note {} \"{}\"",
         note.date,
         note.account,
         escape_string(&note.comment)
-    ))];
+    );
+    write_tags_and_links(&mut header, &note.tags, &note.links);
+    let mut lines = vec![FormatLine::Plain(header)];
     metadata_lines(&note.meta, &config.indent, config, &mut lines);
     lines
 }
 
 /// Render a document directive into format lines.
 pub fn format_document_lines(doc: &Document, config: &FormatConfig) -> Vec<FormatLine> {
-    let mut lines = vec![FormatLine::Plain(format!(
+    let mut header = format!(
         "{} document {} \"{}\"",
         doc.date,
         doc.account,
         escape_string(&doc.path)
-    ))];
+    );
+    write_tags_and_links(&mut header, &doc.tags, &doc.links);
+    let mut lines = vec![FormatLine::Plain(header)];
     metadata_lines(&doc.meta, &config.indent, config, &mut lines);
     lines
 }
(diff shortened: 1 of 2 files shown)


robcohen:
Good catch — applied in e0111a36ea.

You're right that the name promised a full walk and the code did not deliver one. Harmless for the two callers it has (a note's or document's only child node is a `META_ENTRY`, which sits past the header newline the walk has already broken on), but the failure mode if a header ever gains a child node is that every tag after it disappears silently — and this helper exists precisely because that kind of silent drop shipped twice already.

A `Node` now marks `seen_content` and continues. Behavior is unchanged today: full parser suite green, and 693 corpus files still report `non_idempotent=0` / `dropped_comments_in=0`.

robcohen:
Correction to the verification notes on this PR.

I reported `corpus_parity` divergences as "20, unchanged from main (pre-existing)". That measurement was wrong. The test loads a prebuilt `target/wasm32-wasip2/debug/rustledger_ffi_component.wasm`, and **no test command rebuilds it** — the copy on my disk was three weeks old, so it was comparing today's native code against a stale component.

After `cargo build -p rustledger-ffi-component --target wasm32-wasip2`:

```
cargo test -p rustledger-ffi-component-tests   ->  ok, 0 divergences
cargo test --workspace                          ->  green
```

So there is no known-failing test here at all; the actual state is better than I described. The tell I missed was the number's *stability* — it stayed 20 because the component never changed, not because the failure was inherent. Every other verification figure I quoted (clippy, fmt, typos, the 693-file format scripts, the baselines) was measured directly and stands.

Issue #2194, opened 2026-08-30
`query` accepts tags and links and silently drops them


robcohen:
```beancount
2024-01-01 query "n" "SELECT date" #qtag ^qlink
```

Accepted with no diagnostic. `Query` has no `tags` or `links` field, so both are discarded.

## It is in neither camp

Seven directives call `reject_tags_and_links` — `open`, `close`, `commodity`, `event`, `price`, `balance`, `pad`. Three keep them — `transaction`, `note`, `document` (the latter two since #2160/#2185).

`query` does neither. And the message the refusal arm emits states the rule its own omission breaks:

> tags and links belong to transactions, and to note and document directives

This is the same shape as #2160 (`note` accepted tags and threw them away) and #2184 (the formatters deleted them), so it is worth closing rather than leaving as the last member of that family.

## Which way is not obvious

- **Refuse**, like the other seven. beancount's `Query` namedtuple is `(meta, date, name, query_string)` — no tags field — which suggests its grammar has no `tags_links` on `query` either.
- **Keep**, like `note` now does. Requires adding the fields to `Query`, threading them through the visitors, and a cache-version bump.

The evidence for "refuse" is beancount's data model, which I could not confirm against `bean-check` locally (the compat container is not built here). And a real corpus file uses the form:

```
tests/compatibility/files/parser-lima/examples_data_full.beancount:111
2028-01-01 query "fictional SQL" "SELECT money FROM accounts;" #for-testing
```

That file is a *parser-lima* example rather than a beancount test, so it is weak evidence that beancount accepts it — but it does mean the form appears in the wild and a refusal is a user-visible strictness increase. **Run `bean-check` on that line before choosing.**

## Pinned meanwhile

`every_directive_either_keeps_tags_or_refuses_them` in `parser_integration_test.rs` is a census over all eleven directives. It asserts the current accept-and-drop behavior for `query` explicitly, so whichever way this goes, that test fails and forces the census to be updated rather than going stale.

I implemented the refusal, measured it, and backed it out: it makes that corpus file error, and bundling an unverifiable strictness increase into an unrelated PR is the wrong place to decide it.

PR #2195, opened 2026-08-30
fix(parser): a balance tolerance may follow an arithmetic amount

### LINKED ISSUE #2191: rledger format turns a balance with arithmetic + tolerance into a file that no longer parses

robcohen:
Closes #2191.

## The bug

`rledger format` could produce a file that no longer loads:

```
$ cat b.bean
2024-01-15 balance Assets:Cash 0.25 + 0.75 USD ~ 0.01 USD
$ rledger format b.bean
2024-01-15 balance Assets:Cash 0.25 + 0.75 ~ 0.01 USD     <- does not parse
```

Root cause is the **parser**, not the formatter. `directive_arithmetic_value` scanned to the first top-level `CURRENCY` and treated everything before it as the amount expression. With no currency between the amount and the tolerance, it swallowed the tolerance and handed `0.25 + 0.75 ~ 0.01` to the expression parser, which is not arithmetic — so the whole value was reported malformed.

Only the currency-less arithmetic form reached that state, which is why `100.00 ~ 0.01 USD` and `0.25 + 0.75 USD ~ 0.01 USD` both worked. `~` now ends the expression, which is what it means.

## A regression this introduced, caught in review

`directive_arithmetic_value` is shared with `price`, and a price has no tolerance. Stopping at `~` for both turned `price USD 1.10 + 0.05 ~ 0.01 EUR` from a diagnosed error into a silent 1.15 with the tolerance discarded. `ToleranceForm` now says which grammar is being parsed.

That fix left the plain form, `price USD 1.10 ~ 0.01 EUR`, which **main already accepted the same silent way** — `~` is not an arithmetic operator, so the scan took the fast path and fell back to the first NUMBER. Fixing only the half I broke would leave the identical silent truncation one keystroke away, so a top-level `~` on a price is now diagnosed.

## Verification

Values, not just parse success. An earlier bug in this same scan fell back to "take the first NUMBER" and turned `(1 + 5) / 2.1` into a silent assertion against **1** — parsed cleanly, asserted the wrong thing. So every accepted form is checked against the amount it must evaluate to, including a ledger holding `0.25` confirming that `0.25 + 0.75` does not match it:

| form | parses | evaluates to |
|---|---|---|
| `100.00 USD` | yes | 100.00 |
| `100.00 ~ 0.01 USD` | yes | 100.00 |
| `100.00 USD ~ 0.01 USD` | yes | 100.00 |
| `0.25 + 0.75 USD` | yes | 1.00 |
| **`0.25 + 0.75 ~ 0.01 USD`** | **now yes** | 1.00 |
| `0.25 + 0.75 USD ~ 0.01 USD` | yes | 1.00 |
| **`(0.25 + 0.75) ~ 0.01 USD`** | **now yes** | 1.00 |

Price: both tolerance forms diagnosed; the three tolerance-free forms still parse to the right values, because a rejection that also rejected valid input would pass a test asserting only rejection.

## Also here

- **The matrix-wide format-fidelity harness** that #2192 could not land — it fails on `balance_with_arithmetic_and_tolerance` without this fix, which is how the bug was found. Its floor is raised from `> 20` to `>= 35` against a measured 37 of 42, so ten fixtures could not start failing to parse unnoticed.
- **A census of which directives keep tags and which refuse them.** Seven refuse, three keep, each refusal checked against its tag-free form too. `query` is in neither set — it accepts a tag and drops it — pinned here as *current* behavior with #2194 filed to decide it.
- **Cache versions**: loader 32 → 33, wasm 17 → 18, pin 33. Output moves both ways: a stale blob holds an error where this build makes a balance, and a truncated 1.10 where this build makes an error.

Workspace green, clippy 1.97 clean, fmt/typos clean, parser baseline unchanged, `corpus_parity` 0 divergences against a freshly built wasip2 component.


robcohen:
Fixed — the assertion now matches `ParseErrorKind::SyntaxError(m)` and searches `m`, rather than stringifying the whole `ParseError` with `{e:?}` and searching that.

You're right that the Debug form drags in span and context, which the test does not care about. Worth noting it still pins the wording: rewording the message fails the test, verified by changing it. So this narrows what the assertion is coupled to rather than weakening it.

Issue #2186, opened 2026-08-30
tests/baselines/parser-corpus.manifest is stale on main; drift accumulates until a parser PR carries it


robcohen:
Found while regenerating for #2160/#2185. Independent of that change.

`tests/baselines/parser-corpus.manifest` does not match what the current
`main` produces. Running `scripts/regen-corpus-baselines.sh` on a clean `main`
checkout yields:

- **3 modified hashes** — `apyb-financeiro/financeiro_main.beancount`,
  `donearm-ledger/accounts.beancount`, `fava/tests_data_example.beancount`
- **20 new entries** — corpus files with no baseline at all, presumably added
  by `fetch-compat-test-files.sh` after the manifest was last regenerated

## Why it went unnoticed

The `Corpus output baseline` job is path-filtered: it runs when the parser or
formatter changes. Nothing had touched those paths in a while, so the drift
accumulated silently and only surfaced on the next parser change — which then
has to carry it.

That is the problem worth fixing. The gate is sound, but a PR that legitimately
shifts parser output cannot regenerate *only* its own files: the script
regenerates the whole manifest, so unrelated drift rides along in the diff and
has to be explained by whoever happens to be next.

## Suggestion

Either regenerate on a cadence that does not depend on someone touching the
parser, or have the job report drift (without failing) on every PR so it cannot
accumulate. The 20 missing entries in particular mean those corpus files have
had no baseline coverage at all since they were added.

## Note

#2185 regenerates the manifest and therefore includes this drift. Its
description separates the two: 10 modified hashes are from that change (all
files containing `note` directives), the other 3 and all 20 additions are this.

robcohen:
Closing — both halves are resolved and verified on main today.

**Part 1, the stale manifest.** Gone. Against a freshly fetched corpus on current main:

```
corpus files: 735   manifest entries: 735
files with no baseline: 0   orphaned entries: 0
regeneration: no diff
```

The 3 modified hashes and 20 missing entries this issue listed were carried by #2185 and #2189.

**Part 2, the accumulation, which was the real problem.** Addressed by #2196.

One correction to this issue's diagnosis: it attributes the drift to a path filter on the baseline gate. There isn't one — `parser-baselines.yml` has no `paths:` key and runs on every PR. The actual mechanism is two deliberate decisions meeting:

1. An unmanifested downloaded file is a **warning**, not a failure — correct, since upstream can add a file at any time and failing would redden unrelated PRs. But the warning goes to a log nobody reads.
[...]

PR #2196, opened 2026-08-30
ci: report corpus baseline drift on a cadence instead of by accident

robcohen:
Refs #2186.

## Part 1 is already done

The manifest is **current**. On a clean checkout with a freshly fetched corpus:

- `./scripts/fetch-compat-test-files.sh` → 735 files
- `./scripts/regen-corpus-baselines.sh` → **zero diff**
- discovered files 735, manifest entries 735, **0 missing, 0 orphaned**

#2185/#2189 carried the 3 modified hashes and 20 additions the issue lists. Nothing left to regenerate.

## The stated cause isn't the cause

The issue says the gate is path-filtered. It isn't — `parser-baselines.yml` has no `paths:` filter and its header says it runs on every PR. The real mechanism is two deliberate decisions meeting:

1. **An unmanifested downloaded file is a warning, not a failure.** That's correct as written: upstream can add a file at any moment, and failing would turn unrelated PRs red. But the warning goes to a CI log nobody reads, so the file sits with no parser-output coverage at all.

2. **CI's corpus is frozen.** The job restores `tests/compatibility/files` from a cache keyed on `hashFiles('scripts/fetch-compat-test-files.sh')`, and the fetch step is skipped on an exact key hit. The key changes when the *script* changes — never when *upstream* does. So CI keeps replaying whatever corpus existed when that key was created, and every PR agrees there's no drift.

Neither is wrong alone, and tightening either costs exactly what its own comment explains. What's missing is a third thing that looks at a corpus nobody has frozen.

## What this adds

A weekly job that fetches the corpus **fresh — no cache, which is the entire point** — regenerates, and turns any drift into an issue. It reports; it never gates; it closes its own issue when the manifest matches again.

The issue body separates the three cases, because they need different responses:

| group | meaning |
|---|---|
| **New corpus files with no baseline** | no parser-output coverage at all since they appeared |
| **Changed hashes** | source hash moved = upstream churn; output hash moved with source unchanged = a parser change reached main without regenerating |
| **Entries whose file is gone** | upstream deleted or renamed; dead weight |

**The oracle is the repository's own `BASELINE_UPDATE=1` path plus `git diff`** — not a Python reimplementation of the fingerprint, which would be a second implementation of the thing under test, free to disagree with the Rust one in precisely the cases that matter.

## Self-test

Modeled on `nightly-health.yml`, including the self-test that runs **before** the live check and **is** allowed to fail the job. The live check exits 0 whatever it finds, so a broken classifier would otherwise report a permanently reassuring "no drift" — the same silence this workflow exists to break, wearing a green check.

Verified it can fail, two ways:

| break | caught |
|---|---|
| rehash detection disabled (`rehashed = []`) | 3 failures |
| classifier returns clean for everything | 5 failures |

It also pins that an empty diff reports clean and that comment-only lines aren't drift, so it can report *both* answers rather than only the alarming one.

### DIFF
--- a/.github/workflows/corpus-baseline-drift.yml
+++ b/.github/workflows/corpus-baseline-drift.yml
@@ -1,0 +1,78 @@
+# Report parser-baseline drift that no pull request would otherwise carry.
+#
+# The baseline gate in `parser-baselines.yml` runs on every PR, but it cannot
+# catch this class, for two reasons that are both deliberate:
+#
+#   * A downloaded corpus file with no manifest entry is a WARNING there, not
+#     a failure. Upstream can add a file at any moment and failing would fire
+#     on unrelated PRs. The warning goes to a log nobody reads, so such a file
+#     sits with no parser-output coverage at all.
+#
+#   * That job restores `tests/compatibility/files` from a cache keyed on the
+#     FETCH SCRIPT, and skips the fetch on an exact key hit. CI's corpus is
+#     therefore frozen at whatever was fetched when that key was created,
+#     while the upstream repositories it mirrors keep moving. Every PR sees
+#     the same stale corpus and agrees there is no drift.
+#
+# So drift accumulated until somebody changed the parser, whose PR then had to
+# carry and explain churn that had nothing to do with it (#2186).
+#
+# This job exists to break that silence: it fetches the corpus FRESH -- no
+# cache, which is the whole point -- and turns any drift into an issue. It
+# reports; it does not gate, and it never fails on drift. Same shape as
+# `nightly-health.yml`, for the same reason: a result nobody sees is not a
+# result.
+
+name: Corpus Baseline Drift
+
+on:
+  schedule:
+    # Mondays 05:00 UTC. Weekly is enough for upstream corpus churn, and it
+    # avoids the nightly band (kani 06:00 Sun, nightly-health 08:00 daily).
+    - cron: '0 5 * * 1'
+  workflow_dispatch:
+
+concurrency:
+  group: corpus-baseline-drift
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  issues: write
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  drift:
+    name: Corpus baseline drift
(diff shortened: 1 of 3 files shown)


robcohen:
Fixed in the latest commit — both flagged spots now build the string explicitly.

The lead paragraph moved into a `_summary()` helper and the trailer uses a parenthesised single expression, so no list element is two adjacent literals any more. You're right that it's not pedantic here: in a list of issue-body paragraphs, a missing comma is a plausible typo rather than a theoretical one, and it would silently merge two paragraphs instead of erroring.

Same commit fixes something the review prompted me to go looking for: `gh()` returned `""` on failure, which is indistinguishable from a successful empty result. A failed `issue list` would have read as "no existing issue" and opened a duplicate weekly, and a failed `issue create` would have printed a blank URL and left the run green having reported nothing — the exact silence this workflow exists to break. It now raises and the step exits non-zero, self-tested in both directions.

robcohen:
All three addressed.

**Implicit string concatenation** — fixed in cdbfd0daa4, nine runs now join with an explicit `+`. Wrapping in parentheses wasn't enough; adjacency inside parentheses is still adjacency. Worth doing here specifically because these are paragraphs of an issue body, where a missing comma silently merges two of them rather than raising. Verified the rewrite changed nothing by rendering the same diff through the pre- and post-rewrite modules: byte-identical body, sha `e0194f7e3082a81a` on both sides.

**Empty except** — same commit. Reaching that arm *is* the assertion (the `failures.append` above it is what runs otherwise), but nothing said so, and a bare `pass` under an `except` is worth explaining every time.

**`gh()` swallowing non-zero exits** — you're right, and it was already fixed two commits before your comment landed, in `fix(ci): a gh failure must not read as "nothing to report"`. Your reasoning matches exactly what I'd written there: a failed `issue list` reads as "no existing issue" and opens a duplicate weekly, and a failed `issue create` prints a blank URL and leaves the run green having reported nothing — the exact silence this job exists to break. It now raises `GhError` and the step exits non-zero, self-tested in both directions (returning `""` on failure fails the test; so does raising on a successful empty result, which is a real and different answer).

PR #2197, opened 2026-08-31
docs(core): cross-file booking order is not Python's, on purpose

### LINKED ISSUE #2149: Same-date directives from different included files book in a different order than beancount, changing realized gains

robcohen:
Closes #2149.

## The wrong claim

`sort_directives` said its stable sort matched Python's `(date, type_priority, lineno)`. That holds inside one file, where parse order and line number rise together. Across `include`s it does not — and `booking_sort_key`, the doc comment that calls itself the single source of the booking order, repeated it.

Python compares line numbers taken from **different files**, so a directive on line 1 of a file included second sorts ahead of one on line 5 of a file included first. This sort keeps include order.

## Measured against real beancount

Not taken from the issue — reproduced against **beancount 3.2.3**. One buy at 10.00 on line 5 of the first included file, one at 20.00 on line 1 of the second, then a FIFO sale of one unit:

| include order | beancount | rustledger |
|---|---|---|
| first, second | -10.00 | **-20.00** |
| second, first | -10.00 | -10.00 |

Swapping the includes pins each rule down: beancount is invariant because it re-sorts by line number; ours tracks how the ledger was assembled. A 2x difference in reported gain, decided by which file a buy lives in and what line it happens to sit on, with no error or warning either side.

## Which rule stands

Include order, deliberately. Ordering two directives by comparing a line number from one file against a line number from a different file is not a fact about the ledger: **adding a comment to one file silently changes which lot a sale in another file consumes.** Include order at least reflects how the author assembled the ledger.

Beancount's rule is not purely line-number driven either — directives sharing a line number across files fall back to include order through its own stable sort. So this is not "match upstream or not", it is choosing between two tiebreaks neither of which is a fact.

## Why nothing caught it

- The three tests that exercise this ordering (`test_sort_directives_by_date`, `test_sort_directives_by_type_same_date`, `test_sort_directives_pad_before_balance`) do not mention a file or an `include`.
- The compatibility corpus **cannot** reach it: of 761 `.beancount` files, one contains a single `include` and none contains two. The divergence needs two included files with same-date directives at differing line numbers, so no corpus file can express the shape. This is not a gap the corpus happens to miss — it is one it cannot currently express, which is why the fixture ships in-tree.

## What this changes

No behavior change. Both doc comments corrected, an in-tree fixture, a test pinning the rule, and an entry in `docs/reference/compatibility.md` alongside the other deliberate divergences.

**The test is verified to fail**: dropping `file_id` from `canonical_sort_key` so the span offset decides across files produces `-10.00` — beancount's answer exactly, not merely a different one. So the control lands on the real alternative rather than on noise.

Workspace green, clippy 1.97 clean, fmt/typos/rustdoc clean, `corpus_parity` clean against a freshly built wasip2 component.

### DIFF
--- a/crates/rustledger-core/src/directive.rs
+++ b/crates/rustledger-core/src/directive.rs
@@ -648,9 +648,35 @@ impl Directive {
 
 /// Sort directives by date, then type priority.
 ///
-/// This is a stable sort, so directives sharing a date and type keep their
-/// file order — which is the order they are booked in, matching Python's
-/// `(date, type_priority, lineno)`.
+/// This is a stable sort, so directives sharing a date and type keep the
+/// order they were parsed in, which is the order they are booked in.
+///
+/// **Within one file that matches Python's `(date, type_priority, lineno)`.
+/// Across `include`s it deliberately does not**, and the difference is
+/// observable in reported gains rather than merely cosmetic.
+///
+/// Python compares `lineno` values taken from different files, so a directive
+/// on line 1 of a file included second sorts ahead of one on line 5 of a file
+/// included first. This sort keeps include order: everything from the first
+/// included file, in its own order, then the second.
+///
+/// Two same-date lots therefore enter the inventory in different orders under
+/// the two rules, and a FIFO sale whose lot-date comparison ties falls through
+/// to that order. Measured against beancount 3.2.3 on the fixture in
+/// `tests/fixtures/cross-file-order/` — one buy at 10.00 on line 5 of the
+/// first included file, one at 20.00 on line 1 of the second — selling one
+/// unit reports a 2x difference in realized gain, with no error either side.
+///
+/// Include order is kept on purpose. Ordering two directives by comparing a
+/// line number from one file against a line number from a different file is
+/// not a fact about the ledger: adding a comment to one file silently changes
+/// which lot a sale in another file consumes. Include order at least reflects
+/// how the author assembled the ledger. Beancount's own rule is not purely
+/// `lineno` driven either — directives sharing a line number across files
+/// fall back to include order through its stable sort.
+///
+/// Pinned by `cross_file_same_date_directives_keep_include_order` and recorded
+/// in `docs/reference/compatibility.md` (#2149).
 pub fn sort_directives(directives: &mut [Directive]) {
     directives.sort_by_cached_key(booking_sort_key);
 }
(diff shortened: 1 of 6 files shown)


robcohen:
Fixed in f797b1fd7e — "nor" in both places.

Issue #2193, opened 2026-08-30
A balance tolerance silently ignores a second number: `~ 0.001 0.02 USD` uses 0.001


robcohen:
```beancount
2024-01-15 balance Assets:C 1.00 ~ 0.001 0.02 USD
```

Accepted with no diagnostic. The tolerance used is **0.001**; the `0.02` the author wrote is discarded silently. Verified by asserting against an account off by 0.007: it fails, as it would with `~ 0.001` alone, and passes with `~ 0.01`.

Pre-existing — confirmed by toggling the `~` handling added in #2191 and seeing this form accepted both ways, while the form that fix targets flips from error to accepted. So it is not a consequence of that change.

## Why it slips through

`malformed_directive_value` exists precisely to refuse a value carrying more than one NUMBER, but it only runs in the `or_else` fallback after `directive_arithmetic_value` returns `None`. Here the amount resolves fine, so nothing ever inspects the tolerance for a stray second number.

For contrast, the tolerance DOES evaluate arithmetic correctly — `~ 0.005 + 0.005 USD` behaves as 0.01, confirmed against an account off by 0.007. So the reader is not simply taking the first number in general; it is juxtaposition specifically that goes unnoticed.

## Same family as

- #2190 — `custom` emits a value it has just errored on
- #2191 — a price silently discarded a tolerance (fixed there)

All three are "the parser read part of what the author wrote and dropped the rest without saying so". A value that cannot be honored should be diagnosed, not quietly narrowed.

robcohen:
Sibling finding, same root cause: **nothing validates the tolerance clause at all.**

A tolerance currency that differs from the amount's is accepted and ignored:

```
2024-01-15 balance Assets:C 1.00 USD ~ 0.01 EUR    -> accepted, EUR ignored, USD used
2024-01-15 balance Assets:C 1.00 EUR ~ 0.01 USD    -> accepted, USD ignored, EUR used
```

The first currency wins; the trailing one is discarded without a word. Confirmed pre-existing by toggling the `~` handling from #2191 — this form is accepted both ways, while the form that fix targets flips from error to accepted, so the control is real.

## It reaches the file

```
$ cat m.bean
2024-01-15 balance Assets:C 1.00 USD ~ 0.01 EUR
$ rledger format m.bean
2024-01-15 balance Assets:C 1.00 ~ 0.01 USD        <- the EUR is gone from the file
```

[...]

Issue #2190, opened 2026-08-30
MetaValue::Tag / MetaValue::Link are unreachable from any accepted input


robcohen:
`MetaValue` has `Tag` and `Link` variants. Nothing the parser **accepts** produces either one.

## Evidence

Metadata values are rejected everywhere — note and transaction alike:

```beancount
2024-01-20 note Assets:A "n"
  t: #leaked          ; SyntaxError("unexpected input")

2024-01-20 * "x"
  t: #tv              ; SyntaxError("unexpected input")
  t: ^lv              ; SyntaxError("unexpected input")
```

`custom` is the only producer, and it produces the value *while rejecting the input*:

```
2024-01-05 custom "budget" #atag
  -> errors: ["a tag (#atag) is not a valid custom value"]
  -> Custom.values: [Tag(Tag(InternedStr("atag")))]
```

So a consumer that reads `values` without checking `errors` sees a value the parser has declared invalid. That is the divergence worth fixing regardless of which way the grammar question goes.

## Consequences

- `visit_tags` documents that it collects "`MetaValue::Tag` in any directive's or posting's metadata" (`crates/rustledger-core/src/visit.rs:185`). That branch is unreachable: no accepted input puts a `MetaValue::Tag` in metadata.
- `meta_value_to_json` (`rustledger-wasm/src/convert.rs:25`) and the FFI `MetaValue::Text` mapping (`rustledger-ffi-component/src/convert.rs:71`) are likewise unreachable for these two variants, and are covered only by tests that construct the variant by hand.

## Questions this needs answered

1. Does beancount accept a tag or link as a **metadata** value? If yes, the parser is wrong to reject it and the variants are right. If no, the variants are modelling something that cannot exist and the visitor doc is claiming coverage it cannot have.
2. Whichever way (1) goes, `custom` should not emit a value it has just errored on.

Found while probing the collector added in #2189; not caused by it, and not in its scope.

PR #2198, opened 2026-08-31
fix(parser): a query refuses tags and links instead of dropping them

### LINKED ISSUE #2194: `query` accepts tags and links and silently drops them

robcohen:
Closes #2194.

## The gap

`query` was in neither camp. It did not keep tags and links the way `transaction`, `note` and `document` do, and it did not refuse them the way the other seven directives do — so it took them and discarded them, `Query` having no field to hold one. Meanwhile the message the refusal arm emits states the very rule its omission broke:

> tags and links belong to transactions, and to note and document directives

## The direction, now settled

I filed this rather than fixing it because deciding needed `bean-check` and the compat container would not build here. It does now — `nix shell nixpkgs#python3 nixpkgs#uv` plus `uv pip install beancount` — and **3.2.3 answers it**:

```
2024-01-01 query "n" "SELECT date" #qtag
  -> syntax error, unexpected TAG, expecting end of file or EOL
```

Its `Query` namedtuple is `(meta, date, name, query_string)` — nowhere to keep one either. So refusing matches beancount's grammar *and* its model.

## The reason I backed this out before is answered too

A corpus file uses the form, `parser-lima/examples_data_full.beancount:111`, and I could not tell whether that meant beancount accepted it. **It does not.** beancount reports TAG syntax errors on that file at lines 15, 19, 28, 31, 65, 68, 96 and **111**.

That file is a parser-lima corpus carrying forms *lima* accepts. Our accepting line 111 was the divergence — not our rejecting it. The single baseline hash that moves is that file.

## Verification

- Both sigils rejected; the tag-free form still parses.
- The census in `parser_integration_test` moves `query` out of its known-gap arm into the refusing set, where every entry is *also* checked against its tag-free form — a refusal that rejected valid input would satisfy an assertion that only looked for an error.
- **Control**: removing the refusal reports `query accepted #atag it has no field to hold`.
- Cache versions move because parser output does: **loader 33 → 34, wasm 18 → 19, pin 34**. A stale blob holds a `Query` directive for input this build refuses.
- Workspace green, clippy 1.97 clean, fmt/typos clean, `corpus_parity` clean against a freshly built wasip2 component.

## Remaining in this family

#2190 (`MetaValue::Tag`/`Link` unreachable; `custom` emits a value it errored on) and #2193 (tolerance ignores a second number; a mismatched tolerance currency is erased by `format`). Both now have a local beancount to settle direction against.


robcohen:
Good catch, and it was wrong in three places, not one.

You're right that `reject_tags_and_links` records an error and conversion still emits the `Query` — verified, and the same holds for `event` and `balance`, so all eight refusing directives behave that way. It's a diagnostic, not a rejection. The comments claiming "refused outright" and "produces no directive" are corrected in d0e65dfabb.

Your reading of the actual hazard is right too, and I measured it rather than just rewording. A file with a tagged query used to parse clean and was therefore cacheable, so a v33 blob exists for it; replaying that blob skips the parse that would now complain. Against a cache written by a pre-#2194 binary:

```
without the bump   0 parse errors   <- diagnostic silently hidden
with the bump      2 parse errors
```

So the bump is load-bearing rather than precautionary, and that is now what the comments say.

One thing your comment surfaced indirectly: `cargo test` caught my first attempt at the rewrite, because a blank line followed by an indented paragraph inside a doc comment reads as a Rust code block and rustdoc tried to compile `query "n" "SELECT date" #qtag` as source.

PR #2200, opened 2026-08-31
fix(parser): diagnose a balance tolerance that says two things at once

### LINKED ISSUE #2193: A balance tolerance silently ignores a second number: `~ 0.001 0.02 USD` uses 0.001

robcohen:
Closes #2193.

## The rule

Beancount's tolerance grammar is `NUMBER ~ NUMBER CURRENCY` — one currency, trailing — and it rejects any currency written before the `~`. This PR differs in **both** directions, on a single rule: **accept what has exactly one meaning, diagnose what has none or contradicts itself.**

| form | beancount 3.2.3 | before | after |
|---|---|---|---|
| `1.00 ~ 0.01 USD` | ok | ok | ok |
| `0.25 + 0.75 ~ 0.01 USD` | ok | ok | ok |
| `1.00 ~ 0.005 + 0.005 USD` | ok | ok | ok |
| `1.00 USD ~ 0.01 USD` | syntax error | ok | **ok** (kept) |
| `0.25 + 0.75 USD ~ 0.01 USD` | syntax error | ok | **ok** (kept) |
| `1.00 USD ~ 0.01 EUR` | syntax error | ok, EUR discarded | **diagnosed** |
| `1.00 ~ 0.001 0.02 USD` | syntax error | ok, `0.02` discarded | **diagnosed** |

**Laxer on purpose.** `1.00 USD ~ 0.01 USD` states the currency twice and agrees with itself. One reading, and `rledger format` canonicalizes it to `1.00 ~ 0.01 USD` losslessly — refusing a file whose meaning is not in doubt buys nothing.

**Stricter because the other two are not redundancy.** A tolerance denominated in a currency the amount does not use asserts something the model has no field for; two juxtaposed numbers have no reading at all. Both were accepted, read in part, and the remainder dropped in silence — and `rledger format` then wrote that loss back to the file:

```
$ rledger format m.bean          # before this PR
2024-01-15 balance Assets:C 1.00 ~ 0.01 USD     <- the EUR is gone from the file
```

## Arithmetic is untouched

`~ 0.005 + 0.005` and `~ 0.005 * 2` still evaluate. That is why the second check asks whether the tolerance region **fails to parse as an expression**, rather than merely counting numbers — counting would have caught the legitimate arithmetic forms too. Verified by value, not just by acceptance: against an account off by 0.007, `~ 0.005 + 0.005` passes and bare `~ 0.005` fails.

## Controls

| mutation | result |
|---|---|
| disable the currency-mismatch check | fails |
| disable the juxtaposed-number check | fails |
| over-strict: also reject the agreeing-currency form | **fails** |

The third is the one worth having. It stops "just match beancount" being applied later without noticing that it is a decision rather than a cleanup.

## Also

- Every row of the test table was run against **beancount 3.2.3** and records what it does, so the divergence is legible where someone would change it.
- `docs/reference/compatibility.md` gains a section, including the portability note: a file using the accepted-but-non-standard spelling loads here and is a syntax error in beancount and tools built on it.
- Cache versions move for the same reason as v34 — the archived directive is identical, and the diagnostic is what a stale blob would hide. **Loader 34 → 35, wasm 19 → 20, pin 35.**
- Workspace green, clippy 1.97 clean, fmt/typos/rustdoc clean, corpus baseline unchanged, `format-idempotence.sh` reports `checked=693 non_idempotent=0 diagnostics_changed=0`.

### DIFF
--- a/crates/rustledger-loader/src/cache.rs
+++ b/crates/rustledger-loader/src/cache.rs
@@ -489,12 +489,20 @@ const CACHE_MAGIC: &[u8; 8] = b"RLEDGER\0";
 ///     WITHOUT this bump and 2 WITH it. The bump is what surfaces the
 ///     diagnostic.
 ///
+/// v35: a balance tolerance whose currency disagrees with the amount's, or
+///     which carries a second juxtaposed number, is diagnosed instead of
+///     read in part and discarded in part (#2193). Like v34 the DIRECTIVE is
+///     unchanged and the error list is what moves — and like v34 that is
+///     enough, because such a file used to parse clean and was therefore
+///     cacheable, so replaying its v34 blob skips the parse that now
+///     complains.
+///
 /// Public so `rustledger-wasm` can pin its own cache version against this one.
 /// Both caches archive the same `Vec<Directive>`, so a parser change that
 /// alters PARSER OUTPUT has to bump both — and on #1942 only this one was
 /// bumped, which review caught rather than any test. See
 /// `loader_cache_version_is_pinned` in `rustledger-wasm/src/cache.rs`.
-pub const CACHE_VERSION: u32 = 34;
+pub const CACHE_VERSION: u32 = 35;
 
 /// Cache header stored at the start of cache files.
 #[derive(Debug, Clone)]
@@ -1248,7 +1256,9 @@ mod tests {
         // directives -- the tagged `query` is still emitted, alongside a new
         // diagnostic. Either way it is not how a `CostNumber` is archived, so
         // the byte arrays below still hold.
-        const FIXTURE_VERSION: u32 = 34;
+        // v35 (#2193) is the same shape for balance tolerances: new errors,
+        // same archived `CostNumber` encoding.
+        const FIXTURE_VERSION: u32 = 35;
         assert_eq!(
             CACHE_VERSION, FIXTURE_VERSION,
             "CACHE_VERSION advanced past the fixture version; regenerate \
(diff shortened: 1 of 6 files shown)


robcohen:
Applied in the latest commit, and the point led somewhere.

**The two Vecs are gone** — you're right that `cost_region_value` takes the same `&[SyntaxToken]` the tail is already a slice of, so building a `Vec<&SyntaxToken>` and then cloning every token back out bought nothing. The region is sliced out of `tail` now.

**The bigger waste was upstream of it.** The function collected every direct token of the directive *before* checking whether there was a `~` at all — on a path that runs for every `balance` in every ledger, where the vast majority carry no tolerance. That check is now an iterator and the common path allocates nothing.

**And chasing it turned up two more things.** The check was commenting on input it doesn't understand: on `~ .005 + .005` the lexer already reports the real problem, and this added "the second was being discarded" next to it — a claim about input that never parsed. It now requires the region to be expression-shaped.

Which surfaced the actual bug: `.005` is a number **beancount accepts and we reject**, in every numeric position — postings, balance amounts, prices, tolerances, metadata. Filed as #2201; out of scope here since the fix is in the number lexer and moves parser output corpus-wide.

Controls: forcing the early-exit to always fire, and removing the expression-shaped guard, each fail their own assertion.

Issue #2177, opened 2026-08-28
BQL: ORDER BY on an expression target fails; the header and the lookup spell it differently


robcohen:
Found reviewing #2175. Pre-existing — `main` fails identically — but that PR
brings it within reach.

```
$ bean-query -f csv f.bean "SELECT number + 1 FROM #postings ORDER BY number + 1"
number + 1
...

$ rledger query f.bean "SELECT number + 1 FROM #postings ORDER BY number + 1"
error: ORDER BY expression not found in SELECT: (number + 1)
```

The two sides disagree on how to spell the expression:

- the result header comes from `ast::header_name` (since #2175):
  `number + 1`
- `sort_results` looks the target up by `spec.expr.to_string()`, i.e.
  `Display`: `(number + 1)`
- and `find_hidden_order_by_targets` names its synthetic hidden column with
  `Display` too, so the fallback path spells it the same wrong way

Before #2175 the header was `col0`, so the lookup missed for a different
reason. Either way the query has never worked.

## Fix

Name the hidden column and perform the lookup with the same function that
produces the header. That is a three-line change on the face of it, but it
touches the hidden-ORDER-BY-column machinery that #2170 had to repair twice,
so it wants its own change with the wildcard, PIVOT BY and
`ORDER BY <ordinal>` paths re-verified rather than being folded into a header
PR.

Note `find_column` (#2170) will not paper over this: it falls back
case-insensitively, and these two spellings differ structurally, not by case.

PR #2212, opened 2026-09-01
fix(query): MIN and MAX order booleans, false before true

### LINKED ISSUE #2183: BQL: max()/min() over a boolean raise a type error; bean-query orders False < True
Found reviewing #2182 while checking every path that can produce a boolean.
Pre-existing and unrelated to that PR, which changes rendering only — this
fails before any value reaches the formatter.

```
$ bean-query -f csv f.bean "SELECT max(number > 0) FROM #postings"
TRUE

$ rledger query f.bean "SELECT max(number > 0) FROM #postings"
error: type error: cannot compare values
```

Same for `min`. `MAX`/`MIN` work on numbers (`max(number)` returns `10.00`),
so this is specifically that booleans are not orderable in the comparison the
aggregate uses.

## Why bean-query can do it

Python orders `False < True`, so `max` over booleans is just `any` and `min` is
`all`. Our comparison rejects the pair instead of defining an order.

## Scope

`count(number > 0)` works, and boolean columns themselves render correctly
after #2182. This is only the ordering aggregates.

Probably a small fix — define an order for `Value::Boolean` in whatever the
aggregate comparison uses — but worth checking what else that comparison is
shared with before changing it, since making two booleans orderable also makes
them sortable in `ORDER BY`, which is a second behaviour change.

robcohen:
Closes #2183.

```
$ rledger query f.bean "SELECT max(number > 0) FROM #postings"
error: type error: cannot compare values        # before

$ bean-query -f csv f.bean "SELECT max(number > 0) FROM #postings"
TRUE
```

Python orders `False < True` and its `max`/`min` inherit that, so **MAX over booleans is `any` and MIN is `all`**. bean-query gets it for free; we had to define the order.

## The boundary is the interesting part

The order is defined for the **aggregate path only**, and that is measured, not assumed. bean-query refuses the operators on the very same values:

```
$ bean-query f.bean "SELECT (number > 0) < (number > 5) FROM #postings"
CompilationError: operator "less(bool, bool)" not supported
```

So `compare_values` — which backs `<` and `>` — keeps rejecting booleans. Defining the order there would accept a query bean-query rejects at compile time, a divergence in the *opposite* direction from the one being fixed.

That split is expressible only because `value_less_than` is reached from nowhere but `MIN`/`MAX`; I checked its call sites before touching it.

## Verification

End to end against **beanquery 0.2.0 / beancount 3.2.3**, through the real `bean-query` CLI:

```
SELECT MAX(number > 0), MIN(number > 0)
  bean-query:  TRUE,FALSE
  rustledger:  TRUE,FALSE
```

Byte for byte. That query joins `tests/compatibility/bql-queries.toml`, whose header asks for an entry whenever an aggregation function changes.

Worth noting: my first comparison used beanquery's **Python API** plus `csv.writer`, which printed `True`/`False` and looked like a mismatch. That bypasses bean-query's own renderer — the CLI prints `TRUE`/`FALSE`, matching #2182. The oracle has to be the CLI when output format is what's under test.

## Controls

| mutation | result |
|---|---|
| remove the boolean arm from `value_less_than` | fails |
| **also** add it to `compare_values` | **fails** |

The second is the one worth having: it stops the order being extended to the operators later as an obvious tidy-up.

Numbers still aggregate unchanged (`max(number)` → 10, `min(number)` → -10), so the new arm shadows nothing.

### DIFF
--- a/crates/rustledger-query/src/executor/operators.rs
+++ b/crates/rustledger-query/src/executor/operators.rs
@@ -110,8 +110,24 @@ impl Executor<'_> {
 
     /// Check if left value is less than right value.
     ///
-    /// Ordering used by the `MIN`/`MAX` aggregates, and only by them -- see the
-    /// `Boolean` arm.
+    /// One of THREE comparison paths in this file, which deliberately disagree
+    /// about booleans. Anyone tempted to unify them should read this first:
+    ///
+    /// | function | used by | booleans |
+    /// |---|---|---|
+    /// | `compare_values` | the `<` `>` `<=` `>=` operators | rejected |
+    /// | `value_less_than` | the `MIN`/`MAX` aggregates | ordered, `false < true` |
+    /// | `compare_values_for_sort` | `ORDER BY` | ordered, and total |
+    ///
+    /// That is not drift, it is bean-query's shape. It answers
+    /// `max(number > 0)` and sorts a boolean column, and refuses
+    /// `(number > 0) < (number > 5)` with `operator "less(bool, bool)" not
+    /// supported`. Collapsing the three would either accept a query it rejects
+    /// or reject two it answers.
+    ///
+    /// `compare_values_for_sort` is also total where the other two are
+    /// fallible: sorting cannot fail partway through a result set, so it
+    /// orders NULL rather than erroring on it.
     pub(super) fn value_less_than(&self, left: &Value, right: &Value) -> Result<bool, QueryError> {
         let ord = match (left, right) {
             (Value::Number(a), Value::Number(b)) => a.cmp(b),
--- a/crates/rustledger-query/src/executor/operators.rs
+++ b/crates/rustledger-query/src/executor/operators.rs
@@ -109,6 +109,9 @@ impl Executor<'_> {
     }
 
     /// Check if left value is less than right value.
+    ///
+    /// Ordering used by the `MIN`/`MAX` aggregates, and only by them -- see the
+    /// `Boolean` arm.
     pub(super) fn value_less_than(&self, left: &Value, right: &Value) -> Result<bool, QueryError> {
         let ord = match (left, right) {
             (Value::Number(a), Value::Number(b)) => a.cmp(b),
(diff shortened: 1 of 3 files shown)


robcohen:
Review notes on this PR.

**Verified the claim I made about scope.** All four `value_less_than` call sites are `MIN`/`MAX` — lines 476 and 938 inside `"MIN"`, 502 and 965 inside `"MAX"` (the grouped and ungrouped paths). Nothing else reaches it, so confining the boolean order to the aggregates is structurally sound rather than a convention.

**NULLs are handled before the comparison.** Both aggregates skip `Value::Null` (`if matches!(val, Value::Null) { continue; }`), so the new arm never sees one.

**`GROUP BY` agrees**: `SELECT max(number > 0) GROUP BY account` gives `TRUE,TRUE` from both tools.

Two things this turned up that are *not* in scope here:

1. `ORDER BY number > 0` still fails — that is #2177, the next issue in this cluster, and it reproduces identically on `main`.

2. `max(payee != '')` returns `FALSE` here and null in bean-query. **The divergence is in the comparison, not the aggregate**: `payee != ''` already yields `FALSE` per row on `main`, before this PR. All four comparison operators return `FALSE` where bean-query returns NULL, and it changes an answer — `count(payee != '')` is 0 there and 4 here. Filed as #2213.

Worth stating plainly: this PR makes that pre-existing divergence *visible* where it previously hit `type error: cannot compare values`. Trading an error for a wrong value is not obviously an improvement in isolation — but the error was also wrong, since the query is valid and bean-query answers it, and #2213 fixes the actual cause rather than hiding it behind a type error.

robcohen:
Checked the sibling aggregates, since MIN/MAX cannot be the only ones that meet a boolean.

| aggregate | bean-query | rustledger |
|---|---|---|
| `count(number > 0)` | 4 | 4 |
| `first` / `last` | TRUE | TRUE |
| `max` / `min` | TRUE / FALSE | TRUE / FALSE (this PR) |
| `avg(number > 0)` | `CompilationError: no function matches "avg(bool)"` | type error — **agrees** |
| `sum(number > 0)` | `2`, printed as `TRUE` | type error |

`avg` needs nothing: it agrees by refusing. `sum` is the only gap, and bean-query's own behaviour there is broken — the API returns the integer `2` while the CLI prints `TRUE`, because the result column is typed from the argument. "Match bean-query" is ambiguous when its value and its output disagree, so I filed it as #2214 rather than widening this PR to guess.

That also settles the scope question for this PR: MIN and MAX were the only two aggregates where bean-query orders booleans and we errored.

One more note for whoever picks up #2177: **`ORDER BY` needs no ordering work.** `compare_values_for_sort` already handles booleans and is total — that is the third path in the table now documented on `value_less_than`. #2177 is purely the name-lookup mismatch it describes.

Issue #2180, opened 2026-08-29
BQL: #balances.discrepancy needs the checker's computed difference, not a schema entry


robcohen:
Split out of #2154, which is closing: that issue was about columns bean-query
exposes that our schemas did not register, and every one of those is now
registered. This one is different in kind — the data does not exist in our
model, so it cannot be registered.

`#balances` exposes `discrepancy` in bean-query, in its `SELECT *`:

```
$ bean-query -f csv f.bean "SELECT * FROM #balances"
date,account,amount,tolerance,discrepancy

$ rledger query -f csv f.bean "SELECT * FROM #balances"
date,account,amount,tolerance
```

## Why it was not fixed with the rest

`discrepancy` is the difference between the asserted balance and the computed
one. `Balance` carries `date`, `account`, `amount`, `tolerance` and `meta` —
the checker computes the difference during validation and does not store it on
the directive. Surfacing the column means plumbing the checker's result through
to the query layer, not adding a column to a table builder.

Faking it from what the directive has would mean recomputing balances inside
the query executor, which is both expensive and a second implementation of
something the checker already does.

## Related

The other model-level gap from the same census is #2160 — `Note` drops its
tags and links at parse time, so `#notes.tags` and `#notes.links` cannot be
registered either.

Together those three are the entire remaining difference between our system
table schemas and bean-query's: a census of all 55 columns bean-query exposes
across its ten tables finds exactly these unselectable in ours.

Issue #2176, opened 2026-08-28
BQL: string-literal headers do not preserve the query's quote style


robcohen:
Found reviewing #2175 (Copilot flagged the ambiguity; measuring bean-query
showed the rule behind it).

bean-query echoes the quote character the query wrote:

```
$ bean-query -f csv f.bean "SELECT 'plain' FROM #notes"     ->  'plain'
$ bean-query -f csv f.bean "SELECT \"dq\" FROM #notes"      ->  "dq"
$ bean-query -f csv f.bean "SELECT \"o'clock\" FROM #notes" ->  "o'clock"
```

We render every string literal single-quoted unless it contains a single
quote, so `"dq"` heads `'dq'`.

## Why #2175 did not fix it

`Literal::String(String)` does not record which quote character the source
used, so the query layer cannot reproduce it. Matching would mean the parser
keeping the original style on the literal — a change to the AST and every
construction site, to change which quote character a header prints.

#2175 does fix the part that is a defect rather than a divergence: a value
containing a single quote is now double-quoted, so the header is a well-formed
literal instead of the ambiguous `'o'clock'`. That case now agrees with
bean-query as a side effect.

## Worth deciding, not just doing

The remaining gap is cosmetic and affects only string-literal targets, which
are rare in real queries. Recording the quote style would also let `Display`
round-trip more faithfully, which may matter more than the header does.

PR #2215, opened 2026-09-01
fix(query): ORDER BY an expression finds the column it names

### LINKED ISSUE #2177: BQL: ORDER BY on an expression target fails; the header and the lookup spell it differently

robcohen:
Closes #2177.

```
$ rledger query f.bean "SELECT number + 1 FROM #postings ORDER BY number + 1"
error: ORDER BY expression not found in SELECT: (number + 1)

$ bean-query -f csv f.bean "SELECT number + 1 FROM #postings ORDER BY number + 1"
number + 1
-9.00
-2.00
 4.00
11.00
```

The header and the lookup spelled the same expression two ways: `ast::header_name` gives `number + 1`, `Display` parenthesizes and gives `(number + 1)`, and `sort_results` looked up by `Display`. `find_column`'s case-insensitive fallback cannot bridge that — the spellings differ by **punctuation, not case**. Before #2175 the header was `col0`, so the lookup missed for a different reason; the query has never worked.

## Why both sites move together

`sort_results` **looks the target up**. `find_hidden_order_by_targets` **names** the synthetic column when the expression is not reachable in the projection. Changing one alone trades one miss for another — a hidden column named by `Display` is invisible to a lookup by `header_name`.

Each half has a test only it satisfies:

| mutation | result |
|---|---|
| revert the lookup to `Display` | 3 of 4 tests fail |
| revert **only** the hidden-column naming | exactly the hidden-column test fails |

## The neighbouring paths, re-verified rather than trusted

The issue asked for this explicitly, since the hidden-column machinery needed repair twice in #2170. Measured against **bean-query** (beanquery 0.2.0), row for row:

| query | result |
|---|---|
| `ORDER BY 1`, `ORDER BY 2` (ordinals) | match |
| `ORDER BY account, number` (multi-key) | match |
| `ORDER BY number DESC` | match |
| `GROUP BY account ORDER BY sum(number)` | match |
| `SELECT account AS a ORDER BY number + 1` (hidden column) | match |
| `SELECT number > 0 ORDER BY number > 0` | match |

**Wildcard unchanged**: `SELECT * FROM #postings ORDER BY date` output is byte-identical to `main`. (It differs from bean-query — 31 columns to its 5 — but that predates this PR and is untouched by it.)

Numbers differ from bean-query only by its **right-alignment padding** in CSV (`, 10.00` against `,10.00`), which is pre-existing and independent of ordering.

## Corpus

`order-by-expression` joins `tests/compatibility/bql-queries.toml`, verified to be real coverage rather than decoration: run through the harness it reports `match: True`, and `match: False` when the lookup is reverted.

## Note for #2176 / #2180

Sorting needed no ordering work — `compare_values_for_sort` already handles booleans and is total. This was purely the name lookup.

### DIFF
--- a/crates/rustledger-query/src/executor/execution.rs
+++ b/crates/rustledger-query/src/executor/execution.rs
@@ -305,7 +305,11 @@ impl Executor<'_> {
             // output column is named by the alias, so `sort_results` would fail
             // a literal name lookup for the raw column (#1627). In that case we
             // still append a hidden column carrying the raw name.
-            let expr_str = spec.expr.to_string();
+            // Same spelling the header uses, and the same one `sort_results`
+            // now looks up. These two have to agree: a hidden column named by
+            // `Display` would be invisible to a lookup by `header_name`, so
+            // changing one alone trades one miss for another (#2177).
+            let expr_str = crate::ast::header_name(&spec.expr);
             let in_select = query.targets.iter().any(|t| {
                 (t.expr == spec.expr && t.alias.is_none())
                     || t.alias.as_deref() == Some(expr_str.as_str())
--- a/crates/rustledger-query/src/executor/sort.rs
+++ b/crates/rustledger-query/src/executor/sort.rs
@@ -76,7 +76,14 @@ impl Executor<'_> {
                 Expr::Function(func) => {
                     // First try to find a column with the function name (e.g., "sum" for sum(amount))
                     // Then try the full expression string (e.g., "account_sortkey(account)")
-                    let expr_str = spec.expr.to_string();
+                    //
+                    // `header_name`, not `Display`, for the same reason as the
+                    // arm below: `Display` parenthesizes a binary argument, so
+                    // `abs(number + 1)` looked for `abs((number + 1))` against
+                    // an `abs(number + 1)` header. `abs(number)` worked, which
+                    // is why the bug hid here -- an argument has to be compound
+                    // before the two spellings part company (#2177 review).
+                    let expr_str = crate::ast::header_name(&spec.expr);
                     find_column(&result.columns, &func.name)
                         .or_else(|| find_column(&result.columns, &expr_str))
                         .ok_or_else(|| {
--- a/crates/rustledger-query/src/executor/sort.rs
+++ b/crates/rustledger-query/src/executor/sort.rs
@@ -87,8 +87,18 @@ impl Executor<'_> {
                 }
                 _ => {
                     // For other expression kinds (binary ops, literals, etc.),
-                    // look up by string representation (matches hidden column aliases).
-                    let expr_str = spec.expr.to_string();
+                    // look up by the name the HEADER uses.
+                    //
+                    // `Display` is not that name and never was: it parenthesizes,
+                    // so `ORDER BY number + 1` looked for `(number + 1)` while the
+                    // header said `number + 1`, and the sort failed on a query
+                    // bean-query answers. Before #2175 the header was `col0`, so
+                    // the lookup missed for a different reason -- the query has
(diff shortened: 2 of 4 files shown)


robcohen:
Correcting my own claim on this PR, and reporting what checking it turned up.

**I said the paths the issue named were re-verified. PIVOT BY was not** — the issue listed wildcard, PIVOT BY and ordinals, and I checked the first and third. Done now:

`PIVOT BY` behaves **identically on `main` and this branch**, so the change does not touch it. That is what the issue wanted confirmed, and it holds.

But checking it found a pre-existing bug, filed as #2216: **`PIVOT BY` is silently ignored when the query has a `FROM` clause.**

```
SELECT ... GROUP BY account, year PIVOT BY account, year
  -> columns ["account", "2024"]                    pivoted

SELECT ... FROM #postings GROUP BY account, year PIVOT BY account, year
  -> columns ["account", "year", "sum(number)"]     not pivoted
```

Neither the parser nor `apply_pivot` is at fault — the parser populates `pivot_by` for the `FROM` form, and `apply_pivot` produces the right shape on the other one. The `FROM <table>` execution path returns before reaching the pivot step.

That also explains a wrong turn earlier in this review: my first PIVOT check used the CLI, which writes `FROM #postings`, so it looked like PIVOT BY was unimplemented rather than skipped on one path.

No change to this PR — #2216 is independent and pre-existing.

robcohen:
Both comments were right, and the first one was a real gap.

**`Expr::Function` still used `Display`.** I changed the binary-expression arm and left the function arm alone, so `abs(number + 1)` still looked for `abs((number + 1))` against an `abs(number + 1)` header:

```
$ rledger query f.bean "SELECT abs(number + 1) FROM #postings ORDER BY abs(number + 1)"
error: ORDER BY expression not found in SELECT: abs((number + 1))
```

`abs(number)` resolves either way, which is exactly how it hid — the argument has to be **compound** before the two spellings part company, and every function case I reached for was a bare column. Both arms now name the target the way the header does.

**And the test gap you named was the reason I missed it.** There are now two tests, not one, so a regression says which case broke: the compound-argument form and the bare-column form. Reverting only the function arm fails exactly the first, and leaves the other five passing.

Swept wider against bean-query while in there — `number * 2`, unary `-number`, `ORDER BY <expr> DESC`, a function as a secondary sort key, and an aliased target sorted by its alias all match row for row.

Issue #2217, opened 2026-09-01
BQL: pivot column headers do not match bean-query (key column and value/spread order)


robcohen:
With `PIVOT BY` actually applying (#2216), the rows agree with bean-query and the **headers do not** — in two separate ways.

```
SELECT account, year, sum(number), count(number) FROM #postings
GROUP BY account, year PIVOT BY account, year

bean-query:  account/year, 2024/sum(number), 2024/count(number)
rustledger:  account,      sum(number) / 2024, count(number) / 2024
```

1. **Row-key column.** bean-query names it `<key>/<spread>` — `account/year`. We emit just `account`, dropping the spread column's name.
2. **Value columns.** bean-query is `<spread_value>/<column>`; we are `<column> / <spread_value>` — the two halves are **reversed**, and we pad the separator with spaces where bean-query does not.

The single-value case only shows the first difference, because the lone value column is named by the spread value alone on both sides:

```
bean-query:  account/year, 2024
rustledger:  account,      2024
```

## Not pinned today

`apply_pivot` says it matches bean-query's `test_pivot_one_column`, and the validation rules genuinely do — arity, same-column, second-in-GROUP-BY. But the existing tests (`test_parse_pivot_by_two_columns`, `test_parse_pivot_by_one_column_parses_but_executes_with_arity_error`, `test_pivot_by_same_column_rejected`) cover parsing and rejection only. Nothing asserts an output header, which is why the naming could drift without notice.

## Why separate from #2216

That one is a silent wrong answer: the pivot was skipped entirely on any query with a `FROM` clause. This is a labelling difference on output that is otherwise correct, and fixing it changes the currently-working no-`FROM` path too — a different blast radius that deserves its own review.

Worth a `bql-queries.toml` entry once fixed, since the corpus diffs headers as well as rows.

robcohen:
Correcting something I wrote in this issue.

I said a `bql-queries.toml` entry would cover this once fixed, "since the corpus diffs headers as well as rows". **It does not.** `extract_data` in `scripts/compat-bql-test.py` skips every line before the dashed separator — which is the header — and compares only data rows.

Demonstrated: I added a temporary pivot entry to the corpus and ran the harness against both tools. It reported `match: True`, with our header `account,2024` against bean-query's `account/year,2024`. The very divergence this issue is about is invisible to it.

So a corpus entry will *not* protect this once fixed, and the harness gap is filed separately as #2220.

PR #2218, opened 2026-09-01
fix(query): PIVOT BY applies when the query names a table

### LINKED ISSUE #2216: BQL: PIVOT BY is silently ignored when the query has a FROM clause
`PIVOT BY` works without `FROM` and is **silently dropped** with one. No error, no warning — just an un-pivoted result.

```
SELECT account, year, sum(number) GROUP BY account, year PIVOT BY account, year
  -> columns ["account", "2024"]                      pivoted, correct

SELECT account, year, sum(number) FROM #postings GROUP BY account, year PIVOT BY account, year
  -> columns ["account", "year", "sum(number)"]       NOT pivoted
```

Since `rledger query` is normally written with an explicit table, the CLI form is the one that fails:

```
$ rledger query f.bean "SELECT account, year, sum(number) FROM #postings GROUP BY account, year PIVOT BY account, year"
account,year,sum(number)
Assets:A,2024,7.00
Equity:O,2024,-7.00

$ bean-query -f csv f.bean "<same>"
account/year,2024
Assets:A, 7.00
Equity:O,-7.00
```

## Not the parser, and not `apply_pivot`

Both are fine:

- the parser populates `pivot_by: Some(..)` for the `FROM` form — checked directly;
- `apply_pivot` is implemented, with validation mirroring bean-query's `_compile_pivot_by`, and produces the right shape on the no-`FROM` form.

So the `FROM <table>` execution path returns before reaching the pivot step in `execute_select`. Whatever handles a table-qualified query does not run the tail of the pipeline that `apply_pivot` sits in.

## Also worth a look while there

`FROM #transactions` fails differently:

```
error: column 'account' not found in subquery result
```

`account` is a posting-level column, so an error is defensible — but bean-query answers the same query, and the message describes a subquery the user did not write.

## Scope

Pre-existing. Confirmed identical on `main` and on the branch for #2177, so it is not a regression from the ORDER BY work — found while re-verifying the PIVOT BY path that issue asked to check.

Silent wrong output rather than an error, which puts it in the same family as #2213 and #2214.

robcohen:
Closes #2216.

`PIVOT BY` was **silently dropped** by any query with a `FROM` clause. It parsed, it validated, and the un-pivoted result came back with no error. Since `rledger query` is normally written with an explicit table, **the CLI could not pivot at all**:

```
$ rledger query f.bean "SELECT account, year, sum(number) FROM #postings GROUP BY account, year PIVOT BY account, year"
account,year,sum(number)          <- not pivoted
Assets:A,2024,7.00

$ bean-query -f csv f.bean "<same>"
account/year,2024
Assets:A, 7.00
```

## Four tails, one pivot

| path | sort | pivot | limit |
|---|---|---|---|
| `execute_select` (no FROM) | ✓ | **✓** | ✓ |
| `execute_select_from_table` | ✓ | — | ✓ |
| `execute_aggregate_from_table` | ✓ | — | ✓ |
| `execute_select_from_subquery` | ✓ | — | ✓ |

Neither the parser nor `apply_pivot` was at fault — the parser populates `pivot_by` for the `FROM` form, and `apply_pivot` produces the right shape when reached. The step was simply absent from three tails.

## Two fixed, one deliberately not

Both additions sit in the slot `execute_select` uses (after ORDER BY and the hidden-column strip, before LIMIT), and each is reachable and improves something different:

- **`execute_aggregate_from_table`** — the reported case. The pivot applies and the rows match bean-query.
- **`execute_select_from_table`** — cannot pivot, since `PIVOT BY` requires `GROUP BY`. But it now **says so**, with the same message the no-`FROM` path gives, instead of ignoring the clause. bean-query answers that input with a Python traceback, so an error beats both.

**The subquery path is left alone.** Our parser rejects `FROM (SELECT ...)` outright, so nothing can reach it carrying a pivot. A branch no test can exercise is worse than a documented gap — the table above lives in the test file for whoever adds the next pipeline step.

## Controls

| mutation | result |
|---|---|
| remove the `execute_select_from_table` insertion | the without-`GROUP BY` test fails |
| remove the `execute_aggregate_from_table` insertion | the other two fail |

Each insertion has a test only it satisfies. A third test pins that the `FROM` and no-`FROM` forms produce the same shape — the bug was precisely that they had drifted.

## Left for #2217

Headers still differ: `account` against bean-query's `account/year`, and value columns reversed (`sum(number) / 2024` against `2024/sum(number)`). That is labeling on an otherwise-correct result, it changes the currently-working no-`FROM` path too, and it deserves its own review.

### DIFF
--- a/crates/rustledger-query/src/executor/execution.rs
+++ b/crates/rustledger-query/src/executor/execution.rs
@@ -592,6 +592,11 @@ impl Executor<'_> {
         // it here meant a query with a FROM clause parsed its PIVOT BY, built
         // the transformation's validated inputs, and then returned the
         // un-pivoted result with no error (#2216).
+        //
+        // PIVOT BY requires GROUP BY, and a query with GROUP BY is routed to
+        // `execute_aggregate_from_table`, so reaching `apply_pivot` from here
+        // always ends in `PivotWithoutGroupBy`. That is the point: the clause
+        // is refused out loud rather than dropped.
         if let Some(pivot_exprs) = &query.pivot_by {
             result = self.apply_pivot(&result, pivot_exprs, &query.group_by)?;
         }
@@ -728,11 +733,16 @@ impl Executor<'_> {
             self.sort_results(&mut result, order_by, visible_cols)?;
         }
 
-        // Apply PIVOT BY, in the same pipeline slot `execute_select` uses:
-        // after ORDER BY and the hidden-column strip, before LIMIT. Omitting
-        // it here meant a query with a FROM clause parsed its PIVOT BY, built
-        // the transformation's validated inputs, and then returned the
-        // un-pivoted result with no error (#2216).
+        // Apply PIVOT BY, in the pipeline slot `execute_select` uses: after
+        // ORDER BY, before LIMIT. Omitting it here meant a query with a FROM
+        // clause parsed its PIVOT BY, built the transformation's validated
+        // inputs, and then returned the un-pivoted result with no error
+        // (#2216).
+        //
+        // No hidden-column strip to sit after, unlike the other two paths:
+        // `find_hidden_order_by_targets` materializes nothing for an aggregate
+        // query, because an ORDER BY target there must already be in GROUP BY
+        // or be an aggregate, so it is projected already.
         if let Some(pivot_exprs) = &query.pivot_by {
             result = self.apply_pivot(&result, pivot_exprs, &query.group_by)?;
         }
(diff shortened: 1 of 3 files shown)


robcohen:
Correct, and fixed in the follow-up commit.

I wrote one comment and pasted it into both places, so `execute_aggregate_from_table` claimed the pivot runs "after ORDER BY and the hidden-column strip". That path has **no strip** — zero mentions of `num_hidden` — because `find_hidden_order_by_targets` materializes nothing for an aggregate query: an ORDER BY target there must already be in GROUP BY or be an aggregate, so it is projected already.

Each comment now describes the path it is in, and says the thing a reader of *that* path needs:

- the **aggregate** path explains why there is no strip to sit after;
- the **table** path records that reaching `apply_pivot` from there **always** ends in `PivotWithoutGroupBy`, since the function returns early to the aggregate path whenever `group_by.is_some()`. I checked that routing condition rather than assuming it.

On the second point: calling `apply_pivot` anyway, instead of raising the error inline, keeps one source of truth for the validation rules and their wording — the alternative would be a second copy of the message to drift.

Issue #2222, opened 2026-09-01
BQL: multi-currency SUM(position) renders without bean-query's per-currency slots


robcohen:
Surfaced by switching the BQL harness to CSV (#2220). The tabular comparison collapsed runs of whitespace, which hid this.

```
SELECT account, SUM(position) GROUP BY account

bean-query:  Assets:CORP✨," 100 CORP { 1 USD},        "
             Assets:Cash,"                  ,-100 USD"

rustledger:  Assets:CORP✨,100 CORP { 1 USD}
             Assets:Cash,-100 USD
```

bean-query renders a multi-currency inventory as **one field holding a slot per currency** seen across the result, comma-separated, with empty slots padded so the currencies line up vertically. We render only the currencies actually present.

After collapsing whitespace the difference is structural, not cosmetic:

```
bean-query:  100 CORP { 1 USD},        <- trailing empty USD slot
rustledger:  100 CORP { 1 USD}

bean-query:  ,-100 USD                 <- leading empty CORP slot
rustledger:  -100 USD
```

## Which is right

Ours is more readable and does not embed presentation into a data field. bean-query's carries information ours does not: which currencies the *result as a whole* spans, so a reader can see that `Assets:Cash` holds no CORP.

Since the field is one CSV cell either way, neither is more machine-parseable — bean-query's just has a comma-separated substructure a consumer would have to split again.

Affects 16 of 285 runs in a 15-file sample, across `sum-position-by-account`, `first-balance-by-month`, `last-balance-by-month`, `balance-running-assets` and `position-by-date` — i.e. every corpus query that sums a position over a multi-currency ledger.

Found while implementing #2220.

robcohen:
Note on observability, from #2223: the harness compares **values** on the text surface, so it cannot see this divergence.

The slot padding is visible only in CSV output, and values are deliberately not compared there — CSV is a machine surface where `DisplayContext::for_surface` suppresses thousands separators absolutely (#1892), so comparing values as CSV stops exercising every `render_commas` divergence in the registry.

So this needs a targeted test rather than corpus coverage. The five `("*", query)` entries added for it in #2223 were removed again, since nothing in the harness observes the divergence they were masking.

robcohen:
Settled the same way as #2221 — by reading the source, not diffing the output: **the slot layout is bean-query's renderer, and our data already matches its data.**

Two pieces of evidence.

The layout comes from a **renderer**, `query_render.py`:

```
[...]

Issue #2221, opened 2026-09-01
BQL: BALANCES and JOURNAL name their columns differently from bean-query


robcohen:
Surfaced by turning on header comparison in the BQL harness (#2220). These were invisible before, because the harness dropped the header line.

Measured across 15 corpus files, every one of them:

| query | bean-query | rustledger |
|---|---|---|
| `BALANCES` | `account,SUM((position))` | `account,balance` |
| `JOURNAL` | `date,flag,MAXWIDTH(payee, 48),MAXWIDTH(narration, 80),account,position,balance` | `date,flag,payee,narration,account,position,balance` |
| `JOURNAL ... AT UNITS` | `...,units(position),units(balance)` | `...,position,balance` |
| `JOURNAL ... AT COST` | `...,cost(position),cost(balance)` | `...,position,balance` |

The **rows agree** in all of these; only the column names differ.

## These look deliberate, not accidental

bean-query names a shorthand command's columns after the expression it desugars to, so `BALANCES` heads `SUM((position))` and `JOURNAL` exposes its `MAXWIDTH` wrapping. Ours are the plain names a reader would expect — `balance`, `payee`, `narration`.

We do implement the MAXWIDTH *behavior* (#2044): `journal` collapses narration whitespace where a plain `SELECT` does not. It is only the header that differs.

So the question is which output is right, not which is a bug:

- **Keep ours** and record it as a deliberate divergence. Anything scripting against our CSV header would break if we changed it.
- **Match bean-query** for tools that parse headers positionally by name.

## Note

`AT UNITS` / `AT COST` are the more interesting pair: bean-query's header records which valuation was applied (`units(position)` against `cost(position)`), and ours does not — so our headers for those two are **identical** despite the values differing. That is a real loss of information regardless of which convention wins.

Found while implementing #2220.

robcohen:
Settled by reading beanquery's source rather than diffing its output: **these names are not chosen, and we should keep ours.**

`transform_journal` (`compiler.py`) builds the JOURNAL query by parsing a SQL string template, with no `AS` aliases on any target:

```python
cooked_select = parser.parse("""
    SELECT
       date,
       flag,
       MAXWIDTH(payee, 48),
       MAXWIDTH(narration, 80),
       account,
       {summary_func}(position),
       {summary_func}(balance)
    {where}
""".format(...))
```

[...]

Issue #2164, opened 2026-08-27
BQL: result headers echo the query's casing; bean-query normalizes to the column name


robcohen:
Found reviewing #2161, which does not touch this — it affects every column on
every table.

bean-query normalizes a result header to the column's canonical name. We echo
whatever the query wrote.

```
$ bean-query -f csv f.bean "SELECT DATE FROM #notes"
date

$ rledger query -f csv f.bean "SELECT DATE FROM #notes"
DATE
```

Same for `SELECT NARRATION FROM #transactions` (`narration` vs `NARRATION`),
so it is not specific to any table or column.

This matters for anything consuming our CSV by header name: a script written
against `bean-query` that reads the `narration` field gets `NARRATION` from us
and misses it, with no error.

## Where

`Executor::expr_to_name` returns the column expression's source spelling:

```rust
Expr::Column(name) => name.clone(),
```

Resolution itself is already case-insensitive — only the emitted header keeps
the caller's casing. Note the fix must not disturb explicit aliases
(`SELECT meta AS M` should still emit `M`) or the postfix spellings that
`ORDER BY entry.narration` depends on (#1800).

## Related

Third presentation-level divergence found in the #2161 reviews, after #2162
(metadata content) and #2163 (row order). All three share a shape: the column
resolves and the query succeeds, but what comes back differs from bean-query
in a way header-presence checks cannot see. The BQL compat suite reports 100%
against these same tables, so the gap is in what the suite compares, not in
how much it covers.

PR #2223, opened 2026-09-01
fix(ci): the BQL harness compares column names

### LINKED ISSUE #2220: BQL compat harness compares data rows only, so column-name divergences are invisible
`scripts/compat-bql-test.py` diffs bean-query against rledger row by row — and **drops the header before comparing**, so any difference in column names passes as a match.

`extract_data` scans for the dashed separator and only starts collecting after it:

```python
for line in output.split("\n"):
    if not found_sep:
        if stripped and "-" in stripped and _SEPARATOR_RE.fullmatch(stripped):
            found_sep = True
        continue          # everything above the separator, i.e. the header, is discarded
    ...
    rows.append(" ".join(stripped.split()))
```

## Demonstrated

A temporary pivot entry, run against both tools:

```
rustledger header:  account,2024
bean-query header:  account/year,2024
harness verdict:    match: True
```

The headers plainly differ and the harness is satisfied.

## Why it matters more than it sounds

Column names are user-visible output, and this repository has done real work on them that the corpus therefore never covered:

- #2164 and #2175 — how a result column is named (bare column lowercased, function keeps its source spelling, literals rendered with minimal parentheses)
- #2176 — string-literal headers not preserving the query's quote style
- #2217 — pivot headers (`account` against `account/year`, and value columns reversed)

Each of those is exactly the class the BQL corpus exists to catch, and none of them could have been caught by it.

It also means a passing corpus is weaker evidence than it reads: a change that renamed every column would not move a single result.

## Suggested shape

Compare the header line as a row of its own, with the same normalization the data rows get. That would want an audit first — some divergence is already known and deliberate, so the initial run will likely surface entries needing either a fix or a `KNOWN_RUST_DIVERGENCES` entry, and those should be triaged rather than mass-masked.

## Found

Checking whether a corpus entry would protect #2217 once fixed. It would not, which is also a correction to that issue's text.

robcohen:
Closes #2220.

The harness diffed bean-query against rledger row by row and **dropped the header before comparing**, so any difference in column names passed as a match. Column names are user-visible output, and every piece of work this repo has done on them — #2164, #2175, #2176, #2217 — was invisible to the corpus that exists to catch exactly that.

## The obvious fix doesn't work

The default table **truncates a header to its data column width**:

```
$ bean-query f.bean "SELECT account, sum(number) GROUP BY account"
account   sum(n          <- narrow values, header cut
--------  -----

# same query, values 1234567.89:
account   sum(number)    <- wide values, header intact
```

Our number padding differs from bean-query's, so comparing tabular headers would manufacture mismatches that are not divergences.

So both tools are asked for **`-f csv`**, which never truncates — same number of invocations. Rows are now parsed as CSV and normalized **per field** rather than per line, which is strictly better: a value containing a comma stays intact because the reader has already split on the right ones.

## Two real divergence classes were hiding under the old comparison

Both filed, not papered over:

**#2221 — `BALANCES` / `JOURNAL` column naming.** bean-query heads `SUM((position))` and `MAXWIDTH(payee, 48)`; ours are the plain names. Rows agree. Worth noting: our `AT UNITS` and `AT COST` headers are **identical to each other**, where bean-query's record which valuation ran — so that pair loses information regardless of which convention wins.

**#2222 — multi-currency `SUM(position)`.** bean-query returns one field holding a slot per currency seen across the whole result, empties included (`100 CORP { 1 USD},` against our `100 CORP { 1 USD}`). Whitespace collapsing in the tabular reader hid it entirely.

## Registering them needed a query wildcard

`("*", query)` marks a divergence that is a property of the **query** rather than any ledger. A column-naming difference holds on all 735 corpus files, and spelling that as 735 paths would be mass-masking wearing a list.

The pre-existing bare-name guard — which catches registry keys that could never match — now exempts `"*"` while still flagging a genuinely bare filename. Verified both ways.

The design already prevents these entries from becoming permanent cover: `stale_divergence_entries` fails the run if a registered pair starts matching again.

## Validation

Against a **baseline**, which is what the gate actually uses:

| | exit |
|---|---|
| with the registry entries | **0** — no regressions vs the pre-change baseline |
| with one entry removed | **1** — regression detected |

My first attempt at that control ran without `--baseline` and returned 0 either way, so it proved nothing. Worth stating because the harness only gates on baseline regressions, not on absolute mismatches — a run with no baseline never fails.

Harness self-test passes; `py_compile` and `typos` clean.

### DIFF
--- a/scripts/compat-bql-test.py
+++ b/scripts/compat-bql-test.py
@@ -26,12 +26,15 @@ Usage (local, with paths and tools auto-detected):
 from __future__ import annotations
 
 import argparse
+import csv
+import io
 import json
 import multiprocessing
 import os
 import re
 import subprocess
 import sys
+from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -386,9 +389,10 @@ KNOWN_PYTHON_DIVERGENCES: set[tuple[str, str]] = {
 # absorbed by the same allowlist. Reported separately in the per-CI
 # summary so the count of Rust-side divergences is visible at a glance.
 #
-# Both lists are stale-checked at runtime (`stale_divergence_entries`): an entry
-# whose pair now MATCHES bean-query fails the run, so a mask can't outlive the
-# divergence it documents and then absorb a real regression on the same pair.
+# Every registry here is stale-checked at runtime (`stale_divergence_entries`):
+# an entry whose pair now MATCHES bean-query fails the run, so a mask can't
+# outlive the divergence it documents and then absorb a real regression on the
+# same pair.
 #
 # Keyed by `(repo_relative_path, query_name)` with the same surgical-pin
 # semantics as `KNOWN_PYTHON_DIVERGENCES`. Counted as "known" for the effective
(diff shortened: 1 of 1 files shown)


robcohen:
All four review findings were real and are fixed — and chasing the first one surfaced a defect in the PR's own premise that CI had already caught.

**The compat job was failing, and it was right to.** Switching the *value* comparison to `-f csv` turned nine `render_commas` masks stale. Not because rledger and bean-query had come to agree, but because CSV is a machine surface where `DisplayContext::for_surface` suppresses thousands separators absolutely (#1892) — so the harness had stopped looking at the very thing those masks document. Deleting the nine to get green would have been exactly backwards.

Values are back on the text surface; only the *header* pass uses CSV, as a second invocation. Verified against main over 71 corpus files / 1349 runs: **verdicts identical pair-for-pair (0 differences)**, and 204 header divergences newly detected across four queries. Runtime ~1.7x, not 2x — the header pass is skipped when either tool already failed.

**Findings**

1. **`("*", query)` never stale-checked** — correct, and the CI log showed all nine warning as dangling on every run. `_stale_registry_entries` now resolves both wildcard directions (plus `("*", "*")`).

2. **Comma-joined headers can compare equal** — correct. `extract_header` returns a `list[str]`; nothing is re-joined.

3. **`header_match` True when either header is `None`** — correct. A `None` now means "we did not look" and is tracked as `header_skipped`, reported in the summary rather than scored as agreement.

4. **Stale "separator to anchor on" comment** — fixed.

**One more, self-inflicted.** Parking the header wildcards in `KNOWN_RUST_DIVERGENCES` masks the *whole run*, so a future **data** regression in `JOURNAL` on any of the 735 corpus files would have been absorbed by a mask only ever meant to cover a column's name. They now live in `KNOWN_HEADER_DIVERGENCES`, consulted only when scoring headers; rows still gate. Confirmed it gates: removing one entry takes real mismatches 21 → 69, and re-running against the clean baseline reports exactly 48 regressions.

Only four of the original nine wildcards survive. The other five were the #2222 inventory-slot padding, which the text surface legitimately does not see.

**Negative controls.** Every new behavior has one in `--self-test`: mis-wiring the header predicate, counting a skipped header as agreement, dropping the `("*", query)` direction, or hardcoding the bare-name guard's registry list each make the suite fail (exit 1). The guard now derives its list from the same table that pairs each registry with its predicate — `KNOWN_HEADER_DIVERGENCES` escaped that guard when I first added it, and the self-test plants a bare key in each registry in turn to prove it can't happen again.

PR #2225, opened 2026-09-01
fix(query): a comparison with a NULL operand yields NULL

### LINKED ISSUE #2213: BQL: comparing NULL yields FALSE instead of NULL, so count() over a comparison is wrong
A comparison against a NULL column returns `FALSE` here and NULL in bean-query. Measured against **beanquery 0.2.0 / beancount 3.2.3** on a ledger whose transactions have no payee:

| expression | bean-query | rustledger |
|---|---|---|
| `payee != ''` | *(null)* | `FALSE` |
| `payee = ''` | *(null)* | `FALSE` |
| `payee > 'a'` | *(null)* | `FALSE` |
| `payee < 'a'` | *(null)* | `FALSE` |
| `payee IS NULL` | `TRUE` | `TRUE` |
| `NOT (payee = '')` | `TRUE` | `TRUE` |

All four comparison operators, consistently.

## Why it is not cosmetic

`COUNT(expr)` counts non-NULL values, so the difference changes an answer:

```
SELECT count(payee != '') FROM #postings
  bean-query:  0      # every row is NULL, so nothing is counted
  rustledger:  4      # every row is FALSE, which counts
```

`count()` excluding NULLs is already implemented and tested here (`aggregate_count_null_test.rs`); it is the comparison feeding it that produces a non-NULL where there should be none.

## Note on the target semantics

bean-query is **not** strict SQL three-valued logic, so "make it SQL" is the wrong instruction. `NOT (NULL)` is `TRUE` there, which is Python's `not None`, not SQL's `NULL`. The rule to match is Python's: a comparison with a NULL operand yields NULL, and `NOT` coerces.

One more observation, possibly a separate thing: `SELECT count(*) FROM #postings WHERE payee != ''` returns **no row at all** from bean-query, where we return `0`. That is an aggregate-over-empty-input difference rather than a NULL-propagation one, and I have not chased it.

## Found

Reviewing #2212, which makes `MAX`/`MIN` work over booleans. `max(payee != '')` returns `FALSE` here and null in bean-query — but the divergence is in the comparison, not the aggregate: `payee != ''` already yields `FALSE` per row on `main`, before that PR. #2212 only makes the aggregate reachable, so this surfaces where it previously hit `type error: cannot compare values`.

robcohen:
`payee != ''` answered `FALSE` on a payee-less posting where bean-query answers NULL. Closes #2213.

FALSE and NULL are not interchangeable downstream. `COUNT(expr)` skips NULLs but counts every FALSE, so `count(payee != '')` over payee-less rows answered **4** where it should answer **0**. `COUNT` excluding NULLs was already correct and tested — it was the comparison feeding it that manufactured a non-NULL.

All six comparison operators shared one guard, so all six were wrong. The fix is that guard returning `Value::Null`.

## Target semantics

Python's rule, which is what beanquery implements — not strict SQL three-valued logic. `NOT (NULL)` is `TRUE` there (Python's `not None`) where SQL would say NULL. That behavior stays and is now pinned by a test, so a later "make it proper 3VL" change meets the decision rather than assuming the current answer is an oversight.

Filtering is unaffected: `to_bool(NULL)` is false, so `WHERE` still drops the row. That is why this survived — filtering looked right, and only projecting or counting a comparison exposed it.

## Verification

Against beanquery 0.2.0 / beancount 3.2.3, all seven expressions from the issue now agree:

| expression | bean-query | before | after |
|---|---|---|---|
| `payee != ''` | *(null)* | `FALSE` | *(null)* |
| `payee = ''` | *(null)* | `FALSE` | *(null)* |
| `payee > 'a'` | *(null)* | `FALSE` | *(null)* |
| `payee < 'a'` | *(null)* | `FALSE` | *(null)* |
| `payee IS NULL` | `TRUE` | `TRUE` | `TRUE` |
| `NOT (payee = '')` | `TRUE` | `TRUE` | `TRUE` |
| `count(payee != '')` | `0` | `4` | `0` |

Also checked on a mixed ledger that ordinary comparisons are untouched (`'Cafe' = 'Cafe'` → TRUE, `WHERE payee != ''` → 2 rows), agreeing with bean-query on each.

BQL compat harness over 71 corpus files / 1349 runs: **no regression against the pre-change baseline**, no new stale masks, counts unchanged. The corpus barely exercises NULL comparisons, which is part of why this survived.

## On the coverage gap

The full workspace suite (4686 tests) passed **unchanged** after the semantic fix, before any test was added — nothing covered this at all. Two of the five new tests fail against the old behavior; the other three deliberately pass both ways because they pin invariants that must *not* move (`WHERE` filtering, `NOT` coercion, and ordinary two-operand comparisons still being boolean). Without that third group, returning NULL unconditionally would satisfy the rest of the file.

Not addressed here: the issue's secondary note that `SELECT count(*) ... WHERE <no matches>` returns no row from bean-query where we return `0`. That is an aggregate-over-empty-input difference, already recorded as a deliberate divergence in `aggregate_count_null_test.rs`.

### DIFF
--- a/crates/rustledger-query/src/executor/operators.rs
+++ b/crates/rustledger-query/src/executor/operators.rs
@@ -31,9 +31,9 @@ impl Executor<'_> {
     ) -> Result<Value, QueryError> {
         let left = self.evaluate_expr(&op.left, ctx)?;
         let right = self.evaluate_expr(&op.right, ctx)?;
-        // `binary_op_on_values` owns the operator semantics — including the SQL
-        // three-valued-logic NULL short-circuit (a comparison involving NULL is
-        // UNKNOWN → false, so WHERE drops rows with a missing optional field).
+        // `binary_op_on_values` owns the operator semantics — including the
+        // NULL-comparison rule (a comparison involving NULL yields NULL, which
+        // is falsy, so WHERE still drops rows with a missing optional field).
         self.binary_op_on_values(op.op, &left, &right)
     }
 
@@ -258,9 +258,13 @@ impl Executor<'_> {
         left: &Value,
         right: &Value,
     ) -> Result<Value, QueryError> {
-        // Same NULL-comparison rule as `evaluate_binary_op` (see there).
+        // A comparison with a NULL operand is NULL, not FALSE (#2213). The two
+        // are not interchangeable: `COUNT` skips NULLs but counts every FALSE,
+        // so `count(payee != '')` over payee-less rows answered 4 where it
+        // should answer 0. Filtering is unaffected -- `to_bool(NULL)` is false,
+        // so `WHERE` still drops the row.
         if is_comparison(op) && (matches!(left, Value::Null) || matches!(right, Value::Null)) {
-            return Ok(Value::Boolean(false));
+            return Ok(Value::Null);
         }
         match op {
             BinaryOperator::Eq => Ok(Value::Boolean(self.values_equal(left, right))),
--- a/crates/rustledger-query/src/executor/operators.rs
+++ b/crates/rustledger-query/src/executor/operators.rs
@@ -10,7 +10,21 @@ use super::types::{Interval, PostingContext, Value};
 
 /// Whether `op` is an equality or ordering comparison — the operators for which
 /// a NULL operand yields SQL "UNKNOWN" (treated as not-matched).
-const fn is_comparison(op: BinaryOperator) -> bool {
+/// Operators for which a NULL operand makes the whole expression NULL (#2213).
+///
+/// The equality and ordering operators, plus regex matching and set
+/// membership. Verified against beanquery 0.2.0: on a payee-less posting,
+/// `payee ~ 'x'`, `payee !~ 'x'` and `payee IN ('a')` are each NULL there, not
+/// the FALSE/TRUE/FALSE this returned.
+///
+/// `IS NULL` / `IS NOT NULL` are excluded on purpose -- they are null TESTS,
+/// and must answer TRUE/FALSE precisely when the operand is NULL. `AND` and
+/// `OR` are excluded too: they coerce, as `NOT` does.
+///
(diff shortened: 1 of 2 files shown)


robcohen:
Good catch, and it was more than a wording problem — the fix was incomplete.

The comment was imprecise because the guard covered only the six equality/ordering operators. Rather than just narrow the wording, I checked what `~`, `!~`, `IN` and `NOT IN` actually do against beanquery 0.2.0 on a payee-less posting:

| expression | bean-query | before | after |
|---|---|---|---|
| `payee ~ 'x'` | *(null)* | `FALSE` | *(null)* |
| `payee !~ 'x'` | *(null)* | `TRUE` | *(null)* |
| `payee IN ('a')` | *(null)* | `FALSE` | *(null)* |
| `payee NOT IN ('a')` | *(null)* | `TRUE` | *(null)* |

All four were the same #2213 bug. Both regex arms carried a comment asserting the old answer "matches Python beancount behavior" — it does not, and the oracle says so directly. Those arms are now unreachable and the false claim is gone.

`is_comparison` is renamed `propagates_null`, which is what it now decides, and it documents why `IS NULL`/`IS NOT NULL` (null *tests*) and `AND`/`OR` (coercing, like `NOT`) stay out.

**The risk this opened.** An untagged posting reads as "missing", so it would be easy to propagate NULL from `tags` as well — and that would look like a fix while breaking `'food' IN tags`, a common query shape. It is an empty **set**, not NULL (`tags IS NULL` is FALSE in both tools), so the guard doesn't fire and the answer stays FALSE, matching bean-query. `membership_in_an_empty_set_is_false_not_null` pins it, and mutating the guard to treat an empty set as NULL makes it fail.

Re-verified: 4688 workspace tests green, clippy clean on 1.97.0, and the BQL compat harness over 71 files / 1349 runs shows no regression against baseline.

PR #2226, opened 2026-09-01
feat(plugin-types): add ParialEq and Eq

### LINKED ISSUE #2224: [Feature]: Add PartialEq and Eq for DirectiveData and the nested types
### Problem Statement

When writing plugins or importers it would be nice to compare expected types directly in tests, instead of comparing each field manually:

```rust
        assert_eq!(transaction.flag, "*");
        assert_eq!(transaction.payee, None);
        assert_eq!(transaction.narration, "TEST ETF");
        assert!(transaction.tags.is_empty());
        assert!(transaction.links.is_empty());
        assert!(transaction.metadata.is_empty());
        assert_eq!(transaction.postings.len(), 2);
```

### Proposed Solution

Add `#[derive(ParialEq, Eq)]`to all the types in `rustledger-plugin-types`. Could be gated behind test flag.

### Alternatives Considered

_No response_

### Area

### Beancount Compatibility

NikodemMarek:
## Summary

Add `PartialEq` and `Eq` for `rustledger-plugin-types`. This is useful when writing tests for custom plugins.

## Type of Change

## Changes

Add `PartialEq` and `Eq` for `rustledger-plugin-types`.

## Testing

- [x] `cargo test` passes
- [x] `cargo clippy` passes

## Checklist

## Review Focus

- [x] General review is fine

## Related Issues

Closes #2224 

### For Reviewers

### DIFF
--- a/crates/rustledger-plugin-types/src/lib.rs
+++ b/crates/rustledger-plugin-types/src/lib.rs
@@ -491,8 +491,17 @@ pub struct DirectiveWrapper {
 }
 
 impl PartialEq for DirectiveWrapper {
-    fn eq(&self, other: &Self) -> bool {
-        self.date == other.date && self.data == other.data
+    fn eq(
+        &self,
+        Self {
+            directive_type: _,
+            date,
+            filename: _,
+            lineno: _,
+            data,
+        }: &Self,
+    ) -> bool {
+        self.date == *date && self.data == *data
     }
 }
 
@@ -645,13 +654,24 @@ pub struct PostingData {
 }
 
 impl PartialEq for PostingData {
-    fn eq(&self, other: &Self) -> bool {
-        self.account == other.account
-            && self.units == other.units
-            && self.cost == other.cost
-            && self.price == other.price
-            && self.flag == other.flag
-            && self.metadata == other.metadata
+    fn eq(
+        &self,
+        Self {
+            account,
+            units,
+            cost,
+            price,
+            flag,
+            metadata,
+            span: _,
+        }: &Self,
+    ) -> bool {
+        self.account == *account
+            && self.units == *units
+            && self.cost == *cost
+            && self.price == *price
+            && self.flag == *flag
+            && self.metadata == *metadata
     }
 }
 
@@ -1398,7 +1418,10 @@ mod tests {
 
         // Postings with different or missing spans must compare equal
         assert_eq!(posting_a, posting_b, "different spans must compare equal");
-        assert_eq!(posting_a, posting_none, "Some vs None span must compare equal");
+        assert_eq!(
+            posting_a, posting_none,
+            "Some vs None span must compare equal"
+        );
 
         // Differing in an actual field (e.g. account) must compare not equal
         let posting_diff = PostingData {
@@ -1413,13 +1436,16 @@ mod tests {
             metadata: vec![],
             span: Some(span_a),
         };
-        assert_ne!(posting_a, posting_diff, "differing in account must compare not equal");
+        assert_ne!(
+            posting_a, posting_diff,
+            "differing in account must compare not equal"
+        );
     }
 
     #[test]
     fn test_partial_eq_both_directions() {
         let dir1 = DirectiveWrapper {
-            directive_type: "transaction".to_string(),
+            directive_type: "open".to_string(),
             date: "2024-01-01".to_string(),
             filename: Some("main.beancount".to_string()),
             lineno: Some(10),
@@ -1431,6 +1457,7 @@ mod tests {
             }),
         };
         let dir2 = DirectiveWrapper {
+            // Incorrect, to verify the PartialEq behavior.
             directive_type: "transaction".to_string(),
             date: "2024-01-01".to_string(),
             filename: Some("other.beancount".to_string()),
@@ -1443,7 +1470,7 @@ mod tests {
             }),
         };
         let dir3 = DirectiveWrapper {
-            directive_type: "transaction".to_string(),
+            directive_type: "open".to_string(),
             date: "2024-01-01".to_string(),
             filename: Some("main.beancount".to_string()),
             lineno: Some(10),
--- a/crates/rustledger-plugin-types/src/lib.rs
+++ b/crates/rustledger-plugin-types/src/lib.rs
@@ -1289,6 +1289,94 @@ mod tests {
     }
 
     #[test]
+    fn test_directive_wrapper_ignores_filename_and_lineno_in_partial_eq() {
+        let dir_a = DirectiveWrapper {
+            directive_type: "transaction".to_string(),
+            date: "2024-01-01".to_string(),
+            filename: Some("file_a.beancount".to_string()),
+            lineno: Some(10),
+            data: DirectiveData::Open(OpenData {
+                account: "Assets:Bank".to_string(),
+                currencies: vec!["USD".to_string()],
+                booking: None,
+                metadata: vec![],
+            }),
+        };
+
+        let dir_b = DirectiveWrapper {
+            directive_type: "transaction".to_string(),
+            date: "2024-01-01".to_string(),
+            filename: Some("file_b.beancount".to_string()),
+            lineno: Some(999),
+            data: DirectiveData::Open(OpenData {
+                account: "Assets:Bank".to_string(),
+                currencies: vec!["USD".to_string()],
+                booking: None,
+                metadata: vec![],
+            }),
+        };
+
+        let dir_none = DirectiveWrapper {
+            directive_type: "transaction".to_string(),
+            date: "2024-01-01".to_string(),
+            filename: None,
+            lineno: None,
+            data: DirectiveData::Open(OpenData {
+                account: "Assets:Bank".to_string(),
+                currencies: vec!["USD".to_string()],
+                booking: None,
+                metadata: vec![],
+            }),
+        };
+
+        assert_eq!(dir_a, dir_b);
+        assert_eq!(dir_a, dir_none);
+    }
+
+    #[test]
+    fn test_partial_eq_both_directions() {
+        let dir1 = DirectiveWrapper {
+            directive_type: "transaction".to_string(),
+            date: "2024-01-01".to_string(),
+            filename: Some("main.beancount".to_string()),
+            lineno: Some(10),
+            data: DirectiveData::Open(OpenData {
+                account: "Assets:Bank".to_string(),
+                currencies: vec!["USD".to_string()],
+                booking: None,
+                metadata: vec![],
+            }),
+        };
+        let dir2 = DirectiveWrapper {
+            directive_type: "transaction".to_string(),
+            date: "2024-01-01".to_string(),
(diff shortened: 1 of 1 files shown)


robcohen:
Thanks — this is a reasonable ask, and the change itself looks right. I built it locally: `rustledger-plugin-types` compiles, its tests pass, and clippy is clean on 1.97.0 (the toolchain CI uses). `SourceSpan` already derived `PartialEq`/`Eq`, so nothing else needed touching.

Three things before it can go in.

**1. The description doesn't match the code.** It says the derives are "gated behind `test` profile", but they're unconditional. Unconditional is the *correct* choice here — a `cfg(test)` gate would only apply while compiling this crate's own tests, so it would do nothing for a plugin author writing tests in their own crate, which is the stated use case. Could you update the description to match? Worth being explicit since it makes these part of the public API.

**2. Equality includes source position.** `DirectiveWrapper` carries `filename` and `lineno`, so a derived `PartialEq` compares those too. Two structurally identical directives parsed from different lines will not be equal. That may well be what you want, but for the "assert my plugin produced this directive" case it's usually the surprise — comparing `.data` sidesteps it. Worth a line in the doc comment either way, so the next person meets the behavior rather than debugging it.

**3. A test.** A derive that silently stops applying is exactly the kind of thing that goes unnoticed, so please add a small one — and make it cover both directions. An assertion that two identical values are equal passes even if the comparison were degenerate; pair it with two values differing in one field asserting they are *not* equal.

**What's actually blocking the merge** is the CLA — `license/cla` is the only failing check. Signing it via the bot's link on this PR should clear it, and I can merge once CI is green.

robcohen:
Correction to my last comment: the CLA is **not** blocking — `license/cla` is green. I misread it, because commit statuses report under `state` rather than `conclusion` and my query only looked at the latter.

The actual blocker is that CI had never run. Every workflow was sitting at `action_required`, which is what GitHub does for a first-time fork contributor, so the two required checks (`build`, `Python-gated cargo tests`) simply did not exist on the PR. I've approved the 11 pending runs, so they're going now.

The three review points stand: the description's "gated behind `test` profile" doesn't match the unconditional derives (unconditional being the right call), equality including `filename`/`lineno` is worth a doc note, and a test covering both the equal and not-equal directions.

NikodemMarek:
1. I forgot to fix the description after removing the flag, it is now corrected.

2. I think it makes sense to not check `filename` and `lineno` for equality in `DirectiveWrapper`, since as you have said, it probably is not that useful for the users. I have added a test for this and a short clarification at the struct itself.

3. I have added two tests, one for the top-level `DirectiveWrapper`, since the whole crate wont build if any derive is missing, and the second one as in 2.

Thank you for a quick and comprehensive response.

robcohen:
Thanks for turning these around — the description now matches the code, the source-position exclusion is in, and the tests cover both directions. Reviewed the new revision in depth; findings and CI status below, kept in this one comment.

## 1. Blocking: a wrapper does not equal itself after crossing the plugin boundary

`directive_type` is `#[serde(skip_serializing, default)]`, so it is never written to the wire and always deserializes to `""`. Including it in `PartialEq` makes a round-tripped directive unequal to the one that was sent. Checked against the MessagePack format the boundary actually uses, not JSON:

```rust
let bytes = rmp_serde::to_vec(&w)?;                    // w.directive_type == "open"
let back: DirectiveWrapper = rmp_serde::from_slice(&bytes)?;
// back.directive_type == ""
// w == back          -> false
```

That defeats the PR's stated purpose: comparing what a plugin produced — which arrived over the wire — against a locally-built expected value fails for a reason unrelated to the data.

**Fix: drop `directive_type` from `eq`.** Its own doc says it is "derived from data", so `self.data == other.data` already decides it; comparing it adds nothing and costs reflexivity across serialization.

Two things support that reading. Your tests set `directive_type: "transaction"` while the payload is `DirectiveData::Open(..)` — inconsistent, and nothing noticed, because the field is not load-bearing. And the crate's existing `test_directive_sort_order` builds wrappers with `String::new()`.

## 2. Decision: `PostingData.span` reintroduces position sensitivity one level down

`PostingData` carries `span: Option<SourceSpan>`, and `SourceSpan` already derived `PartialEq`, so posting equality includes byte offsets and a file id:

```
posting(span 0..10, file 0) == posting(span 100..110, file 3)  -> false
posting(span 0..10, file 0) == posting(span None)              -> false
```

Same class of problem you just fixed on the wrapper, and it bites the main case of comparing transactions. Its doc says plugins "must preserve this unchanged when modifying an existing posting; set to `None` only for postings the plugin itself synthesizes" — so a synthesized posting can never equal a parsed one, and an expected value would need exact byte offsets to match.

I'd exclude `span` for the same reason you excluded `filename`/`lineno`, but it is a judgement call about what equality means for these types, so I'd rather you make it.

## 3. Metadata equality is order-sensitive, over a source that has no order

`metadata` is `Vec<(String, MetaValueData)>`, so derived `PartialEq` compares it positionally. It is built by iterating `Metadata` — `FxHashMap<String, MetaValue>` (`rustledger-core/src/directive.rs:83`) — and `convert/to_wrapper.rs` collects it without sorting:

```rust
metadata: open.meta.iter().map(|(k, v)| (k.clone(), meta_value_to_data(v))).collect(),
```

So two directives with identical metadata compare unequal if the keys were inserted in a different order. Through the real `directive_to_wrapper` path, same six keys and values, inserted forwards and backwards:

```
order-a: ["epsilon", "zeta", "delta", "alpha", "gamma", "beta"]
order-b: ["beta", "alpha", "delta", "gamma", "zeta", "epsilon"]
equal:   false
```

`FxHash` is not randomized, so this is not run-to-run flakiness — it is silently reproducible, which looks like a real difference rather than an artifact.

**Fixed in #2231** (merged), which sorts metadata by key in `to_wrapper`. Nothing needed from you on this one — once that lands, identical metadata has one wire representation and this stops being a way for equal directives to compare unequal.

## Minor

- `PartialEq` is hand-written while `Eq` is derived. Nothing implements `Hash` on `DirectiveWrapper` today, so there is no bug — but a later `#[derive(Hash)]` would hash the excluded fields and silently break the Hash/Eq contract. A line on the impl saying which fields are skipped and why would stop that.
- Neither test varies `date`, so an `eq` that dropped it would still pass. Cheap to add to the existing `assert_ne!` case.

## CI

Both required checks pass: `build` and `Python-gated cargo tests` are green. Three reds, none blocking:

- **`Validate branch name`** — this PR is from your fork's `main`, and the repo wants `<type>/<description>`. Not in the required set. **Do not rename the branch to fix it** — renaming the head branch closes the PR. It would take a fresh PR from a named branch; otherwise it can be left, since the squash-merge commit carries the conventional-commit message regardless.
- **`Cargo Deny`** and **`quality`** — pre-existing failures on `main`, unrelated to your change.

## Summary

All three findings are one defect in different clothes: equality over incidental representation. (3) is handled in #2231. That leaves **(1) as the only thing blocking**, and (2) as your call — both small. Fixing (1) is enough for me to merge.

robcohen:
CI status now that the runs have finished: both required checks pass — `build` and `Python-gated cargo tests` are green.

Three red checks, none of them blocking, and only one is about this PR:

- **`Validate branch name`** — this PR is from your fork's `main`, and the repo requires `<type>/<description>` (`feat/…`, `fix/…`, …). The check is not in the required set, so it does not block the merge. **Do not rename the branch to fix it** — renaming the head branch closes the PR. If you want it green, it would take a fresh PR from a named branch; otherwise it can be left, since the squash-merge commit carries the conventional-commit message regardless.
- **`Cargo Deny`** and **`quality`** — pre-existing failures on `main`, unrelated to your change.

So the only thing standing between this and a merge is the equality question in my previous comment: `directive_type` in `eq` makes a wrapper unequal to itself after a round-trip through the plugin wire, which is the case the PR exists to serve. The `PostingData.span` point is yours to decide.

robcohen:
Second pass, looking at areas the first one didn't reach. One new finding, and it's the same shape as the two already on the table: equality comparing representation rather than meaning.

## Metadata equality is order-sensitive, over a source that has no order

`metadata` is `Vec<(String, MetaValueData)>`, so the derived `PartialEq` compares it positionally. But it is built by iterating `Metadata`, which is `FxHashMap<String, MetaValue>` (`rustledger-core/src/directive.rs:83`), and `convert/to_wrapper.rs` collects it without sorting:

```rust
metadata: open.meta.iter().map(|(k, v)| (k.clone(), meta_value_to_data(v))).collect(),
```

So two directives with *identical* metadata compare unequal if the keys were inserted in a different order. Through the real `directive_to_wrapper` path, same six keys and same values, inserted forwards and backwards:

```
order-a: ["epsilon", "zeta", "delta", "alpha", "gamma", "beta"]
order-b: ["beta", "alpha", "delta", "gamma", "zeta", "epsilon"]
equal:   false
```

This is the plugin-test case directly: an expected value built by hand versus an actual built by the parser. Nothing about the data differs.

Worth being precise about the failure mode — `FxHash` is not randomized, so this is not run-to-run flakiness. It is silently reproducible, which is worse in one way: it looks like a real difference rather than an artifact.

I don't think fixing this belongs in your PR. The clean fix is to sort by key in `to_wrapper`, which makes the wire deterministic and has value on its own (reproducible MessagePack, stable hashing) — and it costs nothing, because the source is a hash map, so there is no authored ordering being preserved today. That is a separate change to `rustledger-plugin`, and I'm happy to make it.

What your PR does need is a decision: either wait for that, or land with the limitation documented on the impl, so the next person meets it rather than debugging it.

## Where that leaves the three

All three are the same defect in different clothes — equality over incidental representation:

1. `directive_type` — never survives the wire, so a round-tripped wrapper is unequal to itself. **Blocking**, and the fix is to drop the field from `eq`.
2. `PostingData.span` — byte offsets and file id decide posting equality. Your call.
3. `metadata` order — above.

Fixing (1) is enough to merge as far as I'm concerned; (2) and (3) can be follow-ups if you'd rather keep this small. Say which you'd prefer and I'll either do the `to_wrapper` sort now or open an issue for it.

NikodemMarek:
@robcohen was the close accidental, and can I finish and merge this still? or was it intentional?

NikodemMarek:
I've fixed the problems and also renamed the branch, new pr is here: #2240

robcohen:
Yes this was a mistake, I was still tracking it though.

robcohen:
This was closed by accident, and the fault is mine — nothing about your PR was rejected.

My PR #2231 ended its description with "Does **not** close #2226; it removes one of the three findings blocking it." GitHub's closing-keyword parser matched `close #2226` in that sentence and does no negation analysis, so merging #2231 auto-closed this. A sentence written to say the opposite is what did it.

Reopening now. Your work stands as reviewed: the only blocking item is still `directive_type` in `eq`, and the `PostingData.span` question remains yours to decide. The metadata-ordering finding is genuinely resolved by #2231, which is what that sentence was trying to say.

Sorry for the noise.

NikodemMarek:
Worked out for the best, the new pr: #2224 has a correct branch name :) this can be closed again

PR #2229, opened 2026-09-02
fix(query): a string-literal header echoes the quote the query wrote

### LINKED ISSUE #2176: BQL: string-literal headers do not preserve the query's quote style

robcohen:
`SELECT "dq"` headed `'dq'`. bean-query names such a column by reproducing the literal as typed, quote character included. Closes #2176.

`Literal::String` did not record which quote the source used, so the header had to guess — single quotes unless the value contained one. It now carries a `QuotedString`, and the header echoes what was written. Verified against beanquery 0.2.0:

| query | bean-query | before | after |
|---|---|---|---|
| `SELECT 'plain'` | `'plain'` | `'plain'` | `'plain'` |
| `SELECT "dq"` | `"dq"` | `'dq'` | `"dq"` |
| `SELECT "o'clock"` | `"o'clock"` | `"o'clock"` | `"o'clock"` |
| `SELECT 'it''s'` | `'it''s'` | `"it''s"` | `'it''s'` |
| `SELECT 'say "hi"'` | `'say "hi"'` | `'say "hi"'` | `'say "hi"'` |
| `SELECT length('abc')` | `length('abc')` | `length('abc')` | `length('abc')` |

The body is reproduced **verbatim**, as bean-query does: `'it''s'` heads `'it''s'`, the doubled quote left as written rather than re-escaped or collapsed. That is consistent with the value semantics already pinned in `string_literal_escapes_test.rs`, where `''` stays in the value.

## Provenance is not part of the value

`QuotedString` implements `PartialEq` and `Hash` on the text alone. `'x'` and `"x"` are the same string, and AST equality must not depend on typing style — same reasoning as `Spanned`, which keeps equality on its value rather than its span. The test pins the Hash/Eq agreement too, since a `HashMap` keyed on literals would split otherwise.

`quote` is `None` for a literal built in code rather than parsed, which has no style to echo. Those keep the #2175 fallback (single quotes, switching to double when the value contains a single quote and no double), so that fix is untouched and still covered.

`Display` echoes the source quote as well, which makes a parsed literal round-trip as written — the issue notes this may matter more than the header does — and still falls back to double quotes when there is no style.

## Parser

`string_literal` now delegates to the new `quoted_string_literal` rather than restating the grammar. Writing the style-carrying version alongside the old one would have duplicated both quote arms, and a fix applied to one and not the other would make a literal parse differently depending on which caller reached it. Two callers want only the text and stay on the plain wrapper.

## Verification

Header and `Display` are separately controlled: making either ignore the recorded style fails tests (3 integration + 1 unit for the header, 1 for `Display`), and the equality test fails if `PartialEq` starts comparing the style. The fallback rule has its own unit test, placed there because `header_name` is `pub(crate)` and a synthetic literal cannot be reached through `parse` — everything parsed carries a style.

4703 workspace tests green, clippy clean on 1.97.0, typos clean. BQL compat harness over 71 files / 1349 runs: no regression against baseline. The corpus does not exercise string-literal targets, so the oracle comparisons above are what carries this.

### DIFF
--- a/crates/rustledger-query/src/ast.rs
+++ b/crates/rustledger-query/src/ast.rs
@@ -215,11 +215,99 @@ pub enum Expr {
     },
 }
 
+/// Which quote character a string literal was written with.
+///
+/// bean-query echoes the quote the query wrote -- `"dq"` heads `"dq"` and
+/// `'dq'` heads `'dq'` -- so reproducing its headers means carrying the style
+/// from the source (#2176).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum QuoteStyle {
+    /// `'single'`.
+    Single,
+    /// `"double"`.
+    Double,
+}
+
+impl QuoteStyle {
+    /// The character itself.
+    #[must_use]
+    pub const fn char(self) -> char {
+        match self {
+            Self::Single => '\'',
+            Self::Double => '"',
+        }
+    }
+}
+
+/// A string literal, and the quote character its source used if it came from
+/// one.
+///
+/// # Provenance is not part of the value
+///
+/// `PartialEq` and `Hash` consider only [`Self::value`]. `'x'` and `"x"` are
+/// the same string, and an AST comparison that said otherwise would make
+/// query equality depend on typing style. The quote is carried solely so
+/// headers and [`Display`](std::fmt::Display) can echo what was written --
+/// the same reason [`Spanned`](rustledger_core::Spanned) keeps equality on
+/// its value rather than its span.
+///
+/// `quote` is `None` for a literal built programmatically rather than parsed
+/// (see [`Expr::string`]), which has no source style to echo. Those render by
+/// the fallback rule in `header_fragment`.
+#[derive(Debug, Clone, Eq)]
+pub struct QuotedString {
+    value: String,
+    quote: Option<QuoteStyle>,
+}
+
+impl QuotedString {
+    /// A literal read from source, carrying the quote it was written with.
+    #[must_use]
+    pub const fn parsed(value: String, quote: QuoteStyle) -> Self {
+        Self {
+            value,
+            quote: Some(quote),
+        }
+    }
+
+    /// A literal with no source style, built programmatically.
+    pub fn synthetic(value: impl Into<String>) -> Self {
+        Self {
+            value: value.into(),
+            quote: None,
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


robcohen:
Checked this rather than adjusting it, and the formatting is rustfmt's own output — `cargo fmt --check -p rustledger-query` passes on the branch as committed.

I tried the suggested one-line signature to be sure:

```
$ # signature rewritten onto one line
$ cargo fmt --check -p rustledger-query ; echo $?
1
$ cargo fmt -p rustledger-query          # rustfmt puts it straight back
fn quoted_string_literal<'a>()
-> impl Parser<'a, ParserInput<'a>, QuotedString, ParserExtra<'a>> + Clone {
```

So the leading `->` is required here, not a missed `cargo fmt`. It differs from the neighbouring parsers because this one takes **no parameters**: those signatures break after their argument list and so end `) -> impl Parser<...>`, while a zero-argument signature too long for one line can only break before the `->`. `parse_expr_with` at line 890 and its neighbours are the parameterised shape.

Leaving it as rustfmt writes it — changing it would fail the formatting gate in CI. Commits on this repo run rustfmt through the `prek` pre-commit hook, which is what produced this form in the first place (it rejected my first commit attempt and reformatted the hunk).

PR #2233, opened 2026-09-02
fix(query): PIVOT BY names its columns the way bean-query does

### LINKED ISSUE #2217: BQL: pivot column headers do not match bean-query (key column and value/spread order)

robcohen:
Three differences, all in the headers — the rows already agreed. Closes #2217.

| | key column | value columns |
|---|---|---|
| bean-query | `account/year` | `2024/sum(number)`, `2024/count(number)` |
| before | `account` | `sum(number) / 2024`, `count(number) / 2024` |
| after | `account/year` | `2024/sum(number)`, `2024/count(number)` |

## Why this one is matched when #2221 and #2222 are not

Same test in each case: **did upstream choose the result, or is it an artifact?**

- **#2221** (`BALANCES`/`JOURNAL` names) — `transform_journal` parses a SQL string template with no `AS` aliases, so `MAXWIDTH(payee, 48)` falls through to the generic expression-as-name path. Never chosen as a column name.
- **#2222** (inventory slots) — the padded per-currency layout comes from `class InventoryRenderer(ColumnRenderer)`. The API returns a plain `Inventory`, no slots.
- **#2217** (this) — names are built in `EvalPivot.__call__` (`query_compile.py`), in the **evaluator**, and survive to the API:

```python
[d[0] for d in conn.execute(q).description]
# ['account/year', '2024/sum(number)', '2024/count(number)', '2025/sum(number)', '2025/count(number)']
```

Reproducing the first two would mean reproducing an artifact. This one is part of the result's structure.

It is also the better answer on its own terms, which is why it is a fix rather than a concession: the key column genuinely spans two source columns, and columns are emitted **grouped by pivot value** (`product(keys, other(columns))`, and our loop matches) — so leading the name with the pivot value describes the layout the right way round. The old form described it backwards. The spaces were gratuitous.

## A branch the issue did not record

beanquery composes a two-part name only in its `nother > 1` branch. With exactly one non-pivot column it emits the bare pivot value:

```
API: ['account/year', '2024', '2025']
```

We already agreed there — only the key column was wrong. That branch is untouched and now pinned, because a fix that collapsed the two branches would have broken the case that was already right.

## Verification

Both shapes now match bean-query exactly, character for character.

Four new tests, each with a distinct negative control — reverting the key-column name fails 3, reverting the value-column form fails 2, and collapsing the single-column branch fails 1. Separate tests for the key column and for the spacing, because a fix that only reversed the halves would still have been wrong.

Five existing tests pinned the old names; every one was a header assertion with no row change. All updated, and one is now stronger (`columns[0]` compared exactly rather than searched for).

4706 workspace tests green, clippy clean on 1.97.0, typos clean. BQL compat harness over 71 files / 1349 runs: no regression against baseline.

### DIFF
--- a/crates/rustledger-query/src/executor/sort.rs
+++ b/crates/rustledger-query/src/executor/sort.rs
@@ -254,23 +254,47 @@ impl Executor<'_> {
 
         // Build new column names. Layout: [key_col, <value_col × pivot_value>...].
         //
-        // Single-value-column case (the typical SUM(number) shape) gets
-        // just the pivot value as the header — matches bean-query
-        // exactly. Multi-value-column case qualifies with the value
-        // column name (`<value_col> / <pivot_value>`) so the headers
-        // stay unambiguous when there's more than one aggregate. The
-        // asymmetry is deliberate: a single-aggregate header like
-        // `USD` is what users expect from the typical pivot.
+        // Naming follows bean-query's `EvalPivot` (`query_compile.py`), which
+        // sets these names in the EVALUATOR and exposes them through its API —
+        // they are part of the result's structure, not something its renderer
+        // decorates. That is why this one is matched where the `BALANCES` /
+        // `JOURNAL` names (#2221) and the inventory slot layout (#2222) are
+        // deliberately not: those are a name-less query template and a column
+        // renderer respectively, so reproducing them would mean reproducing an
+        // artifact.
+        //
+        // Each of the three parts is also the better answer on its own terms
+        // (#2217):
+        //
+        // * The key column is `<key>/<spread>`, e.g. `account/year`. The
+        //   pivot key genuinely spans two source columns; naming it after
+        //   only the first drops the spread column's name.
+        // * A value column is `<pivot_value>/<value_col>`, e.g.
+        //   `2024/sum(number)`. Columns are emitted grouped by pivot value
+        //   (the loop below is pivot-value-outer, matching bean-query's
+        //   `product(keys, other(columns))`), so leading with the pivot value
+        //   is what the column ORDER already does. Naming it the other way
+        //   round described the layout backwards.
+        // * No spaces around the separator.
+        //
+        // Single-value-column case (the typical SUM(number) shape) keeps just
+        // the pivot value as the header, which bean-query also does — its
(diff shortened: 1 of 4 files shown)


robcohen:
Good catch — corrected in 4c8acf9. That query is `PIVOT BY currency, account`, so `currency` is the key and the pivot values are account names.

Worth noting the comment I replaced had the same axis wrong in the other direction (it called the result account-keyed), and I corrected one half while leaving `<ccy>` in the other, which is worse than either. It now names both explicitly: a `currency/account` key column, then `<account>/SUM(number)`, `<account>/COUNT(*)` per account.

PR #2237, opened 2026-09-02
fix(query): grouped rows keep first-appearance order on every path

### LINKED ISSUE #2235: BQL: the two aggregate pipelines order groups differently, so LIMIT without ORDER BY keeps different rows
Found writing the pipeline-order tests for #2219. Pre-existing on `main`, and unrelated to that issue's bean-query comparison — this is our two code paths disagreeing with each other.

The same query, with and without `FROM #postings`, keeps **different rows** under the same `LIMIT`:

```
SELECT account, year, sum(number) [FROM #postings]
  GROUP BY account, year PIVOT BY account, year LIMIT 2

with FROM:     Assets:A, Equity:O
without FROM:  Assets:A, Assets:B
```

Fixture: `Assets:A` posts only in 2024, `Assets:B` only in 2025, `Equity:O` in both.

## Why

The pipeline is written out at three sites in `execution.rs`:

- `execute_select` (no `FROM`)
- `execute_select_from_table` — cannot reach a pivot, since `PIVOT BY` requires `GROUP BY`, which routes elsewhere
- `finish_aggregate_result` (the `FROM` + `GROUP BY` path)

The two reachable ones produce groups in a different order, so a `LIMIT` with no `ORDER BY` truncates a differently-ordered set. The **column set** agrees — that is what `pivot_output_is_the_same_with_and_without_from` pins (#2216) — so only row identity differs, which is why it went unnoticed.

## Why it matters even though the order is unspecified

Strictly, `LIMIT` without `ORDER BY` has no defined row order in SQL, so neither answer is *wrong*. But two paths through the same engine giving different rows for one query is a defect regardless: it is not reproducible from the query text, and which path runs depends on whether the user wrote `FROM #postings` — a clause that should not change results.

It also makes `LIMIT` a silent data-selection difference rather than a truncation, which is the same class of surprise as the pivot-drops-a-column behavior discussed in #2219.

## Not asserted in the new tests

`pivot_pipeline_order_test.rs` names an explicit `ORDER BY` in every cross-path case for this reason. Pinning today's answer would pin a bug. The column-set agreement is asserted, since that part is specified.

## Suggested resolution

Make the group order deterministic and identical across both paths — most likely by ordering groups the same way in both, or by having one delegate to the other. Worth checking whether the difference is a `HashMap` iteration order, in which case it may also vary by key set rather than being stable.

robcohen:
The no-`FROM` path sorted grouped output by the first column when a query had no `ORDER BY`; the aggregate `FROM` path did not. So the same query returned differently-ordered rows depending on whether the user wrote `FROM #postings` — and with a `LIMIT`, that meant different **rows**. Closes #2235.

```
SELECT account, year, sum(number) [FROM #postings]
  GROUP BY account, year PIVOT BY account, year LIMIT 2

with FROM     Assets:A, Equity:O
without FROM  Assets:A, Assets:B
```

## The sort's justification was wrong

Its comment claimed it matched Python beancount "for deterministic output". Verified against beanquery 0.2.0 — it does not. bean-query returns groups in first-appearance order:

```
SELECT account, year, sum(number) GROUP BY account, year
  bean-query  Assets:A/2024, Equity:O/2024, Assets:B/2025, Equity:O/2025
  sorted      Assets:A/2024, Assets:B/2025, Equity:O/2024, Equity:O/2025
```

The `FROM` path already agreed with bean-query, so removing the sort makes the no-`FROM` path agree with **both**. I checked the implicit-`GROUP BY` case separately, since the branch covered that too and fixing one while breaking the other would have been easy — bean-query gives first-appearance order there as well, and gives the *same* answer with and without `FROM`, as it should.

Nothing is lost by dropping it. `group_postings` already returns groups in first-appearance order and posting order is file order, so the output was deterministic without the sort — which was the stated reason for having it.

## Test strengthening

`pivot_pipeline_order_test.rs` (from #2236, merged an hour ago) had to **exclude** the no-`ORDER BY` cases when it was written, because the two paths disagreed there and pinning either answer would have pinned this bug. Those cases are now asserted, along with the bare grouped order against bean-query's.

Negative control: reinstating the fallback sort fails the cross-path test.

4714 workspace tests green, clippy clean on 1.97.0. BQL compat harness over 71 files / 1349 runs: no regression against baseline.

### DIFF
--- a/crates/rustledger-query/src/executor/execution.rs
+++ b/crates/rustledger-query/src/executor/execution.rs
@@ -87,8 +87,6 @@ impl Executor<'_> {
             || query.group_by.is_some()
             || query.having.is_some();
 
-        // Track whether grouping is applied (explicit or implicit) for fallback sort
-
         if is_aggregate {
             // Determine GROUP BY expressions:
             // - If explicit GROUP BY is provided, use it
--- a/crates/rustledger-query/src/executor/execution.rs
+++ b/crates/rustledger-query/src/executor/execution.rs
@@ -9,8 +9,7 @@ use rustledger_core::{Amount, Directive, Inventory, NaiveDate, Position};
 const PARALLEL_THRESHOLD: usize = 1000;
 
 use crate::ast::{
-    CreateTableStmt, Expr, InsertSource, InsertStmt, OrderSpec, SelectQuery, SortDirection, Target,
-    UnaryOperator,
+    CreateTableStmt, Expr, InsertSource, InsertStmt, SelectQuery, Target, UnaryOperator,
 };
 use crate::error::QueryError;
 
@@ -89,7 +88,6 @@ impl Executor<'_> {
             || query.having.is_some();
 
         // Track whether grouping is applied (explicit or implicit) for fallback sort
-        let mut has_grouping = false;
 
         if is_aggregate {
             // Determine GROUP BY expressions:
@@ -106,9 +104,6 @@ impl Executor<'_> {
                     Some(implicit)
                 }
             };
-
-            // Track if grouping is applied for deterministic fallback sort
-            has_grouping = group_by_exprs.is_some();
 
             // Group and aggregate
             let grouped = self.group_postings(&postings, group_by_exprs.as_ref())?;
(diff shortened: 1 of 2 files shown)


robcohen:
Correct — fixed in 10993a5. I removed `has_grouping` and its assignment with the sort but left the comment that introduced them, so it described a variable that no longer exists and a branch that no longer runs.

Issue #2154, opened 2026-08-25
meta is missing from seven BQL tables, plus discrepancy/tolerance on #balances and links/tags on #notes


robcohen:
> **Scope note.** The five `#entries` columns this issue opened with (`year`, `month`, `day`, `description`, `meta`) shipped in #2157. What remains is 11 columns across seven tables, listed under "Corrected census" below, and `meta` accounts for seven of them. The title and this note were updated to match; the original text is kept intact so the reasoning that found them still reads in order.

The `#entries` table is missing five columns and `#transactions` is missing one. The names are all implemented already and work against `#postings`, so this is a schema registration gap rather than unimplemented functionality.

Companion to #2153, which found the same shape among BQL functions.

## `SELECT *` shows it directly

```
beanquery  SELECT * FROM #entries
  id type filename lineno date year month day flag payee narration description tags links meta accounts   (16)

rustledger SELECT * FROM #entries
  id type filename lineno date           flag payee narration             tags links      accounts   (11)
```

Missing from `#entries`: `year`, `month`, `day`, `description`, `meta`.

`#transactions` agrees on `SELECT *` (7 columns each), but `meta` is selectable explicitly under beanquery and not here:

```
$ rledger query t.beancount "SELECT meta FROM #transactions"
error: failed to execute query: evaluation error: column 'meta' not found
```

Each of the six was confirmed working under beanquery before being reported, for example `SELECT year FROM #entries` returns `2024` and `SELECT day FROM #entries` returns `1`.

## They are not unimplemented, just unregistered

Every one of these names resolves fine against `#postings` and against the default table:

| column | `#postings` | default (no FROM) | `#entries` |
|---|---|---|---|
| `year` | works (`2024`) | works | **not found** |
| `month` | works (`2`) | works | **not found** |
| `day` | works (`1`) | works | **not found** |
| `description` | works (`alpha`) | works | **not found** |
| `meta` | works | works | **not found** |

So the extraction logic exists and the values are available. The `#entries` and `#transactions` schemas just do not list them.

`meta` on `#entries` is cheaper still: the column is already **in that table**, under a private name. `build_entries_table` declares twelve columns, and the twelfth is `_entry_meta`:

```rust
// crates/rustledger-query/src/executor/system_tables.rs
let columns = vec![
    "id", "type", "filename", "lineno", "date", "flag",
    "payee", "narration", "tags", "links", "accounts",
    "_entry_meta",
];
```

[...]

Issue #1663, opened 2026-07-01
load/load-file do not report balance-assertion failures (only validate does)


robcohen:
## Summary

`ledger.load` (and `load-file`) do **not** report `balance` assertion failures, and emit no diff on the balance directive. Only `ledger.validate` reports them. Embedders that load via `load`/`load-file` (rustfava does) therefore surface **no error** for a failing `balance` — the user is told their books reconcile when they don't.

Beancount's `loader.load_file` runs balance checking as part of loading and returns the errors; rustledger's `load` should match so the primary load path is safe for embedders.

## Repro (v0.17.3, FFI component)

Ledger — real balance of `Assets:Cash` is `-5 USD`, but asserts `999 USD`:

```beancount
2024-01-01 open Assets:Cash USD
2024-01-01 open Expenses:X USD
2024-01-02 * "t"
  Expenses:X   5 USD
  Assets:Cash
2024-01-03 balance Assets:Cash   999 USD
```

- `load(src)` → `errors: []`, and the returned `balance` directive carries **no diff** (`{type: balance, date, account, amount: {999 USD}, tolerance: null, meta}`).
- `validate(src)` → correctly reports:
  `{ message: "Balance failed for Assets:Cash: expected 999 USD, got -5 USD", line: 6, severity: error, phase: validate }`

## Expected

`load`/`load-file` should:
1. Include balance-assertion failures in `errors` (same as `validate`), so embedders that don't separately call `validate` still see them; and
2. Ideally attach the computed vs asserted difference to the returned `balance` directive (a `diff-amount`/`diff` field), so UIs can render per-assertion pass/fail without re-deriving it.

If the load/validate split is intentional, please document that embedders **must** call `validate` for balance-check errors — but matching beancount's load-includes-balance-checks behavior is the safer default.

## Impact

In rustfava this means every failing `balance` directive renders green/"passed" in the journal (its red/green keys off the directive diff) and produces no error banner — a silent correctness failure on the feature users rely on to catch bookkeeping mistakes.

PR #2238, opened 2026-09-02
feat(query): #balances exposes the checker's discrepancy

### LINKED ISSUE #2180: BQL: #balances.discrepancy needs the checker's computed difference, not a schema entry

robcohen:
`SELECT * FROM #balances` now returns bean-query's fifth column. Closes #2180.

`discrepancy` is `computed − asserted` for a **failing** assertion and NULL otherwise. Verified against beanquery 0.2.0:

| asserted | accumulated | tolerance | discrepancy |
|---|---|---|---|
| 90.00 | 100.00 | — | `10.00 USD` |
| 120.00 | 100.00 | — | `-20.00 USD` |
| 100.01 | 100.00 | 0.05 | *(null)* |

That last row is the reason this couldn't be faked from the directive: a small non-zero difference **inside** the tolerance is a pass and reports nothing, so the value depends on the tolerance rule as much as on the arithmetic. beancount does the same — `ops/balance.py` sets `diff_amount` only when `abs(diff) > tolerance`, leaving it `None` otherwise.

## Plumbed, not recomputed

The issue's constraint was that re-deriving this in the query layer means a second implementation of balance accumulation. It's worse than that — it also means a second implementation of the *tolerance* rule.

So: the validator already recorded the signed difference per assertion (#1663). It now also records whether that difference exceeded the tolerance, taken from **the same comparison that raises the error**. The loader carries the failing ones onto `Ledger`, beside `capital_gains` and for the same stated reason — a consumer that re-derives either would drift from `rledger check`.

`rustledger-query` can't depend on `rustledger-validate` (checked: it's an optional dep of the loader only), so the executor receives the differences from its host via `set_balance_discrepancies` and reports NULL without them. **The column exists either way**, so a host that skips validation gets bean-query's schema rather than a missing column.

## The query command now validates

That's what computes them. Two things I measured rather than assumed:

- **Cost**: a query on the largest corpus file (18.6k lines) runs in ~0.09s with validation and ~0.12s without — load dominates, and the difference is noise.
- **Output**: diagnostics go to **stderr**, so stdout is byte-identical. bean-query likewise prints a failing assertion before returning its rows, so this moves us toward it in behavior as well as schema.

## Last gap in the census

This was the final item from #2154's census of all 55 columns bean-query exposes across its ten tables (`#notes.tags`/`.links` went with #2160). So `select_star_matches_bean_query_per_table` loses its per-table `gap` field — every expectation is now bean-query's set exactly, and a difference means drift rather than progress. That also fixes the type-inference break that removing the last `Some(...)` would otherwise have caused.

## Verification

Four CLI-level tests, because that's where the whole path runs — loader → checker → executor. Two negative controls, each failing a distinct set: reverting the query path to `validate: false` fails 3, and reporting the difference for passing assertions too fails the tolerance case.

One test pins that stdout stays free of diagnostics, since every consumer parses it.

4718 workspace tests green, clippy clean on 1.97.0, typos clean on all changed files. BQL compat harness over 71 files / 1349 runs: no regression against baseline.

### DIFF
--- a/crates/rustledger-loader/src/process.rs
+++ b/crates/rustledger-loader/src/process.rs
@@ -114,6 +114,30 @@ pub enum ProcessError {
     PluginConversion(String),
 }
 
+/// A balance assertion that FAILED, with the difference the checker computed.
+///
+/// One per failing assertion; a passing one produces no entry. That mirrors
+/// beancount, whose checker sets `diff_amount` only on a failing entry and
+/// leaves it `None` otherwise -- and it is why this cannot be derived from the
+/// directive: a small non-zero difference inside the tolerance is a pass.
+///
+/// Captured during the loader's validation pass for the same reason
+/// [`Ledger::capital_gains`] is: consumers read it rather than re-deriving the
+/// balance, so `rledger check` and `#balances.discrepancy` cannot disagree
+/// (#2180).
+///
+/// Empty when the loader ran without validation, in which case the column
+/// reports NULL rather than a wrong number.
+#[derive(Debug, Clone)]
+pub struct BalanceDiscrepancy {
+    /// The `balance` directive's date.
+    pub date: rustledger_core::NaiveDate,
+    /// The asserted account.
+    pub account: rustledger_core::Account,
+    /// `computed - asserted`, in the asserted currency.
+    pub amount: rustledger_core::Amount,
+}
+
 /// A fully processed ledger.
 ///
 /// This is the result of loading and processing a beancount file,
@@ -158,6 +182,11 @@ pub struct Ledger {
     /// capgains report — read these directly rather than re-booking the stream
     /// and re-deriving them, so they cannot drift from `rledger check`.
     pub capital_gains: Vec<rustledger_booking::CapitalGain>,
+    /// Failing balance assertions and the difference the checker computed,
+    /// one per failure. Empty when the loader ran without validation.
+    ///
+    /// Feeds `#balances.discrepancy` (#2180).
+    pub balance_discrepancies: Vec<BalanceDiscrepancy>,
 }
 
 impl Ledger {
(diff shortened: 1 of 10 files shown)


robcohen:
Real bug — reproduced and fixed in 10adaf6a.

`(date, account, currency)` does not identify an assertion. Two on one date for the same account and currency can assert different amounts and fail differently, and both rows took whichever difference was inserted last:

```
2024-03-01 balance Assets:Cash   90.00 USD
2024-03-01 balance Assets:Cash  120.00 USD

bean-query   10.00 USD, -20.00 USD
before      -20.00 USD, -20.00 USD
after        10.00 USD, -20.00 USD
```

The asserted number is now part of the key, recorded by the validator beside the difference. Two assertions agreeing on all four are the same assertion — same date, account, currency and amount means the same accumulated balance and the same computed difference — so collapsing those is harmless.

Pinned by `same_date_assertions_do_not_share_a_discrepancy`, which fails if the lookup ignores the asserted amount.

I checked whether the same keying appears elsewhere: `attach_balance_diffs` in the FFI (#1663) builds a `(date, account, currency)` map for the same data. It has the identical collision, and it predates this PR. Filing separately rather than widening the scope here — the FFI's contract also differs (it attaches a diff for passing assertions too), so the fix is not a copy of this one.

PR #2240, opened 2026-09-02
feat(plugin-types): add ParialEq and Eq

### LINKED ISSUE #2224: [Feature]: Add PartialEq and Eq for DirectiveData and the nested types

NikodemMarek:
## Summary

Add `PartialEq` and `Eq` for `rustledger-plugin-types`. This is useful when writing tests for custom plugins.

Reopened from #2226, with correct branch name.

## Type of Change

## Changes

Add `PartialEq` and `Eq` for `rustledger-plugin-types`.

## Testing

- [x] `cargo test` passes
- [x] `cargo clippy` passes

## Checklist

## Review Focus

- [x] General review is fine

## Related Issues

Closes #2224 

### For Reviewers

### DIFF
--- a/crates/rustledger-plugin-types/src/lib.rs
+++ b/crates/rustledger-plugin-types/src/lib.rs
@@ -467,6 +467,12 @@ impl PluginError {
 ///
 /// This wrapper provides a uniform interface for all directive types,
 /// with source location tracking for error reporting.
+///
+/// # Equality and Hashing Note
+/// The custom [`PartialEq`] implementation skips `directive_type`, `filename`, and `lineno`
+/// (location metadata and derived fields). If [`std::hash::Hash`] is ever implemented for
+/// `DirectiveWrapper`, those fields must also be excluded to satisfy the `Hash`/`Eq` invariant
+/// (`a == b => hash(a) == hash(b)`).
 #[derive(Debug, Clone, Eq, Serialize, Deserialize)]
 pub struct DirectiveWrapper {
     /// The type of directive (derived from data, not serialized to avoid duplicate keys).
@@ -490,9 +496,6 @@ pub struct DirectiveWrapper {
     pub data: DirectiveData,
 }
 
-// Skips `directive_type`, `filename`, and `lineno` (location metadata and derived field).
-// If `Hash` is ever implemented for `DirectiveWrapper`, those fields must also be excluded
-// to preserve the `Hash`/`Eq` invariant (`a == b => hash(a) == hash(b)`).
 impl PartialEq for DirectiveWrapper {
     fn eq(
         &self,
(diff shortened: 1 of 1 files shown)


robcohen:
Thanks for moving it to a named branch — that also clears `Validate branch name`, which #2226 could not fix without closing itself.

Reviewed this head. Both findings from #2226 are resolved, verified rather than taken on trust:

```
round-trip equal:              true    <- the blocking bug
differing date unequal:        true
differing data unequal:        true
span ignored:                  true
account still compared:        true
```

The destructuring is a better answer than the one I suggested. A new field on `DirectiveWrapper` or `PostingData` now fails to compile until someone decides whether it belongs in equality, which is exactly the failure mode that produced the `directive_type` bug — a field that looked like payload and silently was not.

Local: 28 tests pass in `rustledger-plugin-types`, clippy clean on 1.97.0 (the toolchain CI uses).

I've approved the workflow runs — fork PRs need that on each push, so CI had only 3 checks. Will merge once `build` and `Python-gated cargo tests` are green.

One optional follow-up, not blocking: nothing implements `Hash` on these types today, so the hand-written `PartialEq` breaks no contract — but a later `#[derive(Hash)]` would hash the excluded fields and silently violate Hash/Eq. A line on each impl naming what it skips would stop that. Happy to add it after merge if you'd rather not touch this again.

NikodemMarek:
I've added the hashing notes, I think this PR is ready for final review and a merge.

robcohen:
Merged as 62a87fbd — thanks, and sorry for the detour through #2226.

Final state verified on the merged head: the round-trip case that was blocking now holds, `span` is excluded, and equality is not degenerate (differing dates, data and accounts all still compare unequal). `build`, `Python-gated cargo tests` and `Validate branch name` were all green; `Cargo Deny` and `quality` are pre-existing failures on main, unrelated to this.

Two things you added that improved on the review rather than just satisfying it: the destructuring, which turns "a new field was silently left out of equality" into a compile error — that being the exact shape of the `directive_type` bug — and moving the hashing note onto the structs, where a maintainer adding `#[derive(Hash)]` will actually be looking.

Issue #2171, opened 2026-08-27
BQL: literal and binary-expression targets head as colN; bean-query prints the expression


robcohen:
Found reviewing #2170, which fixes the other branches of the same function but
cannot fix this one cheaply.

`Executor::expr_to_name` heads a literal or a binary/unary expression as
`col0`, `col1`, ... bean-query names it by printing the expression:

```
$ bean-query -f csv f.bean "SELECT 1 FROM #notes"                    -> 1
$ bean-query -f csv f.bean "SELECT 'lit' FROM #notes"                -> 'lit'
$ bean-query -f csv f.bean "SELECT number + 1 FROM #postings"        -> number + 1
$ bean-query -f csv f.bean "SELECT NOT (number > 0) FROM #postings"  -> NOT (number > 0)

$ rledger query ... (all of the above)                               -> col0
```

`colN` tells a reader nothing about which target it names, and a script
reading our CSV by header cannot find the column at all.

## Why #2170 did not just fix it

The obvious fix is `expr.to_string()`, matching what that PR does for function
targets. It produces the wrong string:

| | bean-query | `expr.to_string()` |
|---|---|---|
| `SELECT 'lit'` | `'lit'` | `"lit"` |
| `SELECT number + 1` | `number + 1` | `(number + 1)` |
| `SELECT NOT (number > 0)` | `NOT (number > 0)` | `NOT ((number > 0))` |

bean-query prints with **minimal parentheses**, measured:

```
SELECT (number + 1) * 2   ->  (number + 1) * 2    parens kept where precedence needs them
SELECT number + 1 * 2     ->  number + 1 * 2      none added
SELECT ((number))         ->  number              redundant parens dropped
```

Our `Display for Expr` parenthesizes every binary node unconditionally, and
uses double quotes for string literals.

## What a fix needs

A precedence-aware printer. `Display` cannot simply be changed to omit parens:
it is also what names the synthetic hidden ORDER BY columns
(`find_hidden_order_by_targets` uses the expression's rendered text as the
column's alias) and it appears in error messages, so a rendering that loses
information would break the lookup #2170 just repaired. Most likely a separate
`header_name()` rendering that tracks precedence, leaving `Display` alone.

Shipping `expr.to_string()` as a partial fix was considered and rejected: it
still would not match bean-query, so it buys no compatibility, while churning
every affected header — including the rustfava snapshots — a second time.

## Related

[...]

PR #2241, opened 2026-09-03
fix(query)!: -f json emits positional rows, not keyed objects

### LINKED ISSUE #2178: query -f json silently drops a column when two columns share a name
Found reviewing #2175. Pre-existing — a `main`-built binary behaves
identically — and orthogonal to that PR, but it is silent data loss.

`-f json` renders each row as an object keyed by column name. Two columns with
the same name collapse:

```
$ rledger query -f json f.bean "SELECT date AS A, account AS a FROM #notes"
{
  "columns": ["a", "a"],          <-- two columns
  "row_count": 1,
  "rows": [
    { "a": "Assets:Cash" }        <-- one key; the date is gone
  ]
}
```

The `columns` array is right, so a consumer that trusts it expects two fields
and finds one. The surviving value is the LAST column's; the first is dropped
with no error.

`-f csv` gets this right — it emits both fields positionally.

## How duplicate names arise

- identical aliases: `SELECT date AS a, account AS a` (works on `main` too)
- aliases differing only in case, since #2164 lowercases them:
  `AS A` and `AS a` now collide
- the same expression twice: `SELECT number + 1, number + 1` (#2171)
- a column colliding with an alias: `SELECT account, date AS ACCOUNT`

bean-query produces duplicate CSV headers for these too, so the names
themselves are correct — matching bean-query was verified in #2164. It is only
the JSON shape that cannot represent them.

## Fix is a design decision

bean-query has no JSON output (`beancount|csv|text` only), so there is nothing
to match. Options:

- emit `rows` as arrays of values, positionally, with `columns` as the key —
  faithful, but a breaking change for anything reading `rows[i].name`
- disambiguate duplicates in the object (`a`, `a_2`) — keeps the shape, invents
  names a consumer cannot predict
- reject the query — surprising, since CSV handles it

Worth deciding rather than patching. Note the wasm/FFI query surfaces return
columns and rows separately, so they may already be positional; only the CLI
`-f json` builds objects.

robcohen:
A JSON object cannot hold duplicate keys, and BQL legitimately produces duplicate column names. Keyed rows silently dropped all but the last of each collision **while `columns` still listed them all**, so a consumer that trusted `columns` read fewer fields than it was told to expect. Closes #2178.

```
SELECT date AS a, account AS a

before  {"columns": ["a","a"], "rows": [{"a": "Assets:Cash"}]}          <- date gone
after   {"columns": ["a","a"], "rows": [["2024-01-02","Assets:Cash"]]}
```

## Why positional, not disambiguation

The names are not the problem. bean-query produces the same duplicate headers (verified in #2164), and `-f csv` already emitted both fields positionally. Only the JSON shape could not represent its own results.

Positional also makes the CLI agree with the surfaces that were **always** positional — the component's `rows: list<list<query-value>>` and wasm's `rows: Vec<Vec<CellValue>>`. The CLI's JSON was the odd one out, which is what made it the wrong shape rather than a matter of taste.

Disambiguating (`a`, `a_2`) was considered and rejected: it invents names a consumer cannot predict, and it would make the JSON keys disagree with the CSV/text headers, which are pinned to bean-query's.

## Breaking

Anything reading `rows[i].name` must zip `columns` with the row.

`docs/guides/integration.md` carried exactly that pattern — a Python example doing `row['account']` — and is updated with the zip, the shape documented explicitly, and the reason. The `jq` recipe in `docs/commands/query.md` used `select(.balance > 100)` and is updated too.

Both were run rather than written from memory, which is how I caught that jq's `from_entries` does **not** accept bare `[key, value]` pairs — the obvious idiom errors with `Cannot index array with string ("key")`. The documented recipe is the one that actually works:

```bash
rledger query ledger.beancount "BALANCES" -f json \
  | jq -c '.columns as $c | .rows[] | [$c, .] | transpose
           | map({key: .[0], value: .[1]}) | from_entries'
```

## Verification

Two tests: one that duplicate names keep every value, and one that row width equals column count for both unique and duplicate names — the invariant the keyed shape broke precisely when it mattered. Restoring keyed rows fails 3 tests.

Checked all four collision shapes the issue lists reach the fix: identical aliases, aliases differing only in case (#2164 lowercases them), the same expression twice (#2171), and a column colliding with an alias.

4724 workspace tests green, clippy clean on 1.97.0, typos clean.

### DIFF
--- a/crates/rustledger/src/cmd/query/output.rs
+++ b/crates/rustledger/src/cmd/query/output.rs
@@ -223,19 +223,31 @@ fn write_csv<W: Write>(
     Ok(())
 }
 
+/// Rows are ARRAYS of values, positional against `columns` -- not objects
+/// keyed by column name.
+///
+/// A JSON object cannot hold duplicate keys, and BQL legitimately produces
+/// duplicate column names: `SELECT date AS a, account AS a`, two aliases
+/// differing only in case (since #2164 lowercases them), the same expression
+/// twice (#2171), or a column colliding with an alias. Keyed rows silently
+/// dropped all but the last of each collision while `columns` still listed
+/// them all, so a consumer that trusted `columns` read fewer fields than it
+/// was told to expect (#2178).
+///
+/// The names themselves are correct -- bean-query produces the same duplicate
+/// headers, verified in #2164 -- so the fix belongs in the shape, not the
+/// naming. Disambiguating (`a`, `a_2`) would invent names a consumer cannot
+/// predict AND make these keys disagree with the CSV/text headers.
+///
+/// This matches the other query surfaces, which were always positional: the
+/// component's `rows: list<list<query-value>>` and wasm's
+/// `rows: Vec<Vec<CellValue>>`. `-f csv` emits both fields positionally too.
+/// The CLI's JSON was the only shape that could not represent its own results.
 fn write_json<W: Write>(result: &rustledger_query::QueryResult, writer: &mut W) -> Result<()> {
     let rows: Vec<serde_json::Value> = result
         .rows
         .iter()
-        .map(|row| {
-            let obj: serde_json::Map<String, serde_json::Value> = result
-                .columns
-                .iter()
-                .zip(row.iter())
-                .map(|(col, val)| (col.clone(), value_to_json(val)))
-                .collect();
-            serde_json::Value::Object(obj)
-        })
+        .map(|row| serde_json::Value::Array(row.iter().map(value_to_json).collect()))
         .collect();
 
     let output = serde_json::json!({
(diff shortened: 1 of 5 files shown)


robcohen:
Both comments were right, and the review also pointed at something CI had just caught — fixed in 11f6455.

**The doc examples contradicted themselves.** Both converted rows back into objects (`dict(zip(...))` in Python, `from_entries` in jq), which collapses duplicate column names exactly as the old shape did — and the prose under the Python one claimed the opposite. They are now positional, with the object form offered as the convenience it is: correct when the names are unique, which is the reader's to know.

**The real find was `Run compatibility tests` failing.** "Booking fuzz found an unexplained divergence" and "Booked-value comparison did not come back clean" were not divergences at all — `scripts/compat-booking-fuzz.py` and `scripts/compat-values.py` both read `query --format json` and index each row by column name, so the positional shape broke them. Three call sites, now going through a `_keyed_rows` helper that zips `columns` with each row.

That's the same class of miss as the docs: I audited `crates/` and `docs/` for consumers of this output and did not audit `scripts/`.

Verified by running both rather than trusting the edit — booking fuzz agrees 15/15, the value harness reports 0 posting and 0 price divergences over 8 corpus files. And I made it report **dirty** first: with `_keyed_rows` stubbed to return nothing it reports 7 posting divergences, so the zero is a comparison rather than silence.

PR #2242, opened 2026-09-03
fix(ffi): a balance diff belongs to its own assertion

### LINKED ISSUE #2239: FFI: attach_balance_diffs collapses same-date assertions on one account, so a Balance can carry another's diff
Found reviewing #2238, which fixed the same collision in the query layer. This one is separate code with a different contract, so it is filed rather than folded in.

`attach_balance_diffs` (`crates/rustledger-ffi-component/src/convert.rs`) keys the validator's results by `(date, account, currency)`:

```rust
let map: HashMap<(String, String, String), Decimal> = actuals
    .iter()
    .map(|a| ((a.date.to_string(), a.account.to_string(), a.currency.to_string()), a.diff))
    .collect();
```

Those three do not identify a balance assertion. A ledger may assert the same account and currency twice on one date with **different amounts**:

```beancount
2024-03-01 balance Assets:Cash   90.00 USD
2024-03-01 balance Assets:Cash  120.00 USD
```

`collect()` keeps one entry per key, so both directives are then attached the same `diff` — one of them another assertion's. In the query layer the equivalent produced `-20.00 USD` on both rows where bean-query returns `10.00 USD` and `-20.00 USD`.

## The fix is available but not identical

`BalanceActual` now carries `asserted`, added in #2238 precisely so an assertion can be identified. Adding it to the key here is mechanical.

What is **not** transferable is the filtering. `#balances.discrepancy` reports only FAILING assertions, matching beancount's `diff_amount`. This surface deliberately attaches a diff for every assertion including passing ones, where it is `Some(0)` — a different contract, and presumably load-bearing for whatever renders per-assertion pass/fail (#1663). So the fix is the key, not the semantics.

## Not verified end-to-end

I read the code and reproduced the collision in the query layer, but did not build a WIT-level test for this surface. Worth doing as part of the fix rather than trusting the analogy.

robcohen:
`attach_balance_diffs` keyed the validator's results by `(date, account, currency)`. Those three do not identify a balance assertion — a ledger may assert the same account and currency twice on one date with different amounts, and so different differences. Building a map collapsed them, and one directive received the other's diff. Closes #2239.

```
2024-03-01 balance Assets:Cash    90.00 USD
2024-03-01 balance Assets:Cash   120.00 USD

before  both carry -20.00
after   90.00 carries 10.00, 120.00 carries -20.00
```

bean-query reports `10.00` and `-20.00` here. The query layer had the same collision, fixed in #2238 — which is also what added `BalanceActual::asserted`, so the fix here is the **key**, not the semantics.

This surface deliberately attaches a diff for *every* assertion including passing ones, where it is zero. That is its contract (#1663) and is unchanged, which is why this wasn't folded into #2238.

The asserted number is compared as the string the WIT directive carries. Both sides come from `Decimal::to_string()` on the same value — `amount_from_core` for the directive, the validator's record for the map — so they agree by construction rather than by luck. I checked that rather than assuming, since a string key that silently never matched would drop the diff entirely, which is worse than attaching the wrong one.

## Tested at the WIT level

The collision is in the map the component builds, so a validator unit test cannot see it. When I filed #2239 I noted I'd read the code but not built a component-level test; this is that test.

**The component wasm was rebuilt for both runs.** The checked-in one was three days stale, and testing against it would have measured old code whichever way it came out. Against a freshly built *unfixed* component the test fails with both diffs at `-20.00`; against a freshly built *fixed* one it passes.

Full `parity` suite: 27 passed, including the existing `load_reports_balance_assertion_failure` that pins the #1663 contract.

No WIT change — the wire contract is untouched, only which record attaches to which directive.

4725 workspace tests green, clippy clean on 1.97.0, typos clean.

### DIFF
--- a/crates/rustledger-ffi-component-tests/tests/parity.rs
+++ b/crates/rustledger-ffi-component-tests/tests/parity.rs
@@ -1041,6 +1041,69 @@ fn load_reports_balance_assertion_failure() -> Result<()> {
     Ok(())
 }
 
+/// #2239: two `balance` assertions sharing a date, account and currency each
+/// carry their OWN diff.
+///
+/// `attach_balance_diffs` keyed the validator's results by
+/// `(date, account, currency)`. Those three do not identify an assertion --
+/// a ledger may assert the same account and currency twice on one date with
+/// different amounts -- so building a map collapsed them and one directive
+/// received the other's difference.
+///
+/// A WIT-level test on purpose: the collision is in the map the component
+/// builds, so a unit test on the validator would not see it. Verified against
+/// bean-query, which reports `10.00 USD` and `-20.00 USD` here (the query
+/// layer had the same bug, fixed in #2238).
+#[test]
+fn same_date_balance_assertions_carry_their_own_diffs() -> Result<()> {
+    if !component_path().exists() {
+        eprintln!("skip: component wasm not built");
+        return Ok(());
+    }
+    let (mut store, inst) = instantiate()?;
+    // Accumulates 100.00. Asserting 90 is +10 too much; asserting 120 is -20.
+    let src = "\
+2024-01-01 open Assets:Cash USD
+2024-01-01 open Equity:O USD
+2024-01-02 * \"fund\"
+  Assets:Cash   100.00 USD
+  Equity:O
+2024-03-01 balance Assets:Cash    90.00 USD
+2024-03-01 balance Assets:Cash   120.00 USD
+";
+    let loaded = inst
+        .rustledger_ledger_ledger()
+        .call_load(&mut store, src, "<stdin>", false)?;
+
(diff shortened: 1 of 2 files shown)


robcohen:
Right, and the comment was wrong about the thing it existed to explain — corrected in the latest commit.

`BalanceActual::diff` is the raw signed `computed - asserted`, recorded before the tolerance is consulted. Measured through the component rather than reasoned about:

```
asserted=100.01  diff=-0.01   <- passes within a 0.05 tolerance, diff is NOT zero
asserted=100.00  diff= 0.00   <- zero only on an exact match
```

The comment now says the value is the raw signed difference, zero only on an exact match — and notes that this is precisely why `#balances.discrepancy`, which reports a difference only for failures, could not reuse it.

PR #2244, opened 2026-09-03
fix(lsp): documentColor covers amounts with thousands separators

### LINKED ISSUE #2230: documentColor skips amounts with thousands separators, breaking the alignment format just produced
`textDocument/documentColor` skips any amount containing thousands separators, so in a ledger that declares `option "render_commas" "TRUE"` — the very ledger whose canonical form `rledger format` now *produces* — only the small amounts get a colour. VS Code renders each colour as an inline swatch occupying one character cell, so the coloured lines shift right by a column and the amount column goes ragged.

Found after adopting #1896: before it, no amount in our files had separators, every amount got a swatch, and the shift was uniform and invisible. Writing separators into the files made it uneven.

### Reproduction

`rledger-lsp 0.22.0`, `rustledger-vscode 0.22.0`, Windows.

```beancount
option "title" "Probe"
option "operating_currency" "USD"
option "render_commas" "TRUE"

2026-01-01 open Assets:Cash USD
2026-01-01 open Expenses:Fees USD

2026-01-05 * "P" "grouped and plain on purpose"
  Assets:Cash    -1,999.00 USD
  Expenses:Fees   1,999.00 USD

2026-01-06 * "P" "plain only"
  Assets:Cash       -14.62 USD
  Expenses:Fees      14.62 USD
```

Driving the server over stdio — `initialize`, `initialized`, `didOpen`, then `textDocument/documentColor`:

```
documentColor returned 2 colour(s)

  line 13  '-14.62'   in: Assets:Cash       -14.62 USD
  line 14  '14.62'    in: Expenses:Fees      14.62 USD

  no colour  Assets:Cash    -1,999.00 USD
  no colour  Expenses:Fees   1,999.00 USD
  COLOURED   Assets:Cash       -14.62 USD
  COLOURED   Expenses:Fees      14.62 USD
```

**Expected:** four colours, one per posting amount.
**Actual:** two — the grouped amounts are skipped.

### Why it looks like a parser gap rather than a policy

`rledger format --ledger <root>` on that same file emits the grouped spelling, and `format --check --ledger <root>` accepts it as canonical. So the colour provider is declining to colour text the formatter itself writes. The parser behind the colour range presumably matches something like `-?\d+(\.\d+)?` and stops at the first comma.

### Suggested fix

Strip separators before parsing the number in the colour provider, the way the formatter's own path does. If the colour range is meant to cover the whole numeric token, the range wants widening to include the separators too, so the swatch lands in the same place for both spellings.

### Aside on the swatch itself

Worth considering whether a `documentColor` is the right vehicle for a sign indicator at all. VS Code treats the response as an editable colour literal: it offers a colour picker on the swatch, and the swatch consumes a character cell, which perturbs column alignment that `format` has just carefully established. An inlay hint or a semantic token modifier would convey sign without moving text. Not a blocker — just that the two features are working against each other right now.

For the moment we have set `"editor.colorDecorators": false` for `beancount`, which restores the alignment and loses the sign colour.

robcohen:
The color provider matched the line against `Decimal::to_string()`, which never groups, so an amount written `-1,999.00` matched nothing and got no color. Closes #2230.

The reporter's diagnosis was right — it was a matching gap, not a policy.

## Why it matters more than a missing color

`option "render_commas" "TRUE"` makes the grouped spelling canonical, and `rledger format` writes it. So the provider was declining to color exactly the text the formatter produces. On an aligned block that is worse than no color at all: an editor draws each swatch in a character cell, so coloring the small amounts and not the grouped ones makes the column ragged — undoing the alignment `format` had just established. That is why the reporter saw it only after adopting #1896.

## The fix

A fallback that compares whitespace-delimited tokens with separators removed, returning the range over the **whole token as written** so the swatch lands in the same place for both spellings. The issue asked for both halves and both matter — a range that stopped at the comma would put the swatch mid-number.

Tokenizing on whitespace preserves the existing account-name guard: `Assets:US-100:Bank` is one token and does not compare equal to a bare number.

Balance and price amounts share `find_amount_range` and are fixed with it. They are asserted separately, since they are different arms of the directive match and a posting-only fix would have left them behind.

## Verification

Driven through the **real server over stdio**, on the reporter's exact ledger — the same way they reproduced it:

```
UNFIXED   documentColor returned 2 colour(s)      FIXED   returned 4
  line 12  '-14.62'                                 line 8   '-1,999.00'
  line 13  '14.62'                                  line 9   '1,999.00'
                                                    line 12  '-14.62'
                                                    line 13  '14.62'
```

Unit tests pin both halves independently: removing the fallback fails one, and letting the range stop at the comma fails it too. A test also checks the sign still drives the color, so a provider that colored everything one color would not pass.

4727 workspace tests green, clippy clean on 1.97.0, typos clean.

## Not addressed

The reporter's aside — whether `documentColor` is the right vehicle for a sign indicator at all, given the swatch occupies a character cell and VS Code offers a color picker on something that is not a color literal. An inlay hint or semantic token modifier would convey sign without moving text. That is a design question worth its own issue rather than a change smuggled into a bug fix; say the word and I'll open one.

Their workaround (`"editor.colorDecorators": false`) can be dropped once this ships, at the cost of the alignment question above still being open.

### DIFF
--- a/crates/rustledger-lsp/src/handlers/document_color.rs
+++ b/crates/rustledger-lsp/src/handlers/document_color.rs
@@ -247,15 +247,17 @@ fn find_amount_range(
 /// Whitespace-delimited tokens of `line`, each with its byte offset.
 ///
 /// `str::split_whitespace` discards positions, and the color range needs
-/// them. ASCII-only classification, matching the boundary check above.
+/// them. ASCII-only classification, matching the boundary check above --
+/// BOTH ends, so a non-ASCII space cannot start a token by one rule and
+/// terminate it by another.
 fn whitespace_tokens(line: &str) -> impl Iterator<Item = (usize, &str)> {
     line.char_indices()
-        .filter(|(i, c)| {
-            !c.is_ascii_whitespace() && (*i == 0 || line.as_bytes()[i - 1].is_ascii_whitespace())
+        .filter(|&(i, c)| {
+            !c.is_ascii_whitespace() && (i == 0 || line.as_bytes()[i - 1].is_ascii_whitespace())
         })
         .map(move |(start, _)| {
             let end = line[start..]
-                .find(char::is_whitespace)
+                .find(|c: char| c.is_ascii_whitespace())
                 .map_or(line.len(), |off| start + off);
             (start, &line[start..end])
         })
--- a/crates/rustledger-lsp/src/handlers/document_color.rs
+++ b/crates/rustledger-lsp/src/handlers/document_color.rs
@@ -207,7 +207,58 @@ fn find_amount_range(
         }
     }
 
+    // Nothing matched literally. The line may spell the number with THOUSANDS
+    // SEPARATORS -- `-1,999.00` where `amount_str` is always `-1999.00`,
+    // because it comes from `Decimal::to_string()` and that never groups.
+    //
+    // This is not a rare spelling: `option "render_commas" "TRUE"` makes it
+    // the canonical form, and `rledger format` writes it. So the provider was
+    // declining to color exactly the text the formatter produces, coloring
+    // only the amounts too small to carry a separator -- which in an aligned
+    // block makes the column ragged, since an editor draws each color as a
+    // swatch occupying a character cell (#2230).
+    //
+    // Compare whitespace-delimited tokens with separators removed, and return
+    // the range over the WHOLE token as written so the swatch lands in the
+    // same place for both spellings. Tokenising on whitespace keeps the
+    // account-name guard above: `Assets:US-100:Bank` is one token and does not
+    // compare equal to a bare number.
+    for pattern in &search_patterns {
+        for (pos, token) in whitespace_tokens(line) {
+            if !token.contains(',') {
+                continue;
+            }
+            let ungrouped: String = token.chars().filter(|c| *c != ',').collect();
+            if ungrouped == *pattern {
+                let after_pos = pos + token.len();
+                let (sl, sc) = line_index.offset_to_position(line_start_byte + pos);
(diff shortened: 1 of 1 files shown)


robcohen:
Both fixed in the latest commit.

The whitespace one was a real inconsistency, not just style: the doc claimed ASCII-only classification "matching the boundary check above", but the token **end** used `char::is_whitespace` — Unicode. So a non-ASCII space could fail to start a token by one rule while terminating it by the other. No test reached it and beancount amounts are ASCII-delimited in practice, but the comment described code that did not exist. Both ends are ASCII now.

The `&usize` indexing is destructured in the pattern instead, so it reads as the plain index it is.

Re-verified end to end after the change: the probe against the real server still returns 4 colors on the reporter's ledger, and clippy is clean on 1.97.0.

robcohen:
Opened #2245 for the design question you raised — whether `documentColor` is the right vehicle for a sign indicator at all.

Two things I checked while writing it up that sharpen the framing:

- **The edit hazard is already defused.** `handle_color_presentation` returns `text_edit: None`, so choosing a color in the picker cannot rewrite the ledger. What remains is that VS Code still *offers* a color chooser on non-color data, and its only outcome is a label reading "Negative amount" — a dead-end interaction rather than a dangerous one.
- **The alternative is cheaper than it looks.** `semantic_tokens_provider` is already advertised, `TOKEN_TYPES` already classifies amounts as `NUMBER`, and `TOKEN_MODIFIERS` already exists with three entries. Adding a `negative` modifier would let a theme convey sign with no layout impact at all. Inlay hints are also wired, but a hint occupies space too, so it trades one alignment problem for another.

This PR stays as-is: it makes the current behavior *consistent*, which was the actual bug. #2245 is about whether to keep the vehicle.

Issue #2245, opened 2026-09-03
documentColor conveys amount sign at the cost of alignment; a semantic-token modifier may be the better vehicle


robcohen:
Found while reviewing #2161. Both predate that PR — they live in the shared
`Metadata` → `Value` conversion, not in the per-table columns #2161 adds.

## 1. `#entries.meta` drops `filename` and `lineno`

beancount stores `filename` and `lineno` in every directive's metadata, and
`#entries.meta` surfaces them. We return only user-declared keys.

```
2020-01-04 note Assets:Cash "a note"
  nval: "nm"
```

```
$ bean-query -f csv f.bean "SELECT meta FROM #entries WHERE type='note'"
{'filename': '/tmp/f.bean', 'lineno': 10, 'nval': 'nm'}

$ rledger query -f csv f.bean "SELECT meta FROM #entries WHERE type='note'"
{nval: nm}
```

Note this is per-table in bean-query, and #2161 confirmed we match the other
side of it: `#notes.meta` returns `{'nval': 'nm'}` in bean-query too, without
`filename`/`lineno`. So the fix belongs on `#entries` specifically, not in the
shared conversion — changing it globally would break the five tables #2161
just brought into agreement.

## 2. Empty metadata is NULL, not an empty map

```
2020-01-03 note Assets:Cash "no meta"
```

```
$ bean-query -f csv f.bean "SELECT meta FROM #notes"
{}

$ rledger query -f csv f.bean "SELECT meta FROM #notes"
(empty)

$ rledger query -f csv f.bean "SELECT meta IS NULL FROM #notes"
true
```

bean-query renders NULL as `""` in CSV and an empty map as `{}`, so these are
distinct values there, not a formatting artifact — `meta IS NULL` is `false`
in bean-query and `true` for us.

This one IS in the shared conversion and affects every table carrying a
`meta` column, including the five added in #2161.

## Why this wasn't caught

The compat suites compare directives that *have* metadata. Both divergences
only appear on the absent case, which no fixture exercised. Worth a corpus
file with a bare directive of each kind.

## Not included

`SELECT *, meta` is a syntax error in bean-query and accepted by us. That is
parser permissiveness unrelated to metadata, and arguably ours is the better
behavior; noting it here only so it is not rediscovered as part of this.

robcohen:
[...]

Issue #2219, opened 2026-09-01
BQL: PIVOT BY pipeline order differs from bean-query for LIMIT and ORDER BY


robcohen:
With `PIVOT BY` now applying on `FROM` queries (#2216), two pipeline-ordering differences become reachable from the CLI. Both are pre-existing — `main`'s no-`FROM` path behaves the same way — and both are in where `apply_pivot` sits, so they affect every path.

Fixture: three accounts, postings in 2024 and 2025.

## 1. LIMIT applies before the pivot in bean-query, after it here

```
SELECT account, year, sum(number) FROM #postings
GROUP BY account, year PIVOT BY account, year LIMIT 2

bean-query:  account/year,2024              <- 2025 column absent
             Assets:A, 10.00
             Equity:O,-30.00

rustledger:  account,2024,2025
             Assets:A,10.00,5.00
             Equity:O,-30.00,-5.00
```

bean-query limits the **pre-pivot rows** to 2, so only the years those two rows carry become columns. We pivot first and limit the **pivoted rows**, so every year survives as a column.

The two readings are genuinely different features: `LIMIT` on grouped rows against `LIMIT` on pivot output.

## 2. bean-query discards ORDER BY when pivoting

```
SELECT ... GROUP BY account, year ORDER BY account DESC PIVOT BY account, year

bean-query:  Assets:A, Assets:B, Equity:O   <- ascending, DESC ignored
rustledger:  Equity:O, Assets:B, Assets:A   <- descending, as asked
```

Control: **without** `PIVOT BY`, bean-query honors the same `ORDER BY account DESC` (Equity:O first). So the clause is dropped specifically when pivoting.

Here **ours is arguably the better behavior** — it does what the query said. "Match bean-query" would mean adopting a rule that silently ignores an explicit clause, which needs a decision rather than a patch.

## Scope

Not fixed in #2218, which is about the pivot being skipped entirely. These are ordering choices baked into `execute_select`'s pipeline since before that PR; changing them moves the currently-working no-`FROM` path too, and #2 is a question about which behavior is right rather than a bug to close.

Related: #2217 (pivot column headers).

Issue #2214, opened 2026-09-01
BQL: SUM over a boolean errors here; bean-query computes 2 but prints TRUE


robcohen:
```
$ rledger query f.bean "SELECT sum(number > 0) FROM #postings"
error: type error: SUM requires numeric or position values

$ bean-query -f csv f.bean "SELECT sum(number > 0) FROM #postings"
TRUE
```

Python sums booleans as integers, so bean-query does compute something sensible — but what it prints is not it.

## bean-query's value and its output disagree

Through the API, the value is an integer:

```python
conn.execute("SELECT sum(number > 0) FROM #postings").fetchall()
# [(2,)]        two of the four postings are positive
```

The CLI prints `TRUE`. The result column is typed `bool` because the argument is, so the renderer formats the integer `2` through the boolean formatter. **The number is right and the output is wrong**, which makes "match bean-query" ambiguous for this one — matching the value means printing `2` and diverging from its output; matching the output means reproducing a display bug.

## The neighbours, for contrast

Measured against beanquery 0.2.0 / beancount 3.2.3:

| aggregate | bean-query | rustledger |
|---|---|---|
| `count(number > 0)` | 4 | 4 |
| `first(number > 0)` | TRUE | TRUE |
| `last(number > 0)` | TRUE | TRUE |
| `max` / `min` | TRUE / FALSE | TRUE / FALSE after #2212 |
| **`avg(number > 0)`** | `CompilationError: no function matches "avg(bool)"` | type error — **agrees** |
| **`sum(number > 0)`** | `2`, printed as `TRUE` | type error |

So `sum` is the only one left, and `avg` is already correct by agreeing to refuse.

## Suggested resolution

If this is worth doing, compute the integer and print it as an integer — `2`, not `TRUE`. That matches bean-query's *value*, is what a user asking "how many postings are positive" means, and does not reproduce a rendering bug. Worth a divergence note in `docs/reference/compatibility.md` either way, since the outputs will differ.

Low priority: `count(number > 0)` already answers the same question correctly in both tools.

## Found

Reviewing #2212, checking whether the sibling aggregates needed the same treatment as MIN/MAX. `avg` did not — it agrees by refusing.

PR #2246, opened 2026-09-03
feat(lsp): convey amount sign as a semantic-token modifier too

robcohen:
A negative amount now carries a `negative` modifier on its `NUMBER` token, and the token covers the **sign**. Themes color it with no layout impact. Part of #2245.

## documentColor stays

You asked to support as many IDEs as possible, and that settles the trade-off I'd flagged: `documentColor` reaches clients that don't implement semantic tokens, so dropping it would trade one gap for another.

Keeping both is what the two features are actually for — the swatch is a *decoration*, the modifier is *text color* — so they don't conflict. And it turns the existing escape hatch from lossy into free: a user who sets `editor.colorDecorators: false` to protect the alignment `rledger format` establishes currently loses the sign signal entirely. After this they keep it.

## Why the client half matters as much as the server half

This is the part I'd underestimated in #2245, and it's why "add a modifier" isn't the whole job: **a custom modifier is ignored by any theme that doesn't know its name.**

`packages/vscode/package.json` contributed no semantic-token metadata at all — only `commands`, `configuration`, `languages`. So it now declares the modifier and maps `number.negative` through `semanticTokenScopes` onto scopes existing themes already style, with `constant.numeric` as the fallback. Without that, the server change would be invisible in the editor we ship.

`negative` is custom because LSP defines no standard modifier for sign. Clients that don't recognise it ignore it rather than erroring.

## Detecting the sign

The lexer emits `-1,999.00` as `Minus` + `Number` (deliberately, so subtraction parses), so sign isn't on the number token. A minus is unary when the preceding token doesn't **end an operand** — after a number or closing paren it's subtraction.

The emitted token starts at the minus when the two are adjacent, so the span matches what `documentColor` reports. Otherwise one feature would cover `1,999.00` and the other `-1,999.00`, and a theme and a swatch would disagree about where the amount begins.

## Verification

Through the real server over stdio:

```
line 6 '-1,999.00'  modifiers=1000  NEGATIVE
line 7 '1,999.00'   modifiers=0000
```

Getting that probe right took two tries worth recording: the request is dispatched **asynchronously**, so piping `shutdown`/`exit` in the same batch made the server exit before the worker replied — the missing response was my harness, not the server.

Unit tests pin the parts a naive implementation gets wrong: `5 - 3` marks nothing (binary minus), and a detached `- 5` doesn't swallow the space into the token.

4728 workspace tests green, clippy clean on 1.97.0, typos clean, `package.json` valid.

## Note

Independent of #2244 (the #2230 fix) — different files, either order merges fine. #2245 stays open for the remaining question of whether the swatch should eventually go; this PR deliberately doesn't answer it.

### DIFF
--- a/crates/rustledger-lsp/src/handlers/semantic_tokens.rs
+++ b/crates/rustledger-lsp/src/handlers/semantic_tokens.rs
@@ -95,6 +95,11 @@ mod token_type {
 mod token_modifier {
     pub const DEFINITION: u32 = 1 << 0;
     pub const DEPRECATED: u32 = 1 << 1;
+    /// Index 2 in `TOKEN_MODIFIERS`. Declared but not yet applied anywhere --
+    /// present so the bit indices here line up with the legend and the next
+    /// modifier added cannot be off by one.
+    #[expect(dead_code, reason = "keeps bit indices aligned with TOKEN_MODIFIERS")]
+    pub const READONLY: u32 = 1 << 2;
     pub const NEGATIVE: u32 = 1 << 3;
 }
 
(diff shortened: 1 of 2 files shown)


robcohen:
The first one was a real bug, not a polish note — fixed in cdf64dd.

`semanticTokens/range` and `semanticTokens/full` held **separate copies** of the token arithmetic, and I wired the negated-number span into only one. A client asking for a range — which editors do on large files — saw `1,999.00` with no modifier where a full request gave `-1,999.00` with one. Two answers for the same text depending on which request was made, which is worse than the feature simply not existing.

Both now go through one `raw_token`, so they cannot drift again. That is the same duplicate-path shape as the two SUM accumulators in #2214 and the three pipeline sites in #2219, and I walked into it the same way: I found one call site, wired it, and did not ask whether there was a second.

Pinned by comparing the two collectors **against each other** over a whole-document range rather than against a literal, so they stay equal even if the expected span later changes. The test also asserts the fixture really contains a negative — otherwise two collectors agreeing on nothing would pass. Making the range collector ignore negatives again fails it.

`token_modifier::READONLY` added too. You are right that its absence made `NEGATIVE` at `1 << 3` look like it might be off by one; it is declared unused with the reason, so the next modifier added lines up with the legend.

PR #2247, opened 2026-09-03
fix(ci): restore the nightly compat budget and stop a false stale report

robcohen:
Two unrelated problems behind the same health issue (#2232).

## 1. compat.yml was timing out — my regression

#2223 added a second CSV invocation per run to compare column names. I measured ~1.7× on a 71-file sample, judged it acceptable, and **never checked it against the job's cap**. On the full corpus:

| | BQL step | job |
|---|---|---|
| Sep 1 (before #2223) | 18m37s | 29 min, success |
| Sep 2 / Sep 3 (after) | 34m36s | hit `timeout-minutes: 45`, cancelled |

Both nightlies were killed *after* BQL and never reached "Commit results to compatibility branch", so the trend lost two days — the exact failure the concurrency comment in that workflow warns about.

**Headers are now compared on a sample of files per query.** A header is the rendered target expression, so it depends on the query, not the ledger. Measured across the corpus: every query yields exactly **one** distinct header pair however many files it runs on — which is also why the divergence registry keys these `("*", query)` rather than enumerating files.

The cost is `queries × N` regardless of corpus size — 380 header passes against 735 files instead of ~14,000 — so the sample is generous rather than minimal. That property is what makes this a fix rather than a trade.

Two guards, because a cap that hides a divergence is worse than the runtime it saves:

- **Sampled files disagreeing** about a query's headers means the property the sample rests on is false → the run errors. That is what would happen if a `PIVOT BY` query were added, since a pivot names its columns after the data.
- **A query with no usable comparison** — every sampled file failing in one of the tools, which ~28% of runs do — errors rather than reporting a silent zero. My first attempt sampled 4 files and hit exactly this: 19 usable comparisons out of 76 budgeted.

Verified: verdicts unchanged (58/0/21 over 71 files), the same four queries still report header divergences, and **both guards fire when forced**. Runtime 1m44s → 1m10s against a 1m02s baseline. A pleasant accident: forcing zero coverage also trips the existing stale-mask check, so that registry independently notices a query losing header coverage.

## 2. nightly-health called a healthy workflow stale

Separately, the report claimed `bench.yml` had not run in 10 days, linking a real run from 2026-08-24. It had in fact run **six hours earlier**, successfully, as it had every day for weeks.

`latest_scheduled_run` took element zero of `gh run list`; it now takes the maximum by `createdAt`.

This does **not** claim to fix a diagnosed bug. The same command by hand returns the correct run, so I could not reproduce the cause and I am not going to invent one. What it removes is the dependence on list *order* — the only assumption that could produce that output. The self-test feeds a deliberately out-of-order page, since a correctly-ordered fixture cannot tell the two implementations apart; reverting to `runs[0]` fails it.

A false alarm matters more here than the diff size suggests: a health reporter exists to be believed, and one that cries wolf gets muted — which is the failure it was written to prevent.

## Note

The issue closes itself when everything is green, so it should clear on the next nightly if both fixes hold. Worth watching the next `compat.yml` scheduled run to confirm the job completes and publishes.

### DIFF
--- a/scripts/compat-bql-test.py
+++ b/scripts/compat-bql-test.py
@@ -1846,6 +1846,7 @@ def main() -> int:
     # DATA. Verified rather than assumed: if the sampled files disagree with
     # each other about a query's headers, the property the cap rests on is
     # false and the cap is hiding divergences on the files it skipped.
+    header_sampling_broken = False
     by_query: dict[str, set] = {}
     for r in results:
         if r.py_header is not None and r.rs_header is not None:
@@ -1860,6 +1861,7 @@ def main() -> int:
             f"{HEADER_SAMPLE_PER_QUERY} of them is not representative. Raise "
             "HEADER_SAMPLE_PER_QUERY or compare this query on every file."
         )
+        header_sampling_broken = True
 
     compared_headers = sum(1 for r in results if r.py_header is not None)
     print(
@@ -1877,6 +1879,7 @@ def main() -> int:
             f"::error::query {q!r} got NO header comparison -- every sampled "
             "file failed in one of the tools. Its column names are unchecked."
         )
+        header_sampling_broken = True
 
     skipped_headers = sum(1 for r in results if r.header_skipped)
     if skipped_headers:
@@ -2080,6 +2083,13 @@ def main() -> int:
             f"\nNo BQL regressions vs baseline "
             f"({len(baseline_passing)} baseline-passing pairs checked)."
         )
+
+    # An annotation alone does not fail a GitHub Actions step, so a guard that
+    # only printed would let a run whose header sampling is unsound report
+    # SUCCESS -- the check-that-cannot-fail this file exists to avoid.
+    # Deliberately last, so every other diagnostic is printed first.
+    if header_sampling_broken:
+        return 1
 
     return 0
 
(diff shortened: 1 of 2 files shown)


robcohen:
Ran the **full uncapped corpus** locally, because the PR's own compat run does not prove this: PR and push runs pass `--max-files 150` while only the scheduled run is uncapped, so the green check above is on a fifth of the files.

Full corpus, 735 files:

```
total runs:     13965
Header checks:  380 of 13965   (19 queries x 20 files)
::error lines:  0
header divergences still detected across:
  balances-by-account, journal-assets, journal-assets-at-cost, journal-assets-at-units
```

380 rather than ~10,000 is the whole point — the header cost is `queries x N` and does not grow with the corpus, so it stays ~26x smaller at full scale rather than only on the sample I developed against. The same four queries still report their divergences, and neither guard fired.

Sizing the nightly from the measured numbers: the header pass added ~16 minutes for ~10,000 successful runs, so ~0.1s each; 380 of them is roughly **36 seconds**. That puts the BQL step back near its 18m37s baseline and the job near 29 minutes against the 45-minute cap.

I am calling that a well-grounded prediction rather than a fact — the nightly runs on GitHub's runners, not mine. The real confirmation is the next scheduled `compat.yml` completing and reaching "Commit results to compatibility branch", which is the step the last two nightlies never got to.

robcohen:
You're right, and it's the worst version of the mistake: the comments I wrote beside those guards said "the run errors" and it didn't. A guard that only annotates is a check that cannot fail — which is the exact failure this harness exists to prevent, reintroduced by the change meant to protect it.

Fixed in a2c9fea. `main` now returns 1 when either guard trips, placed last so every other diagnostic still prints first.

Verified by **exit code**, not by reading the output — which is the only way this could have been caught in the first place:

```
healthy run                              exit 0
query with no usable header comparison   exit 1
headers varying across sampled files     exit 1
```

PR #2249, opened 2026-09-04
feat(lsp): amountColors turns the sign swatch off from any editor

robcohen:
A client can decline `documentColor` for amounts by passing `amountColors: false` in its initialization options; the server then doesn't advertise the capability at all. Part of #2245.

## Why this and not removal

The decision on #2245 was **keep both**, since `documentColor` reaches clients without semantic-token support. That left one real problem unsolved: the swatch occupies a character cell, so a colored line sits a column right of an uncolored one and the alignment `rledger format` produces comes apart (#2230) — and the only escape was `editor.colorDecorators`, which is **VS Code's**. An editor without an equivalent had no way to decline the decoration.

Given the goal of reaching as many editors as possible, the fix is a server-side switch rather than relying on each client having its own.

## Turning it off is lossless

That's the part worth checking rather than assuming, so I drove the real server over stdio:

```
default              colorProvider=True   negative-modifier tokens=1
amountColors=false   colorProvider=None   negative-modifier tokens=1
amountColors=true    colorProvider=True   negative-modifier tokens=1
```

With the swatch off, the capability is gone **and** the negated amount still reports its `negative` semantic-token modifier (#2246). So a user who wants their alignment back keeps the sign signal they were previously trading away — which is exactly what the reporter of #2230 gave up with `editor.colorDecorators: false`.

## Opt-out, not opt-in

Default on; an absent key leaves it on; the VS Code client sends the option **only** when the setting is false. Every existing client sees exactly what it sees today.

A non-boolean leaves the default alone rather than being read as "off" — silently disabling a feature because a client sent the wrong type is worse than ignoring the value. Both halves are pinned: loosening the parse to `unwrap_or(false)` fails one test, flipping the default fails both.

## Scope

Server plus the VS Code client (`rustledger.amountColors`). Other clients get it by passing the initialization option, which is the point — it needs no per-editor feature.

4736 workspace tests green, clippy clean on 1.97.0, typos clean, `tsc --noEmit` clean, `package.json` valid.

#2245 stays open for the remaining question — whether `documentColor` should eventually be retired — which now has considerably less urgency, since its cost is opt-out-able in every editor rather than only in VS Code.

### DIFF
--- a/crates/rustledger-lsp/src/ledger_state.rs
+++ b/crates/rustledger-lsp/src/ledger_state.rs
@@ -40,12 +40,36 @@ fn extract_price_currency(price: &PriceAnnotation) -> Option<String> {
 }
 
 /// Configuration for the LSP server, parsed from initialization options.
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct LspConfig {
     /// Path to the root journal file (e.g., "main.bean").
     /// When set, the LSP loads this file and all its includes for
     /// complete diagnostics and completions across the entire ledger.
     pub journal_file: Option<PathBuf>,
+    /// Whether to advertise `textDocument/documentColor` for amount signs.
+    ///
+    /// On by default. Turning it off suppresses the color swatch an editor
+    /// draws beside each amount, which occupies a character cell and so shifts
+    /// colored lines relative to uncolored ones -- perturbing the alignment
+    /// `rledger format` establishes (#2230, #2245).
+    ///
+    /// This is the SERVER-side switch, and it exists because the client-side
+    /// one is not portable: VS Code offers `editor.colorDecorators`, but an
+    /// editor without that setting has no way to decline the decoration. Sign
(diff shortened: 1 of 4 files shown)


robcohen:
Both fixed in 966a27a.

**The scope one was a real bug.** The extension reads this with `getConfiguration("rustledger", root)` — per workspace root — so without `"scope": "resource"` VS Code would not honor a folder-level override in a multi-root workspace. `rustledger.journalFile` declares it for exactly that reason, and it is the setting I based this one on, so I should have carried the scope across with it.

**The double parse** was the same shape as three other things this session: one config, two call sites, same answer today and free to diverge tomorrow. `Server::new` now takes the already-parsed `LspConfig` instead of re-deriving it, so there is one parse.

Re-verified end to end after the refactor rather than trusting that a move is a move — `amountColors: false` still drops the `colorProvider` capability while the negated amount keeps its semantic-token modifier.

PR #2279, opened 2026-09-08
feat(ofx): read FITID and LEDGERBAL

robcohen:
Two **Now**-tier roadmap items, done together because they are the same parsing pass. Both came out of [discussion #2126](https://github.com/rustledger/rustledger/discussions/2126).

```console
$ rledger extract bank.qfx -a Assets:Bank -c USD
2024-01-15 * "COFFEE" ^ofx-id-with-spaces-x
  Assets:Bank  -50.00 USD
  Expenses:Unknown

2024-02-01 balance Assets:Bank 1234.56 USD
```

## FITID → `^ofx-<id>` link

A **link**, not a tag: this is identity, not a category, and thousands of unique ids would swamp the tag namespace in autocomplete and tag queries. (@petemounce asked why beancount-import used a link; that is the reasoning.)

Sanitized to the lexer's link charset `[a-zA-Z0-9-_/.]`. A FITID is an opaque bank string, and one containing a space or a colon would otherwise emit a ledger **rustledger itself could not read**. An id that sanitizes to nothing yields no link, since `^ofx-` alone would be a link every such transaction shares.

## LEDGERBAL → balance assertion, dated `DTASOF + 1`

**This is the part worth reviewing.** A beancount `balance` asserts the balance at the *start* of its date, before that day's transactions. A bank's `DTASOF` figure is the close of business, *after* them. Emitting it verbatim asserts the previous day's balance, and would fail on any statement whose last day has activity.

I confirmed the semantics against the validator rather than assuming them:

```beancount
2024-01-15 * "spend"
  Assets:Bank  -10.00 USD
  Expenses:X

2024-01-15 balance Assets:Bank  0.00 USD    ← passes (pre-transaction balance)
2024-01-15 balance Assets:Bank -10.00 USD   ← E2001, expected -10.00, got 0
```

So `DTASOF 2024-01-31` emits `2024-02-01 balance`.

**A partial `LEDGERBAL` produces nothing.** A balance without a date cannot be placed, and a date without an amount asserts nothing; guessing at one is how a wrong assertion gets into someone's ledger. **The assertion is only emitted alongside transactions** — its job is to prove the transaction set is complete, and with nothing extracted there is nothing to prove.

## Behavior change

OFX imports now emit an extra directive when the statement carries `LEDGERBAL`. That is the roadmap's intent ("every OFX import becomes verified-complete"), but it does change output for existing users.

Five existing tests asserted a bare `directives.len()`, which no longer says what they mean. They now count transactions and assertions **separately** — more precise than what they replaced, rather than a bumped number.

## Verification

- 229 importer tests, 484 CLI tests, clippy and `cargo doc -D warnings` clean.
- Both features fail their tests when reverted.
- End to end through the built binary: the emitted file passes `rledger check`, which is the real proof the FITID sanitizer is correct.

## Roadmap

Updated both rows. Balance extraction phase 1 is shipped (CSV is phase 2). FITID is emitted but **not yet consumed by dedup** — preferring the id over the fuzzy key is the remaining half, and the one that makes dedup exact and rename-proof.


petemounce:
I just posted https://github.com/rustledger/rustledger/discussions/2268#discussioncomment-18345813 and I think that's another use-case for this PR's adding unique-ID `^`s to imported transactions - the ability to force (or reconstruct) the order of transactions according to the raw export. That sounds valuable based on the thread of discussion noting how beancount internal sorting functions today.

Issue #2109, opened 2026-08-21
[Bug]: Wrong VS Code extension sha256


davidedellagiustina:
### Description

VS Code extension fails to build after updating to v0.22.0:
```
error: hash mismatch in fixed-output derivation '/nix/store/d6r3hyrgqyz7z7r6jj8y5pw0a4wcjsiv-rustledger-vscode-vsix-0.22.0-npm-deps.drv':
         specified: sha256-+uUy4mzGgOyEv3e5O9oSkVZ/qZB1P6qaT+v4iBqfGuk=
            got:    sha256-rAiJeaxzfI0dZmHem8PE7ct3ssuee9FvXO6DVzIIkvc=
error: Cannot build '/nix/store/6mwrpvqf5s0rz52gfpdlhd3cd2glmanh-rustledger-vscode-vsix-0.22.0.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/fdfmhfsfbyyr1ada4nzyajbmzqqsc64k-rustledger-vscode-vsix-0.22.0
error: Cannot build '/nix/store/jc36cdcvizjgwg4di4dwq5758bqprhv7-vscode-extension-rustledger-vscode-0.22.0.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/6202smbga2ypywnf1f8msxa1yva49ga7-vscode-extension-rustledger-vscode-0.22.0
```

### Steps to Reproduce

1. Point flake input to release v0.22.0.
2. Rebuild system.
3. Observe error.

### Expected Behavior

Rebuild succeeds.

### Version

0.22.0

### Operating System

Linux

### Additional Context

I think this hash is problematic: https://github.com/rustledger/rustledger/blob/07177b081106ef3734d895c2c7c792a6e044e29c/flake.nix#L151

Obviously this occurs only if you try to evaluate the VS Code extension derivation.

robcohen:
Confirmed, and your diagnosis was exactly right — that hash is the problem. Thanks for the precise report; the `specified:`/`got:` pair made this a five-minute reproduction.

**Root cause.** `npmDepsHash` was set on 2026-08-09 in #2004, when the flake package was added, and `packages/vscode/package-lock.json` moved twice after it: 2026-08-13 (#2038, a js-yaml advisory) and 2026-08-17 (#2079, a dependency roll-up). Neither updated the hash. v0.22.0 is the first release to contain this derivation at all, so it shipped unbuildable.

[...]

PR #2283, opened 2026-09-08
ci(nix): refresh the vscode npmDepsHash automatically

robcohen:
`flake.nix` pins `npmDepsHash` over `packages/vscode/package-lock.json`, because `buildNpmPackage` hashes the vendored dependency tree. Dependabot updates the lockfile and has no way to know a Nix expression hashes it, so **every npm bump to that package lands red**:

```
error: hash mismatch in fixed-output derivation
         specified: sha256-Oc3yvMFep3AogJpFa5Ji4zIMXCIF/PGLJc7AM98mr7w=
            got:    sha256-RAG2KxhwVzCbaywKCclZHf5ySBdA/RBjbtd7NPAV+2c=
```

That is #2109, #2202, #2243, and both dependabot PRs open right now (#2269, #2272). `build` requires the `VS Code Extension (Nix)` job, so none of them can merge unattended.

**There was no tooling for it.** Nothing in `.github/`, `scripts/`, or the justfile mentioned `prefetch-npm-deps`. The repair was folklore, which is why it recurred rather than got handled.

## What this adds

**A workflow** that recomputes the hash on any PR touching the lockfile and commits it to the branch.

It is deliberately **separate** from the `VS Code Extension (Nix)` job. That job verifies; this one repairs. A verifier that silently rewrites the thing it is checking is worse than one that fails.

**It handles forks.** It cannot push to a fork branch, so it always writes the correct value to the job summary regardless:

> ### npmDepsHash is stale
> `packages/vscode/package-lock.json` changed, so `flake.nix` needs:
> ```nix
> npmDepsHash = "sha256-…";
> ```

A contributor whose PR it cannot fix still gets the exact line to change, instead of a bare Nix hash mismatch.

**The push is self-terminating.** It only commits when the computed hash differs, so the run that push triggers finds a match and stops. No loop.

**A local recipe**, `just nix-refresh-vscode-hash`, plus a CONTRIBUTING section explaining why the failure happens at all.

## Verification

I verified the command before writing a workflow around it, and verified the recipe both ways rather than only the happy path:

```
on main (hash correct):     ✓ npmDepsHash is already correct (sha256-Oc3yvM…)
after corrupting flake.nix: ✓ npmDepsHash sha256-AAAA… -> sha256-Oc3yvM…
                            ✓ flake.nix restored byte-identically
```

A repair tool that only ever reports "already correct" would be indistinguishable from one that does nothing.

## Note

This does not fix #2269 or #2272 themselves — they were opened before the workflow existed. Once this merges, closing and reopening them (or any rerun) will pick it up. Worth knowing that they interfere: each changes the lockfile, so whichever merges first invalidates the other's hash, and the second needs a fresh run.

### DIFF
--- a/.github/workflows/vscode-npm-deps-hash.yml
+++ b/.github/workflows/vscode-npm-deps-hash.yml
@@ -1,0 +1,114 @@
+# Keep `flake.nix`'s `npmDepsHash` in step with the VS Code extension's
+# lockfile.
+#
+# `buildNpmPackage` pins the hash of the vendored npm dependency tree, so any
+# change to `packages/vscode/package-lock.json` invalidates it and the Nix
+# build fails with a fixed-output mismatch. Dependabot updates the lockfile and
+# has no way to know a Nix expression hashes it, so *every* npm bump in this
+# package landed red until somebody recomputed the hash by hand (#2109, #2202,
+# #2243). `build` requires that job, so those PRs could never merge unattended.
+#
+# This recomputes the hash and commits it to the pull request. It is separate
+# from the `VS Code Extension (Nix)` job on purpose: that job verifies, this one
+# repairs, and a verifier that silently rewrites what it is checking is worse
+# than one that fails.
+name: VS Code npmDepsHash
+
+on:
+  pull_request:
+    paths:
+      - packages/vscode/package-lock.json
+      - flake.nix
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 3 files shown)


robcohen:
Superseded by #2284. The branch used a `ci/` prefix, which the branch-name gate rejects; same commit on a `chore/` branch.

PR #2331, opened 2026-09-16
fix(query): take `date ± days` and `date - date` in BQL

### LINKED ISSUE #2324: [Compat]: BQL `date ± <integer>` is rejected
### Description

beanquery accepts an integer number of days either side of `+`/`-` with a date. rledger rejects it, while implementing the same operation twice over, as `date_add()` and as `date ± interval(n, 'unit')`.

### Example

```
SELECT account, sum(position) AS total 
WHERE account ~ '^Expenses'
AND date >= 2026-07-01 AND date < 2026-07-01 + 365
GROUP BY account ORDER BY total DESC
```

 | expression | rledger 0.24.0 | bean-query (beanquery 0.2.0) |
  |---|---|---|
  | `2026-07-01 + 365` | ❌ type error | `2027-07-01` |
  | `365 + 2026-07-01` | ❌ type error | `2027-07-01` |
  | `2026-07-01 - 365` | ❌ type error | `2025-07-01` |
  | `date_add(2026-07-01, 365)` | ✅ `2027-07-01` | ✅ `2027-07-01` |
  | `2026-07-01 + interval(365, 'day')` | ✅ `2027-07-01` | ❌ not a function |

Requiring `interval()` may well be deliberate, since it names the unit rather than leaving it implicit. Flagging this just in case parity with beanquery's syntax is a goal.

### Repro
```bash
rledger query any.beancount "SELECT 2026-07-01 + 365 AS v LIMIT 1"
# => error: failed to execute query: type error: arithmetic requires numeric values
```

### Python Beancount Version

beanquery 0.2.0

### Rustledger Version

0.24.0

robcohen:
Closes #2324.

`beanquery` accepts an integer day count on either side of `+`, and the number of days between two dates for `-`. rledger rejected all three with `type error: arithmetic requires numeric values`, while implementing the same shift twice over as `date_add(date, n)` and `date + interval(n, 'day')`. A window written the obvious way, `date < 2026-07-01 + 365`, did not work as typed.

| expression | before | after | beanquery 0.2.0 |
|---|---|---|---|
| `2026-07-01 + 365` | type error | `2027-07-01` | `2027-07-01` |
| `365 + 2026-07-01` | type error | `2027-07-01` | `2027-07-01` |
| `2026-07-01 - 365` | type error | `2025-07-01` | `2025-07-01` |
| `2026-07-01 - 2026-06-01` | type error | `30` | `30` |
| `2026-07-01 + 30.0` | type error | `2026-07-31` | type error |
| `2026-07-01 + 0.5` | type error | error naming the rule | type error |
| `365 - 2026-07-01` | type error | type error | type error |
| `date_add(2026-07-01, 0.5)` | **`2026-07-01`** | error naming the rule | rejects the call |
| `date_add`, `+ interval(n, 'day')` | works | unchanged | — |

## One rule, two callers

`Value::as_day_count` reads a number as a count of days, and both the `date ± n` operators and `DATE_ADD` go through it, so a function and an operator cannot answer the same number two different ways.

The rule is deliberately **wider than beanquery in one direction and narrower in another**:

- **Wider**: a whole-valued decimal such as `30.0` is accepted. beanquery refuses every decimal (`operator "add(date, decimal)" not supported`). The meaning is unambiguous, and refusing it is pedantry rather than safety.
- **Narrower**: a genuine fraction is refused. A date has day resolution, so `date + 0.5` has no answer to give, and guessing one is worse than saying so.

## A silent wrong answer, fixed on the way

`DATE_ADD` ran `Decimal::to_i64` on its second argument, which **truncates**. On main:

```
SELECT date_add(2026-07-01, 0.5)   ->  2026-07-01     (added no days at all)
SELECT date_add(2026-07-01, 1.9)   ->  2026-07-02     (added one)
```

Neither is flagged. beanquery refuses the call outright; this now reports which rule was broken instead of guessing which day the caller meant.

`date - date` is computed the way `DATE_DIFF` computes it — same expression — so the operator and the function cannot drift apart.

## Verified

- Every row of the table above was run against `beanquery 0.2.0` / `beancount 3.2.3` and against binaries built from `main` and from this branch.
- The reporter's own query shape (`WHERE ... AND date >= 2026-07-01 AND date < 2026-07-01 + 365 GROUP BY ...`) returns the in-window rows only.
- Three new tests: the accepted forms, the forms that must stay errors, and `DATE_ADD`'s fractional count.
- 715 `rustledger-query` tests pass; workspace clippy at 1.98.1 and `cargo fmt` are clean.

`docs/reference/bql.md` gains a Date Arithmetic section stating the whole-days rule and pointing at `interval` for units other than days.


federicobond:
Thanks for que quick response! ❤️

